PR #14, opened 2026-02-08
MDBook download + cache speedup

### LINKED ISSUE #3: Speed up mdbook action
We're using the officially recommended mdbook GitHub Action, but it compiles mdbook from source every time which takes a few mins and wastes compute. It seems like we should be fine to grab the binaries from the mdbook release on GitHub or at least cache out built binaries in CI.

juanmrad:
Fixes #3

This Action changes from building mdbook to downloading the release and caching it. Should speed up the process of our action by looking from cache or downloading the release over building from source.

### DIFF
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -47,14 +47,11 @@ jobs:
           tar -xzf mdbook.tar.gz -C mdbook-bin --strip-components=1
           rm mdbook.tar.gz
 
-      - name: Add mdBook to PATH
-        run: echo "${{ github.workspace }}/mdbook-bin" >> $GITHUB_PATH
-
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v5
       - name: Build with mdBook
-        run: mdbook build docs
+        run: ${{ github.workspace }}/mdbook-bin/mdbook build docs
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v3
         with:


.github/workflows/mdbook.yml
```diff
@@ -28,11 +28,28 @@ jobs:
...
         run: |
-          curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf -y | sh
-          rustup update
-          cargo install --version ${MDBOOK_VERSION} mdbook
+          ARCH="x86_64-unknown-linux-gnu"
+          ARCHIVE="mdbook-v${MDBOOK_VERSION}-${ARCH}.tar.gz"
+          URL="https://github.com/rust-lang/mdBook/releases/download/v${MDBOOK_VERSION}/${ARCHIVE}"
+          mkdir -p mdbook-bin
+          curl -sSL -o mdbook.tar.gz "$URL"
+          tar -xzf mdbook.tar.gz -C mdbook-bin --strip-components=1
```
vinaysrao1: I just looked up --strip-components command. looks like it is stripping leading directories in the path. making sure that mdbook tarball has leading directory names. (I could be wrong)

vinaysrao1 (changes requested):

.github/workflows/mdbook.yml
```diff
@@ -28,11 +28,28 @@ jobs:
...
+          path: mdbook-bin
+          key: mdbook-${{ env.MDBOOK_VERSION }}-${{ runner.os }}-${{ runner.arch }}
+
       - name: Install mdBook
+        if: steps.cache-mdbook.outputs.cache-hit != 'true'
         run: |
-          curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf -y | sh
-          rustup update
-          cargo install --version ${MDBOOK_VERSION} mdbook
+          ARCH="x86_64-unknown-linux-gnu"
```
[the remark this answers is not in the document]
juanmrad: This only runs on CI in an ubuntu runner.

.github/workflows/mdbook.yml
juanmrad: just doubled checked, and yes the released tarballs have one top-level dir with the binary inside.

cassidyjames:
At a glance I think this will fail because we're downloading and caching `mdbook-bin` but then calling `mdbook`. I'm on mobile at the moment and can more fully review it on a desktop later today.

PR #16, opened 2026-02-08
Node Upgrades for Images and compatibility

juanmrad:

This PR upgrades nodeJS servions to Node24 from our current Node20. This is to keep up with LTS as 20 is out of LTS. 

This PR also upgrade HMA as a wider dependency version upgrade.


vinaysrao1 (approved):
The move to node 24 is fine. We probably need an issue to check for others. I randomly checked a few packages from .devops/migrator/package.json and found that with node 24 we don't need glob anymore and we may need to upgrade cassandra-driver to 4.8

vinaysrao1 (approved):

PR #63, opened 2026-02-16
NCMEC <-> Queue Improvements and default NCMEC queue

juanmrad:

Adds a configurable Default NCMEC queue in NCMEC settings. When reviewers (or rules) use Enqueue to NCMEC, jobs are sent to this queue when set; otherwise they continue to use the org’s default manual review queue.

NCMEC Settings (/dashboard/settings/ncmec): New “Default NCMEC queue” dropdown. When set, “Enqueue to NCMEC” sends jobs to that queue instead of the org default.

## Tests:

JobRouting test that enqueue with an explicit queueId adds the job to that queue and skips routing.

Screenshot 2026-02-16 at 5 03 25 PM

### DIFF
--- a/.devops/migrator/src/scripts/api-server-pg/2026.02.10T00.00.00.add_default_ncmec_queue_to_ncmec_org_settings.sql
+++ b/.devops/migrator/src/scripts/api-server-pg/2026.02.10T00.00.00.add_default_ncmec_queue_to_ncmec_org_settings.sql
@@ -1,0 +1,12 @@
+-- Add default NCMEC queue to ncmec_org_settings so Enqueue to NCMEC can target a specific queue.
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD COLUMN IF NOT EXISTS default_ncmec_queue_id character varying(255) NULL;
+
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD CONSTRAINT ncmec_org_settings_default_ncmec_queue_fkey
+  FOREIGN KEY (default_ncmec_queue_id)
+  REFERENCES manual_review_tool.manual_review_queues(id)
+  ON DELETE SET NULL;
+
+COMMENT ON COLUMN ncmec_reporting.ncmec_org_settings.default_ncmec_queue_id IS
+  'When set, Enqueue to NCMEC sends jobs to this queue instead of the org default.';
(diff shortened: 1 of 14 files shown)


.devops/migrator/src/scripts/api-server-pg/2026.02.10T00.00.00.add_default_ncmec_queue_to_ncmec_org_settings.sql
```diff
@@ -0,0 +1,12 @@
+-- Add default NCMEC queue to ncmec_org_settings so Enqueue to NCMEC can target a specific queue.
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD COLUMN IF NOT EXISTS default_ncmec_queue_id character varying(255) NULL;
+
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD CONSTRAINT ncmec_org_settings_default_ncmec_queue_fkey
```
vinaysrao1: for idempotency should we say "add constraint if not exists"?

vinaysrao1 (changes requested):

.devops/migrator/src/scripts/api-server-pg/2026.02.10T00.00.00.add_default_ncmec_queue_to_ncmec_org_settings.sql
juanmrad: we are adding a constraint here. PostgreSQL doesn't support `ADD CONSTRAINT IF NOT EXISTS` we could complicate this adding a do block, but migrations wouldn't run more than once,.

server/services/ncmecService/ncmecService.ts
```diff
@@ -214,6 +221,7 @@ export class NcmecService {
           params.ncmecPreservationEndpoint ?? undefined,
         ncmec_additional_info_endpoint:
           params.ncmecAdditionalInfoEndpoint ?? undefined,
+        default_ncmec_queue_id: params.defaultNcmecQueueId ?? undefined,
```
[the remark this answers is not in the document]
juanmrad: yeah you are right I forgot to check this I mostly added it due to lint lol. But sending an undefined to kysely will cause it not to update over what we want which is being null so it uses the default queue.

vinaysrao1 (approved):

PR #72, opened 2026-02-19
Remove Cove marketing video

serendipty01:
I believe this video got missed.


julietshen (approved):
Thank you so much for this!

cassidyjames (approved):
Thanks! I double checked and it doesn't appear to be referenced anywhere else in the codebase, either. Good find.

Issue #40, opened 2026-02-10
Replace Craco with Vite


cassidyjames:
Craco is deprecated, so move to a better framework.

Related: https://medium.com/@sargun.kohli152/migrating-from-cra-to-vite-a-step-by-step-guide-f6b2a4b6545d

PR #74, opened 2026-02-20
Migrate to vite

pawiecz:

Fixes: #40

## Tests

[Standard instance setup steps](https://github.com/roostorg/coop/blob/main/README.md#getting-started) remain unchanged and produce the same results (changes are hidden under `npm` scripts abstractions).

## (Optional) Rollout Plan

Updated dev environments may seem slow on the first start (hence a note in the README `Getting started` section).

### DIFF
--- a/client/.dockerignore
+++ b/client/.dockerignore
@@ -3,6 +3,9 @@
 !src/
 !public/
 !package*.json
+!index.html
+!vite.config.ts
+!vite-env.d.ts
 !postcss.config.js
 !tailwind.config.js
 !tsconfig.json


client/package.json
juanmrad: The jest config block and the overrides for `react-scripts/react-dev-utils` are still here. Since we've moved to vitest (configured in `vite.config.ts`), these are now dead config. Can we clean them up?

juanmrad (changes requested):
Tested locally -- both server and client start and work correctly. Vite dev server starts in ~235ms and the proxy to the backend works as expected. Production build also completes successfully in ~7s. Nice migration!

A few cleanup items below, but the core migration is solid.

Also to note. The build produces a 3MB Dashboard chunk. Agree with @cassidyjames  that performance optimization should be a follow-up issue -- Vite's `build.rollupOptions.output.manualChunks` could help split this up.

pawiecz:
I'm unable to reproduce `vite build` failure locally (it finishes successfully) while it [continues to fail in the CI job](https://github.com/roostorg/coop/actions/runs/23192722174/job/67394421108?pr=74#step:4:54).

PR #84, opened 2026-02-27
feat: Add OIDC SSO

### LINKED ISSUE #335: Add OIDC SSO authentication
Currently, Coop only supports SAML for Single Sign-On (SSO). We should add OpenID Connect (OIDC) support to enable modern authentication flows and provide better compatibility with identity providers that support OIDC.

serendipty01:
fixes #335

Currently, Coop only supports SAML for Single Sign-On (SSO). This PR adds OpenID Connect (OIDC) support to enable modern authentication flows and provide better compatibility with identity providers that support OIDC.

This PR is using [openid-client](https://npmx.dev/package/openid-client) library

## TODO
- [x] Lint
- [x] Test with Okta
- [x] Test more scenarios
- [x] Clean up logs and comments
- [x] Add env variables to .env.example
- [x] Check whether to use `client_secret_basic` or `client_secret_post`
- [x] Update docs

## Tests
Tested locally using Authelia
will add screenshots/video later

## Summary by CodeRabbit

* **New Features**
  * Added OIDC as a supported SSO login method alongside existing SAML.
  * Redesigned SSO Settings page with separate SAML and OIDC configuration tabs and method switching capability.
  * Added issuer domain validation and OIDC discovery for credential verification.

* **Documentation**
  * Added comprehensive architecture, development, and user guides.


juanmrad (changes requested):

server/services/orgSettingsService/orgSettingsService.ts
```diff
@@ -19,19 +20,15 @@ export type OrgSettingsPg = {
...
-      }
-    | {
-        saml_enabled: false;
-        sso_url: string | null;
-        cert: string | null;
-      }
-  );
+    saml_enabled: boolean;
+    sso_url: string | null;
+    // TODO: rename this to something more descriptive like sso_cert
```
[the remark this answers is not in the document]
serendipty01: I did not want to touch the [existing SAML code](https://github.com/roostorg/coop/blob/71ffb3d8622f697398820bffb9ddcd99ca4bf0ce/server/services/orgSettingsService/orgSettingsService.ts#L26)

I have separate commit for fixing SAML: [Diff Link](https://github.com/serendipty01/coop/compare/add-oidc-sso...serendipty01:coop:fix-saml-sso?expand=1). I was thinking of raising a PR after more testing and refining the code. 

Should i do the SAML changes in this PR or raise a new PR ?

server/services/orgSettingsService/orgSettingsService.ts
```diff
@@ -175,20 +176,81 @@ function makeOrgSettingsService(pgQuery: Kysely<OrgSettingsPg>) {
...
+      samlEnabled: boolean;
       ssoUrl: string;
       cert: string;
     }) {
       try {
         await pgQuery
           .updateTable('public.org_settings')
           .where('org_id', '=', input.orgId)
-          .set({ sso_url: input.ssoUrl, cert: input.cert })
+          .set({ saml_enabled: input.samlEnabled, sso_url: input.ssoUrl, cert: input.cert })
```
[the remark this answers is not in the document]
serendipty01: updated the PR

server/api.ts
```diff
@@ -269,6 +286,118 @@ export default async function makeApiServer(deps: Dependencies) {
...
+
+      // Reconstruct callback URL with code/state from IdP redirect.
+      // authorizationCodeGrant needs: base = registered redirect_uri, query = code+state from IdP.
+      const currentUrl = new URL(`${API_BASE_URL}/api/v1/oidc/login/callback`);
+      for (const [k, v] of new URLSearchParams(req.url.includes('?') ? req.url.split('?')[1] : '')) {
+        currentUrl.searchParams.set(k, v);
+      }
+
+      const checks: { pkceCodeVerifier: string; expectedState?: string } = { pkceCodeVerifier: codeVerifier };
+      if (state) checks.expectedState = state;
```
juanmrad: I believe state is a CSRF protection. we shouldn't allow state to be empty here. Also we are silently skipping here.

server/api.ts
```diff
@@ -269,6 +286,118 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        return res.redirect('/');
+      }
+
+      const { code_verifier: codeVerifier, state, org_id: orgId } = sessionData;
+      // Clear session OIDC state immediately (one-time use)
+      delete req.session.oidc;
+
+      const oidcSettings = await deps.OrgSettingsService.getOidcSettings(orgId);
+      if (!oidcSettings || !oidcSettings.client_id || !oidcSettings.client_secret || !oidcSettings.issuer_url) {
+        return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=sso_login_failed`);
```
juanmrad: I think here we could add a small log line so on the server we know the error. given we are now swallowing all errors and not reporting on the front-end.

server/graphql/modules/org.ts
```diff
@@ -52,8 +79,14 @@ const typeDefs = /* GraphQL */ `
     userStrikeThresholds: [UserStrikeThreshold!]!
     userStrikeTTL: Int!
     isDemoOrg: Boolean!
+    samlEnabled: Boolean!
     ssoUrl: String
     ssoCert: String
+    oidcEnabled: Boolean!
+    issuerUrl: String
+    clientId: String
+    clientSecret: String
```
juanmrad: Is there a reason we need to return the secret?. I think we can just ignore it here to prevent leaking it. I'We should be good to remove from Org type and its resolver entirely. The client already knows to show a placeholder when OIDC is enabled (it checks oidcEnabled to show `•••••••• (enter new value to change)` in the input field). No schema field needed.

And drop clientSecret from the SwitchSSOMethod response fragment in SSOSettings.tsx.

.devops/migrator/src/scripts/api-server-pg/2026.03.02T00.00.00.add-oidc-sso.sql
```diff
@@ -0,0 +1,21 @@
+ -- Add OIDC columns
+ALTER TABLE public.org_settings
+ADD COLUMN IF NOT EXISTS oidc_enabled boolean DEFAULT false NOT NULL,
+ADD COLUMN IF NOT EXISTS issuer_url character varying(255),
+ADD COLUMN IF NOT EXISTS client_id character varying(255),
+ADD COLUMN IF NOT EXISTS client_secret text;;
```
juanmrad:
```suggestion
ADD COLUMN IF NOT EXISTS client_secret text;
```

remove double semi colon.

server/api.ts
```diff
@@ -269,6 +286,118 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        return next(makeBadRequestError('OIDC not enabled for this organization.', { shouldErrorSpan: true }));
+      }
+      if (!oidcSettings.client_id || !oidcSettings.client_secret || !oidcSettings.issuer_url) {
+        return next(makeInternalServerError('Missing OIDC credentials for org.', { shouldErrorSpan: true }));
+      }
+
+      const { API_BASE_URL } = process.env;
+      if (!API_BASE_URL) {
+        return next(makeInternalServerError('API_BASE_URL not configured.', { shouldErrorSpan: true }));
+      }
```
juanmrad: would be good to consistently use `ConfigService`

server/graphql/modules/org.ts
```diff
@@ -638,18 +718,91 @@ const Mutation: GQLMutationResolvers = {
...
+    }
+    if (!user.getPermissions().includes('MANAGE_ORG')) {
+      throw new AuthenticationError(
+        'User does not have permission to manage SSO settings',
+      );
+    }
+
+    const samlSettings = await context.services.OrgSettingsService.getSamlSettings(user.orgId);
+    if (samlSettings?.saml_enabled && input.oidcEnabled) {
+      throw new Error('OIDC cannot enabled as SAML is enabled.');
```
juanmrad:
```suggestion
      throw new Error('OIDC cannot be enabled as SAML is enabled.');
```

server/graphql/modules/org.ts
```diff
@@ -638,18 +718,91 @@ const Mutation: GQLMutationResolvers = {
...
       throw new AuthenticationError('User required.');
     }
+    if (!user.getPermissions().includes('MANAGE_ORG')) {
+      throw new AuthenticationError(
+        'User does not have permission to manage SSO settings',
+      );
+    }
+    const oidcSettings = await context.services.OrgSettingsService.getOidcSettings(user.orgId);
+    if (oidcSettings?.oidc_enabled && input.samlEnabled) {
+      throw new Error('SAML cannot enabled as OIDC is enabled.');
```
juanmrad:
```suggestion
      throw new Error('SAML cannot be enabled as OIDC is enabled.');
```

juanmrad (commented):
a couple more comments. 😅 Thanks for all the hard work.

juanmrad:
@serendipty01 left you a couple more comments. let me know if you need help, and again thanks for contributing.

server/test/arbitraries/ContentType.ts
```diff
@@ -47,7 +47,7 @@ export const GeohashArbitrary = fc
   });
 
 export const DateStringArbitrary = fc
-  .date()
+  .date({ noInvalidDate: true })
```
serendipty01, line 52: this change is for fixing a test failure

juanmrad (approved):
@serendipty01 Just to doubly confirm you manually tested this change correct?

server/utils/crypto.ts
ThisIsMissEm, line 1: You may want to use boringnode/encryption here instead of handle rolling cryptography code: https://github.com/boringnode/encryption/tree/1.x

server/.env.example
```diff
@@ -82,3 +82,10 @@ GROQ_SECRET_KEY=
 ITEM_QUEUE_TRAFFIC_PERCENTAGE='0.05'
 UI_URL=http://localhost:3000
 GRAPHQL_MAX_DEPTH=10
+
+API_BASE_URL=http://localhost:8080
+
+# generate by running any of the below commands
+# openssl rand -base64 32
+# node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"
+SSO_ENCRYPTION_KEY=
```
ThisIsMissEm: You probably also want this to just be an `ENCRYPTION_KEY` instead of SSO specific

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+      delete req.session.oidc;
+
+      const oidcSettings = await deps.OrgSettingsService.getOidcSettings(orgId);
+      if (!oidcSettings || !oidcSettings.client_id || !oidcSettings.client_secret || !oidcSettings.issuer_url) {
+        // eslint-disable-next-line no-console
+        console.error('[OIDC] Missing or incomplete OIDC settings for org', orgId);
+        return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=sso_login_failed`);
+      }
+
+      const issuerUrl = normalizeIssuerUrl(oidcSettings.issuer_url);
```
ThisIsMissEm: you should probably normalize on configuration, rather than each login attempt.

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+
+      const config = await discoverOidcConfig(issuerUrl, oidcSettings.client_id, oidcSettings.client_secret);
+
+      // Reconstruct callback URL with code/state from IdP redirect.
+      // authorizationCodeGrant needs: base = registered redirect_uri, query = code+state from IdP.
+      const currentUrl = new URL(`${deps.ConfigService.apiUrl}/api/v1/oidc/login/callback`);
```
ThisIsMissEm: wouldn't `new URL(req.url, deps.ConfigService.apiUrl)` work here instead of manually copying the search parameters? Or, perhaps better: `req.originalUrl`

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        console.error('[OIDC] No email in token claims or userinfo');
+        return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=no_email`);
+      }
+
+      const user = await User.findOne({ where: { email: String(email) } });
+
+      if (!user || user.orgId !== orgId) {
+        // eslint-disable-next-line no-console
+        console.error('[OIDC] User not found or org mismatch', { email, orgId, userOrgId: user?.orgId });
+        return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=user_not_found`);
```
ThisIsMissEm: no email and user_not_found should probably both be the same error as to try to stop credential guessing through oauth

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+      }
+
+      if (!user.loginMethods.includes('oidc')) {
+        user.loginMethods = [...user.loginMethods, 'oidc'];
+        await user.save();
+      }
```
ThisIsMissEm: This is probably an unsafe upgrade to OIDC for authentication, and should only be done if the user is completing a flow for connecting their local account with an OIDC account; otherwise, it should not be done, as this would allow an attacker gaining control over the OIDC server to then access another internal tool as long as they know someone's email that likely has an account on said tool.

server/api.ts
ThisIsMissEm: Use `sso_login_failed` for both of these.

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=oidc_not_enabled`);
+      }
+      if (!oidcSettings.client_id || !oidcSettings.client_secret || !oidcSettings.issuer_url) {
+        // eslint-disable-next-line no-console
+        console.error('[OIDC] Missing credentials for org', orgId, { hasClientId: Boolean(oidcSettings.client_id), hasSecret: Boolean(oidcSettings.client_secret), hasIssuer: Boolean(oidcSettings.issuer_url) });
+        return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=oidc_misconfigured`);
+      }
+
+      const callbackUrl = deps.SSOService.getSSOOidcCallbackUrl();
+      const issuerUrl = normalizeIssuerUrl(oidcSettings.issuer_url);
```
ThisIsMissEm: normalize on storage, not on read

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+      req.login(user as any, (err) => {
+        if (err) return next(err);
+        res.redirect(`${deps.ConfigService.uiUrl}/dashboard`);
+      });
+    } catch (e) {
+      return res.redirect(`${deps.ConfigService.uiUrl}/login/sso?error=sso_login_failed`);
+    }
+  });
+
+  app.get('/oidc/login/:orgId', async (req, res, next) => {
```
ThisIsMissEm: this should be a POST request because it's starting an authorization flow.

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        redirect_uri: callbackUrl,
+        scope: 'openid email',
+        code_challenge: codeChallenge,
+        code_challenge_method: 'S256',
+      };
+
+      const redirectTo = oidcClient.buildAuthorizationUrl(config, params);
+
+      const state = oidcClient.randomState();
+      redirectTo.searchParams.set('state', state);
```
ThisIsMissEm: you can move the state generation up to along with codeVerifier and challenge, and then just pass in params as a `URLSearchParams({ ... })` with a `state` key/property. `buildAuthorizationUrl` should automatically include the state then in the authorization redirect URL

server/api.ts
```diff
@@ -265,6 +292,131 @@ export default async function makeApiServer(deps: Dependencies) {
...
+
+      // Store PKCE data + orgId in session for callback
+      req.session.oidc = { code_verifier: codeVerifier, state, org_id: orgId };
```
ThisIsMissEm: double check that this is stored securely, and not in a plaintext cookie or something.

ThisIsMissEm (commented):
Left some comments that may be worth addressing before merge.

server/api.ts
serendipty01: fixed

server/api.ts
serendipty01: fixed

server/api.ts
serendipty01: fixed

server/api.ts
serendipty01: fixed

server/api.ts
serendipty01: fixed

server/api.ts
```diff
@@ -256,6 +293,246 @@ export default async function makeApiServer(deps: Dependencies) {
     },
   );
 
+  app.get('/oidc/login/callback', async (req, res, next) => {
```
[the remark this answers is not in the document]
serendipty01: not required

server/api.ts
```diff
@@ -256,6 +293,246 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        res.redirect(`${deps.ConfigService.uiUrl}/dashboard`);
+      });
+    } catch (e) {
+      return res.redirect(
+        `${deps.ConfigService.uiUrl}/login/sso?error=sso_login_failed`,
+      );
+    }
+  });
+
+  app.post('/oidc/login/:orgId', async (req, res, next) => {
```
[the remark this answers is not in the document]
serendipty01: not required

server/api.ts
```diff
@@ -256,6 +293,246 @@ export default async function makeApiServer(deps: Dependencies) {
     },
   );
 
+  app.get('/oidc/login/callback', async (req, res, next) => {
+    if (req.query.error) {
+      return res.redirect(
+        `${deps.ConfigService.uiUrl}/login/sso?error=${req.query.error}`,
+      );
+    }
```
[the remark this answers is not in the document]
serendipty01: done

server/api.ts
```diff
@@ -256,6 +293,246 @@ export default async function makeApiServer(deps: Dependencies) {
...
+        await oidcClient.calculatePKCECodeChallenge(codeVerifier);
+      const state = oidcClient.randomState();
+
+      const params = new URLSearchParams({
+        redirect_uri: callbackUrl,
+        scope: 'openid email',
+        code_challenge: codeChallenge,
+        code_challenge_method: 'S256',
+        state,
+      });
```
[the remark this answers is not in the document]
serendipty01: done

server/graphql/modules/org.ts
```diff
@@ -765,22 +839,106 @@ const Mutation: GQLMutationResolvers = {
...
+
+    const oidcSettings =
+      await context.services.OrgSettingsService.getOidcSettings(user.orgId);
+
+    if (oidcSettings?.oidc_enabled && input.samlEnabled) {
+      throw new Error('SAML cannot be enabled as OIDC is enabled.');
+    }
 
     if (!user.getPermissions().includes('MANAGE_ORG')) {
       throw forbiddenError('User does not have permission to manage SSO settings');
```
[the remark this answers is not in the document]
serendipty01: done

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -87,138 +271,233 @@ export default function SSOSettings() {
...
+              {isEnablingFromPassword
+                ? `Enable ${activeTab} authentication?`
+                : `Switch SSO method to ${activeTab}?`}
+            </DialogTitle>
+          </DialogHeader>
+          <DialogDescription>
+            {isEnablingFromPassword
+              ? `Enabling ${activeTab} is irreversible. All org users will be required to authenticate through the new provider immediately.`
+              : `This will disable ${currentMethod} and enable ${activeTab}. Your ${currentMethod} credentials will be preserved in case you switch back, but users will need to authenticate through the new provider immediately.`}
+          </DialogDescription>
```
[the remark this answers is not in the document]
serendipty01: fixed

juanmrad (commented):
You may have accidentally added the PNPM migration code here. please remove.

client/src/utils/oidc.test.ts
```diff
@@ -0,0 +1,35 @@
+import { isValidIssuerDomain, normalizeIssuerDomain } from './oidc';
+
+describe('normalizeIssuerDomain', () => {
+  test('keeps a bare domain unchanged', () => {
+    expect(normalizeIssuerDomain('your-tenant.auth0.com')).toBe(
+      'your-tenant.auth0.com',
+    );
+  });
+
+  test('strips protocol and path from a discovery URL', () => {
```
ThisIsMissEm: you'll also want a test for the horrid oidc thing of: `https://auth.example/tenant/.well-known/openid-configuration` which keycloak uses

client/src/utils/oidc.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    ).toBe('your-tenant.auth0.com');
+  });
+});
+
+describe('isValidIssuerDomain', () => {
+  test('accepts a normalized domain', () => {
+    expect(isValidIssuerDomain('your-tenant.auth0.com')).toBe(true);
+  });
+
+  test('accepts a discovery URL by normalizing it first', () => {
```
ThisIsMissEm, line 44: This should probably actually just fetch the discovery URL and use the `issuer` field from that response. So if the path ends with `/.well-known/openid-configuration` (or potentially also `/.well-known/oauth-authorization-server`) then resolve the discovery URL and use `issuer` as indicated in that document. Though that does make this async.

client/src/utils/oidc.ts
```diff
@@ -0,0 +1,29 @@
+export function normalizeIssuerDomain(value: string) {
```
ThisIsMissEm, line 1: This logic isn't actually correct for OIDC, as `issuer` can contain a path. e.g., a discovery URL of: `https://auth.site.example/realms/example/.well-known/openid-configuration` can have it's `issuer` as `https://auth.site.example/realms/example/`

So this logic would currently be incompatible with OIDC

https://openid.net/specs/openid-connect-core-1_0.html#IssuerIdentifier

Essentially this comes from when the well-known RFC draft allowed a path prefix, which was removed from the final spec.

client/src/webpages/auth/LoginSSO.tsx
```diff
@@ -27,6 +30,16 @@ export default function LoginSSO() {
   const [errorModalVisible, setErrorModalVisible] = useState(false);
 
   const [getSSORedirectUrl, { loading }] = useGQLGetSsoRedirectUrlLazyQuery();
+
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search);
+    const error = params.get('error');
```
ThisIsMissEm, line 38: there's also an `error_description` parameter that's possible to use here. Here's the list: https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-09.html#section-4.1.2.1

client/src/webpages/auth/SignUp.tsx
```diff
@@ -133,6 +134,21 @@ export default function SignUp() {
...
+      await signUp({
+        variables: {
+          input: {
+            email: tokenInfo.email,
+            firstName,
+            lastName,
+            role: tokenInfo.role,
+            orgId: tokenInfo.orgId,
+            inviteUserToken: token!,
+            loginMethod: 'OIDC',
```
ThisIsMissEm: You probably just want to pass the token and any overridden fields rather than have the client decide completely, since this could allow a maliciously crafted request to look like it did come from OIDC when it in fact did not.

client/src/webpages/auth/ssoRedirect.ts
```diff
@@ -0,0 +1,17 @@
+export function isOidcLoginStartUrl(redirectUrl: string): boolean {
+  return redirectUrl.includes('/api/v1/oidc/login/');
```
ThisIsMissEm: Should this be `endsWith`? It's also weird this is happening on the client-side instead of the server-side.

client/src/webpages/auth/ssoRedirect.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export function redirectToSsoUrl(redirectUrl: string) {
```
ThisIsMissEm: I see you're using `POST` to the authorization URL because OIDC does allow it, OAuth does not necessarily: https://datatracker.ietf.org/doc/html/rfc6749#section-3.1

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -28,44 +41,80 @@ gql`
...
-    }
-
-    if (data.myOrg.ssoCert != null) {
-      setSsoCert(data.myOrg.ssoCert);
-    }
+    if (data?.myOrg == null) return;
+    const org = data.myOrg;
+    if (org.ssoUrl != null) setSsoUrl(org.ssoUrl);
+    if (org.ssoCert != null) setSsoCert(org.ssoCert);
+    if (org.issuerUrl != null) setIssuerUrl(normalizeIssuerDomain(org.issuerUrl));
```
ThisIsMissEm: This is coming from saved data so probably shouldn't be being normalized. Normalization / validation should happen on the save path.

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -75,14 +124,148 @@ export default function SSOSettings() {
...
+      },
+      onError: (e) => {
+        toast.error(`Error updating SAML credentials: ${e.message}`);
+        if (closeDialog) setShowSwitchDialog(false);
+      },
+    });
+  };
+
+  const handleSaveOidc = (closeDialog = false) => {
+    const normalizedIssuerUrl = normalizeIssuerDomain(issuerUrl);
```
ThisIsMissEm, line 171: You probably want to do OIDC discovery here instead of just string normalization, since the configuration document includes the actual `issuer` value.

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -75,14 +124,148 @@ export default function SSOSettings() {
...
+    if (!isValidIssuerDomain(normalizedIssuerUrl)) {
```
ThisIsMissEm, line 172: This could do `URL.canParse(issuerUrl.trim())`

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -75,14 +124,148 @@ export default function SSOSettings() {
...
+      toast.error('Domain is not valid (e.g. your-tenant.auth0.com)');
+      return;
+    }
+    updateSSOOidcCredentials({
```
ThisIsMissEm: this graphql query is what should actually validate that the data is correct, and return back the new settings as validated & saved in the database, replacing the form state.

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -28,44 +41,80 @@ gql`
...
+
 export default function SSOSettings() {
-  const [ssoUrl, setSsoUrl] = useState<string | undefined>(undefined);
-  const [ssoCert, setSsoCert] = useState<string | undefined>(undefined);
+  const [activeTab, setActiveTab] = useState<Tab>('SAML');
+  const [ssoUrl, setSsoUrl] = useState('');
+  const [ssoCert, setSsoCert] = useState('');
+  const [issuerUrl, setIssuerUrl] = useState('');
+  const [clientId, setClientId] = useState('');
+  const [clientSecret, setClientSecret] = useState('');
```
ThisIsMissEm: I would strongly recommend a `useForm` here, though react can also give you a formData object: https://react.dev/reference/react-dom/components/form#handle-form-submission-on-the-client

So you don't need controlled inputs.

There's also libraries like https://www.react-hook-form.com/api/useform/ for a `useForm` hook if you need something more complex.

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -75,14 +124,148 @@ export default function SSOSettings() {
...
+  const oidcCallbackUri = callbackData?.getSSOCallbackUrls.oidcCallbackUrl ?? '';
+
+  const stringIsAValidUrl = (s: string) => {
+    try {
+      // eslint-disable-next-line no-new
+      new URL(s);
+      return true;
+    } catch (_) {
+      return false;
+    }
```
ThisIsMissEm:
```suggestion
    return URL.canParse(s)
```

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -75,14 +124,148 @@ export default function SSOSettings() {
...
+
+  const handleSaveSaml = (closeDialog = false) => {
+    if (!stringIsAValidUrl(ssoUrl)) {
+      toast.error('SSO URL is not a valid URL');
+      return;
+    }
+    updateSSOSamlCredentials({
+      variables: { input: { samlEnabled: true, ssoUrl, ssoCert } },
+      refetchQueries: ['GetSSOCredentials'],
+      onCompleted: () => {
```
ThisIsMissEm, line 157: This should update the form with the response body, such that post-save the form state represents what the server actually saved.

client/src/webpages/settings/SSOSettings.tsx
```diff
@@ -75,14 +124,148 @@ export default function SSOSettings() {
...
+        input: {
+          oidcEnabled: true,
+          issuerUrl: normalizedIssuerUrl,
+          clientId,
+          clientSecret,
+        },
+      },
+      refetchQueries: ['GetSSOCredentials'],
+      onCompleted: () => {
+        setIssuerUrl(normalizedIssuerUrl);
```
ThisIsMissEm: This should be the value the server indicates, not the state value.

client/src/webpages/settings/SSOSettings.tsx
ThisIsMissEm, line 1: This file is pretty large, it would make sense to split it up into sub components:
- OIDC Form
- SAML Form
- Password Form

db/src/scripts/api-server-pg/2026.06.02T21.37.38.add-oidc-sso.sql
```diff
@@ -0,0 +1,21 @@
+ -- Add OIDC columns
+ALTER TABLE public.org_settings
+ADD COLUMN IF NOT EXISTS oidc_enabled boolean DEFAULT false NOT NULL,
+ADD COLUMN IF NOT EXISTS issuer_url character varying(255),
```
ThisIsMissEm: `issuer_url` and `client_id` can be longer than `255` characters, make sure this isn't truncating.

docs/DEVELOPMENT.md
ThisIsMissEm, line 1: Not sure why changes to `docs/` are in this PR, these look general rather than SSO related.

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -51,6 +51,18 @@ export async function verifyEmailPasswordCredentials(
       });
     }
 
+    const oidcSettings = await deps.orgSettingsService.getOidcSettings(
+      user.orgId,
+    );
+
+    if (oidcSettings?.oidc_enabled) {
```
ThisIsMissEm, line 58: Isn't requiring SSO separate from allowing Password? e.g., you could support SAML, OIDC and Password, but upgrade to a preferred method.

server/graphql/datasources/userKyselyPersistence.ts
```diff
@@ -361,6 +361,26 @@ export async function kyselyUserUpdate(
...
 
+export async function kyselyUserUpdateLoginMethods(
+  db: UsersDb,
+  userId: string,
+  loginMethods: readonly LoginMethod[],
+): Promise<GraphQLUserParent | undefined> {
+  const row = await db
+    .updateTable('public.users')
+    .set({
+      login_methods: [...loginMethods],
```
ThisIsMissEm: Why are you don't a spread here? isn't `LoginMethod[]` already essentially a `string[]`?

server/graphql/modules/authentication.ts
```diff
@@ -68,6 +76,23 @@ const Query: ResolverMap = {
...
+  },
+  async getSSOCallbackUrls(_: unknown, { orgId }, context) {
+    const user = context.getUser();
+    if (user == null || user.orgId !== orgId) {
+      throw unauthenticatedError('Authenticated user required');
+    }
+    if (!user.getPermissions().includes('MANAGE_ORG')) {
+      throw forbiddenError('User does not have permission to view SSO settings');
+    }
+    return {
```
ThisIsMissEm, line 93: This return type is a little odd. I'd expect:
```
return {
  type: 'saml' | 'oidc',
  issuer: string,
  redirectUrl: string
}
```

And probably have SSOService expose a `getConfiguration` or something, which internally figures out whether SSO or none is used.

server/graphql/modules/org.ts
```diff
@@ -18,6 +21,29 @@ import { type Context } from '../resolvers.js';
...
+) {
+  const user = context.getUser();
+  if (user == null || user.orgId !== org.id) {
+    throw unauthenticatedError('Authenticated user required');
+  }
+  if (!user.getPermissions().includes('MANAGE_ORG')) {
+    throw forbiddenError(
+      'User does not have permission to manage SSO settings',
+    );
+  }
```
ThisIsMissEm, line 47: This authorization block is repeated here and in `authentication.ts` consider making a `assertUserCanManageOrg(user?)` function.

server/graphql/modules/org.ts
```diff
@@ -18,6 +21,29 @@ import { type Context } from '../resolvers.js';
...
+  const settings = await context.services.OrgSettingsService.getOidcSettings(
```
ThisIsMissEm, line 48: why's OIDC Settings on `OrgSettingsService` and not `SSOService`?

server/graphql/modules/org.ts
```diff
@@ -146,11 +176,35 @@ const typeDefs = /* GraphQL */ `
...
-  input UpdateSSOCredentialsInput {
+  input UpdateSSOSamlCredentialsInput {
+    samlEnabled: Boolean!
     ssoUrl: String!
     ssoCert: String!
   }
 
+  input UpdateSSOOidcCredentialsInput {
+    oidcEnabled: Boolean!
+    issuerUrl: String!
```
ThisIsMissEm, line 179: As mentioned above, you probably want to do a `discoveryUrl` for OIDC, and then discover the issuer server-side.

server/graphql/modules/org.ts
```diff
@@ -146,11 +176,35 @@ const typeDefs = /* GraphQL */ `
...
+  input SwitchSSOMethodInput {
+    method: SSOMethod!
+    # SAML fields (required when method = SAML)
+    ssoUrl: String
+    ssoCert: String
+    # OIDC fields (required when method = OIDC)
+    issuerUrl: String
+    clientId: String
+    clientSecret: String
+  }
```
ThisIsMissEm: You probably want `@oneOf` here:
```suggestion
  input SwitchSSOMethodInput @oneOf {
    oidc: OIDCConfigInput
    saml: SAMLConfigInput
    disabled: Boolean
  }
  
  input OIDCConfigInput {
    issuer: String!
    clientId: String!
    clientSecret: String!
  }
  
  input SAMLConfigInput {
    ssoUrl: String!
    ssoCert: String!
  }
```

Which will only allow submission with either `oidc`, `saml` or `disabled` keys, not both.

You could potentially do:
```graphql
enum SSODisabledMarker { DISABLED }

input SwitchSSOMethodInput @oneOf {
  oidc: OIDCConfigInput
  saml: SAMLConfigInput
  disabled: SSODisabledMarker
}
```

If you want to prevent `disabled: false` as input.

server/graphql/modules/org.ts
```diff
@@ -174,7 +228,9 @@ const typeDefs = /* GraphQL */ `
     setOrgDefaultSafetySettings(
       orgDefaultSafetySettings: ModeratorSafetySettingsInput!
     ): SetModeratorSafetySettingsSuccessResponse
-    updateSSOCredentials(input: UpdateSSOCredentialsInput!): Boolean!
+    updateSSOSamlCredentials(input: UpdateSSOSamlCredentialsInput!): Boolean!
+    updateSSOOidcCredentials(input: UpdateSSOOidcCredentialsInput!): Boolean!
```
ThisIsMissEm: These can probably be collapsed into one mutation `updateSSOSettings` by using that `@oneOf` trick above, and should probably return the updated settings for SSO (since the server validates)

server/graphql/modules/org.ts
```diff
@@ -174,7 +228,9 @@ const typeDefs = /* GraphQL */ `
...
+    switchSSOMethod(input: SwitchSSOMethodInput!): Org!
```
ThisIsMissEm: What's the difference between this and the update mutations above?

server/graphql/modules/org.ts
```diff
@@ -63,8 +89,12 @@ const typeDefs = /* GraphQL */ `
     userStrikeThresholds: [UserStrikeThreshold!]!
     userStrikeTTL: Int!
     isDemoOrg: Boolean!
+    samlEnabled: Boolean!
     ssoUrl: String
     ssoCert: String
+    oidcEnabled: Boolean!
+    issuerUrl: String
+    clientId: String
```
ThisIsMissEm, line 120:
```suggestion
    sso: SSOSettings
```

Would be cleaner and keep all these options together. It can be a union of `SAMLSettings` and `OIDCSettings` discriminated by `type`, or an array of those options, if you want to retain the previous settings.

server/graphql/modules/org.ts
```diff
@@ -789,24 +863,102 @@ const Mutation: GQLMutationResolvers = {
     });
     return gqlSuccessResult({}, 'UpdateUserStrikeTTLSuccessResponse');
   },
-  async updateSSOCredentials(_, { input }, context) {
+  async updateSSOSamlCredentials(_, { input }, context) {
```
ThisIsMissEm: As mentioned above, I'd recommend a single mutation, not two code paths.

server/graphql/modules/org.ts
```diff
@@ -789,24 +863,102 @@ const Mutation: GQLMutationResolvers = {
...
     const user = context.getUser();
     if (!user) {
       throw unauthenticatedError('User required.');
     }
-
-    if (!user.getPermissions().includes(UserPermission.MANAGE_ORG)) {
+    if (!user.getPermissions().includes('MANAGE_ORG')) {
```
ThisIsMissEm:
```suggestion
    if (!user.getPermissions().includes(UserPermission.MANAGE_ORG)) {
```

Appears to be what should be being used here, based on current code in main.

server/graphql/modules/org.ts
ThisIsMissEm: These mutations also have very little to no input validation, when that really is the server's job.

server/graphql/modules/rule.ts
ThisIsMissEm, line 1: Not sure why this change exists here?

server/iocContainer/index.ts
ThisIsMissEm, line 1: This would be good to split into a tiny focused PR

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+>;
+
+export function registerOidcRoutes(app: Express, deps: OidcDeps) {
+  const { KyselyPg } = deps;
+
+  app.get('/oidc/login/callback', async (req, res, next) => {
+    if (req.query.error) {
+      return res.redirect(
+        `${deps.ConfigService.uiUrl}/login/sso?error=${encodeURIComponent(String(req.query.error))}`,
+      );
```
ThisIsMissEm:
```suggestion
      const redirect = new URL(`/login/sso`, deps.ConfigService.uiUrl)
      redirect.search = new URLSearchParams({
        error: req.query.error
        error_descriptions: req.query.error_description
      }).toString()
      
      return res.redirect(redirect);
```

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+
+type OidcDeps = Pick<
+  Dependencies,
+  'OrgSettingsService' | 'SSOService' | 'ConfigService' | 'KyselyPg'
+>;
+
+export function registerOidcRoutes(app: Express, deps: OidcDeps) {
+  const { KyselyPg } = deps;
+
+  app.get('/oidc/login/callback', async (req, res, next) => {
```
ThisIsMissEm: This should probably first gate on if SSO via OIDC is enabled or not, and if not return a 404.

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+      );
+
+      const claims = tokens.claims();
+      let email: string | undefined = claims?.email as string | undefined;
+
+      if (!email) {
+        const userinfo = await oidcClient.fetchUserInfo(
+          config,
+          tokens.access_token,
+          claims!.sub,
```
ThisIsMissEm: I think this should be:
```suggestion
          claims?.sub,
```
Since your if condition isn't gating on there being a `sub`

Better would be:
```
if (!email && !!claims?.sub) {
  // ...
}
```

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+        );
+        email = userinfo.email;
```
ThisIsMissEm: perhaps: 
```suggestion
        // userinfo.email can be undefined here, hence the gate on `!email` below:
        // https://github.com/panva/oauth4webapi/blob/main/src/index.ts#L2987
        email = userinfo.email;
```

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+      }
+
+      if (!email) {
```
ThisIsMissEm, line 148:
```suggestion
      if (typeof email !== "string") {
```
This will correctly type-narrow and avoid the `String(email)` below

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+      if (!user.loginMethods.includes('oidc')) {
+        const updatedUser = await kyselyUserUpdateLoginMethods(
+          KyselyPg,
+          user.id,
+          [...user.loginMethods, 'oidc'],
+        );
+        if (updatedUser != null) {
+          loginUser = updatedUser;
+        }
+      }
```
ThisIsMissEm: This should probably only happen if the user is linking SSO to their account, not if they're attempting to login, otherwise you can get nasty attacks for account takeovers.

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+
+      req.login(loginUser, (err) => {
+        if (err) return next(err);
+        res.redirect(`${deps.ConfigService.uiUrl}/dashboard`);
+      });
+    } catch (e) {
```
ThisIsMissEm, line 178: Perhaps log the error here too?

server/routes/oidcRoutes.ts
ThisIsMissEm: You could also make this `/oidc/:orgId/callback` instead, allowing you to retrieve the org from the URL instead of the session.

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+      return res.redirect(
+        `${deps.ConfigService.uiUrl}/login/sso?error=sso_login_failed`,
+      );
+    }
+  });
+
+  app.post('/oidc/login/:orgId', async (req, res, next) => {
```
ThisIsMissEm: I'd invert this:
```suggestion
  app.post('/oidc/:orgId/login', async (req, res, next) => {
```

Though I'd usually call this `start` instead of `login` because login and signup are the "same"

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+    try {
+      const { orgId } = req.params;
```
ThisIsMissEm, line 197: This should probably have input validation to validate it's an `orgId`, as `req.params.orgId = 'abc'` or `req.params.orgId = 123` would both be valid here.

server/routes/oidcRoutes.ts
```diff
@@ -0,0 +1,204 @@
...
+        // eslint-disable-next-line no-console
+        console.error('[OIDC] Missing credentials for org', orgId, {
+          hasClientId: Boolean(oidcSettings.client_id),
+          hasSecret: Boolean(oidcSettings.client_secret),
+          hasIssuer: Boolean(oidcSettings.issuer_url),
+        });
+        return res.redirect(
+          `${deps.ConfigService.uiUrl}/login/sso?error=oidc_misconfigured`,
+        );
+      }
```
ThisIsMissEm: These should probably both return an opaque error, and use proper URL construction, rather than string based URL construction. `getOidcSettings` should probably contain this logic to validate instead of doing it inline here.

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+            callbackUrl: deps.SSOService.getSSOSamlCallbackUrl(orgId),
+            issuer: deps.SSOService.getSSOSamlIssuer(),
+          });
+        },
+      },
+      async (_req, profile, done) => {
+        try {
+          const user = await kyselyUserFindByEmail(
+            KyselyPg,
+            String(profile?.email),
```
ThisIsMissEm: What happens if `profile.email` is undefined? Shouldn't we validate the input before making a query?

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+          if (!user.loginMethods.includes('saml')) {
+            const updatedUser = await kyselyUserUpdateLoginMethods(
+              KyselyPg,
+              user.id,
+              [...user.loginMethods, 'saml'],
+            );
+            if (updatedUser != null) {
+              return done(null, updatedUser);
+            }
+          }
```
ThisIsMissEm: ditto on the upgrade path should be explicit rather than implicit, as to prevent account takeover.

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+              shouldErrorSpan: true,
+            }),
+          );
+        }
+      },
+      async (_req, profile, done) => {
+        try {
+          const user = await kyselyUserFindByEmail(
+            KyselyPg,
+            String(profile?.email),
```
ThisIsMissEm: same issue as above.

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+          return done(null, user);
+        } catch (e) {
+          return done(
+            makeInternalServerError('Unknown error during login attempt', {
+              shouldErrorSpan: true,
+            }),
+          );
+        }
+      },
+      async (_req, profile, done) => {
```
ThisIsMissEm: Why's this function repeated? Turns out it's because this is a logout path.

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+          if (!user.loginMethods.includes('saml')) {
+            const updatedUser = await kyselyUserUpdateLoginMethods(
+              KyselyPg,
+              user.id,
+              [...user.loginMethods, 'saml'],
+            );
+            if (updatedUser != null) {
+              return done(null, updatedUser);
+            }
+          }
```
ThisIsMissEm:
```suggestion
```

This is a logout pathway, it definitely shouldn't be mutating the user: https://npmx.dev/package/@node-saml/passport-saml#user-content-configure-strategy-for-multiple-providers

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+    ),
+  );
+
+  app.get(
+    '/saml/login/:orgId',
+    passport.authenticate('saml', { failureRedirect: '/', failureFlash: true }),
+  );
+
+  app.post(
+    `/saml/login/:orgId/callback`,
```
ThisIsMissEm, line 101: ditto on URLs here:
```suggestion
  app.get(
    '/saml/:orgId/login',
    passport.authenticate('saml', { failureRedirect: '/', failureFlash: true }),
  );

  app.post(
    `/saml/:orgId/callback`,
```

For consistency with OIDC

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+            }),
+          );
+        }
+      },
+    ),
+  );
+
+  app.get(
+    '/saml/login/:orgId',
+    passport.authenticate('saml', { failureRedirect: '/', failureFlash: true }),
```
ThisIsMissEm: this would redirect you to the API server's `/` not the UI.

server/routes/samlRoutes.ts
```diff
@@ -0,0 +1,155 @@
...
+  );
+
+  app.post(
+    `/saml/login/:orgId/callback`,
+    express.urlencoded(),
+    passport.authenticate('saml', {
+      failureRedirect: '/',
```
ThisIsMissEm: Ditto here, it'd redirect to API base not UI base.

server/services/orgSettingsService/orgSettingsService.ts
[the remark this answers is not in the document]
ThisIsMissEm: it would be better to use a discriminated union here, or a separate object. That way you either have:
```
sso: { type: 'disabled' } | { type: 'saml', ... } | { type: 'oidc', ... }
```

and each type has it's own specific properties that are required.

server/services/orgSettingsService/orgSettingsService.ts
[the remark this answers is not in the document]
ThisIsMissEm: Yes, you should, debatable if it should be in the controller (the mutation) or the service.

server/services/orgSettingsService/orgSettingsService.ts
```diff
@@ -175,20 +177,90 @@ function makeOrgSettingsService(pgQuery: Kysely<OrgSettingsPg>) {
...
           .executeTakeFirst();
         return true;
       } catch (e) {
         return false;
       }
     },
+    async getOidcSettings(orgId: string) {
```
ThisIsMissEm, line 229: Should this return `OIDCSettings | null` where `null` is returned if `oidc_enabled` isn't true?

ThisIsMissEm (changes requested):
I'd recommend a single path way for updating the settings, via the `@oneOf` graphql input path. The normalization logic here also isn't correct since it's not actually doing discovery here to fetch the `issuer` from the metadata document, it's just doing string normalization.

serendipty01:
making the PR as draft as i haven't tested yet and changes are incomplete

server/services/SSOService/ssoOidcService.ts
```diff
@@ -0,0 +1,31 @@
...
+export async function discoverOidcConfig(
+  issuerUrl: string,
+  clientId: string,
+  clientSecret: string,
+) {
+  const config = await oidcClient.discovery(
+    new URL(issuerUrl),
+    clientId,
+    clientSecret,
+  );
```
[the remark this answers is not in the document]
serendipty01: will raise a separate PR for this

client/src/utils/oidc.test.ts
serendipty01: fixed

client/src/webpages/settings/SSOSettings.tsx
serendipty01: done

client/src/webpages/auth/ssoRedirect.ts
serendipty01: moved to server side

server/graphql/modules/org.ts
serendipty01: done

server/graphql/modules/org.ts
serendipty01: fixed

server/graphql/modules/org.ts
serendipty01: fixed

server/graphql/modules/org.ts
serendipty01: will do in a followup PR

client/src/webpages/settings/SSOSettings.tsx
serendipty01: will take in a followup PR

client/src/webpages/settings/SSOSettings.tsx
serendipty01: will do in a followup PR

docs/DEVELOPMENT.md
serendipty01: removed, accidentally got commited

server/graphql/datasources/userKyselyPersistence.ts
serendipty01: done

server/graphql/modules/org.ts
serendipty01: fixed

server/routes/oidcRoutes.ts
serendipty01: done

server/routes/oidcRoutes.ts
serendipty01: done

server/routes/oidcRoutes.ts
serendipty01: done

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/samlRoutes.ts
serendipty01: will raise a separate PR, since this will be a breaking change

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/samlRoutes.ts
serendipty01: fixed

server/routes/oidcRoutes.ts
serendipty01: fixed

server/routes/oidcRoutes.ts
serendipty01: fixed

server/routes/oidcRoutes.ts
serendipty01: fixed

server/routes/oidcRoutes.ts
serendipty01: done, and added redirect to /login/sso?error=sso_login_failed rather than returning a 404

server/routes/oidcRoutes.ts
serendipty01: done

server/routes/oidcRoutes.ts
serendipty01: done

server/routes/oidcRoutes.ts
serendipty01: done

db/src/scripts/api-server-pg/2026.06.02T21.37.38.add-oidc-sso.sql
serendipty01: made them TEXT

PR #91, opened 2026-02-28
Integration Plugin Backend: migration, GraphQL API, and plugin signals

juanmrad:

Adds persistence and API for plugin integrations (migration + signal auth), exposes them via GraphQL (availableIntegrations, integrationConfigs, setPluginIntegrationConfig, signal logo URLs), and wires plugin signal loading (loadPluginSignals, PluginSignalAdapter) so plugin signals are available to the rule engine.

## Changes

- **Migration:** `integration_configs` table for per-org plugin config to allow any integration to store credential/configs in (JSONB) to be used on settings page.
- **Signal auth:** Read/write plugin config in `signalAuthService` / dbTypes.
- **GraphQL:** integrationManifests datasource; IntegrationApi uses registry + new config; availableIntegrations, org.integrationConfigs (incl. plugin), setPluginIntegrationConfig; signal resolvers for integrationLogoUrl / integrationLogoWithBackgroundUrl; rule/insights updates for plugin signals.
- **Plugin signals:** loadPluginSignals, PluginSignalAdapter, credential getters, SignalType/SignalId updates, SignalsService wiring.
- **Conditions/rules:** condition evaluator and rule types updated so rules can use plugin signals.

### DIFF
--- a/server/graphql/modules/integration.ts
+++ b/server/graphql/modules/integration.ts
@@ -1,14 +1,15 @@
 import { AuthenticationError } from 'apollo-server-express';
 
-import { isConfigurableIntegration } from '../../services/signalAuthService/index.js';
+import { getIntegrationRegistry } from '../../services/integrationRegistry/index.js';
 import { Integration } from '../../services/signalsService/index.js';
 import { isCoopErrorOfType } from '../../utils/errors.js';
-import { assertUnreachable } from '../../utils/misc.js';
 import {
   makeIntegrationConfigUnsupportedIntegrationError,
-  type TIntegrationCredential,
 } from '../datasources/IntegrationApi.js';
+import type { TIntegrationConfigWithMetadata } from '../datasources/IntegrationApi.js';
 import {
+  type GQLIntegrationConfig,
+  type GQLIntegrationMetadata,
   type GQLMutationResolvers,
   type GQLQueryResolvers,
 } from '../generated.js';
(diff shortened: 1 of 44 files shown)


server/services/signalsService/helpers/makeCachedCredentialsGetters.ts
```diff
@@ -21,11 +27,40 @@ export function makeCachedCredentialGetters(
...
+      cacheByIntegrationId.set(integrationId, c);
+    }
+    return c;
+  }
+
+  async function close(): Promise<void> {
+    const builtIn = [
+      getApiCredentialForIntegration('GOOGLE_CONTENT_SAFETY_API'),
+      getApiCredentialForIntegration('OPEN_AI'),
+      getApiCredentialForIntegration('ZENTROPI'),
```
vinaysrao1: My reading of cached() in caching.ts, creates a new cache. So close() calling getApiCredentialForIntegration is not closing existing cache but creating new ones can closing them.

server/graphql/modules/integration.ts
```diff
@@ -139,19 +185,28 @@ const typeDefs = /* GraphQL */ `
...
+      input: SetPluginIntegrationConfigInput!
+    ): SetIntegrationConfigResponse!
   }
 `;
 
-const IntegrationApiCredential: ResolverMap<TIntegrationCredential> = {
+const IntegrationApiCredential: ResolverMap<TIntegrationConfigWithMetadata['apiCredential']> = {
   __resolveType(it) {
-    const integrationName = it.name;
+    const integrationName = (it as { name?: string }).name ?? '';
```
vinaysrao1: "it" is apiCredential type. the changed mergeManifest, does not add a name attribute to apiCredential (the old code did) in IntegrationApi.ts

server/services/signalsService/SignalsService.ts
```diff
@@ -134,12 +152,17 @@ export class SignalsService {
       this.hmaService,
     );
 
+    const pluginEntries = getIntegrationRegistry().getPluginEntries();
+    const pluginSignals = loadPluginSignals(pluginEntries, signalAuthService);
```
vinaysrao1: Should we check for plugin signals name collision with builtin signals. Rare event, but worth checking

vinaysrao1 (changes requested):

server/services/signalsService/SignalsService.ts
juanmrad: Agreed. I’ve added a check before merging plugin signals: we build a set of built-in signal type ids and throw if any plugin `signalTypeId` is in that set, with a message like: `Plugin signal type "<id>" collides with a built-in signal; use a different signalTypeId`. So we fail fast at startup if a plugin reuses a built-in id.

server/graphql/modules/integration.ts
juanmrad: You’re right fixed. now passes an `apiCredential` object that includes the integration id so the union resolver can discriminate: `apiCredential: { ...apiCredential, name: integrationId }`.

vinaysrao1 (approved):

PR #106, opened 2026-03-05
Add typeName to webhook return

calebmcquaid:

There was a request by a client to add a `typeName` return in the webhook response. This implements that request.

## Tests

I added the necessary setup steps: Add a content type, rules, and policies. Added webhook.site to the rules to see the output. Curled a post submission to trigger the rule and see the outcome.

Screenshot 2026-03-05 at 3 19 57 PM
Screenshot 2026-03-05 at 3 19 45 PM
Screenshot 2026-03-05 at 3 16 44 PM

### DIFF
--- a/server/rule_engine/ActionPublisher.ts
+++ b/server/rule_engine/ActionPublisher.ts
@@ -79,7 +79,7 @@ export function getUserFromActionTarget(it: ActionTargetItem) {
  */
 export type ActionTargetItem =
   | ItemSubmission
-  | { itemId: string; itemType: Pick<ItemType, 'id' | 'kind'> & Partial<Pick<ItemType, 'name'>> };
+  | { itemId: string; itemType: Pick<ItemType, 'id' | 'kind' | 'name'> };
 
 export function isFullSubmission(
   input: ActionTargetItem,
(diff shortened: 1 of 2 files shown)


server/rule_engine/ActionPublisher.ts
```diff
@@ -79,7 +79,7 @@ export function getUserFromActionTarget(it: ActionTargetItem) {
  */
 export type ActionTargetItem =
   | ItemSubmission
-  | { itemId: string; itemType: Pick<ItemType, 'id' | 'kind'> };
+  | { itemId: string; itemType: Pick<ItemType, 'id' | 'kind'> & Partial<Pick<ItemType, 'name'>> };
```
juanmrad: Since typeName was a client requirement for the webhook response, name should be required, not optional.

This ensures typeName is always present in webhook payloads. If there are code paths that construct

ActionTargetItem without name, we should update it to fetch the full ItemType or pass name explicitly.

calebmcquaid: Good call, updated!

juanmrad (approved):

PR #124, opened 2026-03-18
[Code Simplification] Remove use of body-parser for native express json

juanmrad:

Part of code simplification this removes the body parser package as we have express and there is an existing way to parse json. 

## Tests

tested locally.


julietshen (approved):
woohoo, thanks juan!

PR #133, opened 2026-03-20
[Code Simplification] Removal of Snowflake from codebase

juanmrad:

This PR removes all references to snowflake from our codebase. It cleans up the snowflake code, the packages as well as its references. It moves any snowflake related types to common data warehouse types instead.


vinaysrao1 (approved):
Awesome work. Can't find anything.

PR #135, opened 2026-03-21
Add repository field to published packages for npm provenance

juanmrad:

As seen on https://github.com/roostorg/coop/actions/runs/23366011303/job/67979984974 package publishing failed due to missing repository field.

This pr adds the repository field to `types/package.json` and `migrator/package.json` to fix npm publish failures caused by sigstore provenance verification


vinaysrao1 (approved):
Approving, but please check the following before merging: do we need to remove any code (like in the manifests file for model cards) because we can now import from the types package?

PR #141, opened 2026-03-24
[Code Simplification] Remove legacy `.devops` infrastructure and move migrations to `db/`

juanmrad:

As discussed there is no current benefit to the code that lived in `.devops` aside from the migration scripts. As such we should remove legacy infrastructure code from `.devops/` (AWS CDK, CDKTF, Pulumi, Helm charts) that we no longer maintain. Each adopter most than likely will have their own way to deploy to their infra.

Moved the migrator scripts to a more appropriate `db` folder to remove the connotation of "devops"

Added a a Deployment section to the README pointing to the [`0.1` branch](https://github.com/roostorg/coop/tree/0.1/.devops) for historical AWS deployment reference in case it is useful for adopters.


vinaysrao1 (approved):
Woohoo! Massive clean up. Thank you!

PR #151, opened 2026-03-25
Fix ClickHouse outage crashing all dashboard pages

### LINKED ISSUE #145: If Clickhouse is down, it brings down all of Coop
`allRuleInsights` query in the Overview/Dashboard runs on every page load and throws `ECONNREFUSED` if ClickHouse is unavailable. This should fail gracefully (return empty data) instead of crashing the Apollo client. Every page becomes unusable even though ClickHouse is only needed for analytics.

Image

juanmrad:

Fixes #145. When ClickHouse is down, every dashboard page crashed because analytics query errors propagated uncaught through resolvers and client components.

Added better handling of the clickhouse error. Also added a lightweight checker for health on clickhouse so we can properly show that the db is down.

Open to suggestions on the error messages there.

Screenshot 2026-03-24 at 8 37 29 PM

### DIFF
--- a/server/graphql/resolvers.ts
+++ b/server/graphql/resolvers.ts
@@ -98,8 +98,25 @@ const Query: GQLQueryResolvers = {
       return null;
     }
 
-    // TODO: this response type actually isn't right; remove cast and fix errors.
-    return context.dataSources.ruleAPI.getAllRuleInsights(user.orgId) as any;
+    try {
+      // TODO: this response type actually isn't right; remove cast and fix errors.
+      return (await context.dataSources.ruleAPI.getAllRuleInsights(
+        user.orgId,
+      )) as any;
+    } catch {
+      return null;
+    }
+  },
+  async isWarehouseAvailable(_, __, context) {
+    try {
+      await context.services.DataWarehouse.query(
+        'SELECT 1',
+        context.services.Tracer,
+      );
+      return true;
+    } catch {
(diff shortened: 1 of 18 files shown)


server/graphql/resolvers.ts
```diff
@@ -98,8 +98,25 @@ const Query: GQLQueryResolvers = {
...
 
-    // TODO: this response type actually isn't right; remove cast and fix errors.
-    return context.dataSources.ruleAPI.getAllRuleInsights(user.orgId) as any;
+    try {
+      // TODO: this response type actually isn't right; remove cast and fix errors.
+      return (await context.dataSources.ruleAPI.getAllRuleInsights(
+        user.orgId,
+      )) as any;
+    } catch {
+      return null;
```
vinaysrao1: empty catch?

server/graphql/resolvers.ts
```diff
@@ -98,8 +98,25 @@ const Query: GQLQueryResolvers = {
...
+  },
+  async isWarehouseAvailable(_, __, context) {
+    try {
+      await context.services.DataWarehouse.query(
+        'SELECT 1',
+        context.services.Tracer,
+      );
+      return true;
+    } catch {
+      return false;
```
vinaysrao1: Same as above

vinaysrao1 (approved):
nits about try - catch.

PR #160, opened 2026-03-27
Fix unbounded queries crashing review queues

### LINKED ISSUE #161: Fix Unbounded queries on MRT queue

juanmrad:

Fixes #161 

Fix unbounded MRT queries causing dashboard crashes under high queue depth. Add `.limit(1000)` to `getSkippedJobsForRecentDecisions` to prevent unbounded Postgres result sets.

Eliminate N+1 in `getTotalPendingJobsCount` by skipping redundant per-queue `checkQueueExists` Postgres calls when queues are already loaded

## Tests

Created a script to load up >50k reports to test. Video shows UI still working snappy as expected
[Loom video](https://www.loom.com/share/05954275ff2347dd80539458a288583b)


server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -798,15 +798,21 @@ export default class QueueOperations {
...
+    orgId: string;
+    queueId: string;
+    limit?: number;
+  }) {
+    const concurrencyLimit = pLimit(10);
     const { orgId, queueId } = opts;
     const queue = await this.#getBullQueue(orgId, queueId);
-    // Get up to 50 jobs per queue.
-    const legacyJobs = await queue.getJobs(undefined, 0, 50);
+    const maxJobs = Math.min(opts.limit ?? 50, 50);
```
vinaysrao1: This is capped to 50 because it is preview, right? Should we set a check for negative numbers?

vinaysrao1 (changes requested):

vinaysrao1 (approved):
Sorry, I missed your comments earlier

PR #172, opened 2026-03-31
docs: Clean up redundant NCMEC docs

cassidyjames:
There was a lot of duplicated content between the (already very long) user guide and the dedicated NCMEC docs page. Instead, let's summarize it for the user guide, then link to the dedicated docs page where relevant.

It also looks like the NCMEC reporting docs previously got put in the middle of the user roles section; instead, let's move it below so it makes a little more sense. :)

Rendered: https://github.com/roostorg/coop/blob/cassidyjames/docs-ncmec/docs/USER_GUIDE.md

---

Follow-up to #169, as mentioned in https://github.com/roostorg/coop/issues/169#issuecomment-4165634471

### DIFF
--- a/docs/NCMEC.md
+++ b/docs/NCMEC.md
@@ -41,7 +41,7 @@ If you want to automatically detect known CSAM via hash matching, you need a **s
 
 ## NCMEC Settings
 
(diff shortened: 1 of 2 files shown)


cassidyjames:
This will basically become obsolete with the larger rework of the docs, but I want to get it in as an iterative improvement.

PR #173, opened 2026-04-01
chore(content-proxy): use native fetch

43081j:

We can switch away from `axios` to using native fetch since we don't use
any of the extra feature axios provides.

This reduces the dependencies a little.

Note: I just saw this while roaming around the repo. if you're not interested in PRs like this, let me know and I will hold off 👍  

## Tests

Existing tests should cover this change, but I did run this locally and ensure images and what not still loaded.

### DIFF
--- a/content-proxy/package-lock.json
+++ b/content-proxy/package-lock.json
@@ -10,7 +10,6 @@
       "license": "ISC",
       "dependencies": {
         "@babel/preset-env": "^7.24.5",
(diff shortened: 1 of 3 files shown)


cassidyjames:
@43081j thank you for the PR! We're discussing this right now; it makes sense to move to native fetch, but we may want to keep the package overrides due to transitive dependencies.

PR #175, opened 2026-04-04
[dependabot] Configure to prevent major bump of node

juanmrad:

Configure dependabot docker updates on all directories. also prevent major version upgrades. This will not do LTS->LTS upgrades, but that is ok as major upgrades would require some manual testing. while minor patches bring security upgrades.


.github/dependabot.yml
```diff
@@ -7,6 +7,14 @@ version: 2
...
+      - "/db"
+      - "/hma"
+      - "/nodejs-instrumentation"
     schedule:
-      interval: "weekly"
\ No newline at end of file
+      interval: "weekly"
+    ignore:
+      - dependency-name: "node"
+        update-types: ["version-update:semver-major"]
```
vinaysrao1: nit, needs a newline

vinaysrao1 (approved):

PR #176, opened 2026-04-04
[Simplification] Remove Content Proxy as not needed reference

juanmrad:

As discussed the content proxy was a reference code on how to create a proxy for content, but this is not something we want to maintain as any content proxy deployment would work for this purpose. Users can still provide their own content proxy for iframe content and is configured via `VITE_CONTENT_PROXY_URL` for coop to use.


vinaysrao1 (approved):
Perhaps add a comment in .env.example for VITE_CONTENT_PROXY_URL

PR #216, opened 2026-04-14
[Vulnerabilities] Remove lodash by migrating from better mutation to plugin-functional and fix minor lint warnings

### LINKED ISSUE #185: Migrate from `eslint-plugin-better-mutation` to `eslint-plugin-functional`
We currently use `eslint-plugin-better-mutation` for its `no-mutation` rule to discourage direct object mutation in server code. While the plugin works, it has not been updated in a while, and contains a vulnerable `lodash` dependency.

I propose moving to [`eslint-plugin-functional`](https://github.com/eslint-functional/eslint-plugin-functional) which provides a comparable `functional/immutable-data` rule and is a well-maintained alternative. It also has no `lodash` dependency and it is more lightweight.

### Migration steps
Install `eslint-plugin-functional` and remove `eslint-plugin-better-mutation`
Replace the `better-mutation/no-mutation` rule with `functional/immutable-data` in `.eslintrc.cjs`, mapping the `allowThis: true` option to the equivalent configuration
Update all `eslint-disable better-mutation/no-mutation` comments (~25 test files) to reference the new rule name or determine if we could change bahavior in code to not mutate objects either doing a deep copy or shallow copy. 
Verify lint passes and no new violations are introduced.

juanmrad:

Fixes #185

Changes from `eslint-plugin-better-mutation` to `eslint-plugin-functional` and updated all comments as well as ran the `npm lin fix` to reduce amount of warnings. There will be a follow up pr to help reduce lint warnings further. 

## Tests

ran `npm run lint` and warnings have gone down.


server/routes/integration_logos/serveIntegrationLogo.ts
```diff
@@ -29,7 +30,8 @@ export default function serveIntegrationLogo(
     // Path was validated at plugin load (under package root); safe to send.
     res.setHeader('Cache-Control', 'public, max-age=86400');
     res.sendFile(filePath, (err) => {
-      if (err != null && !res.headersSent) { // eslint-disable-line @typescript-eslint/no-unnecessary-condition -- sendFile callback err is Error | null per types
+      if (err != null && !res.headersSent) {
```
vinaysrao1, line 33: When I ran "npx eslint" I get a lot (82) of warnings. 

Three of them are because of this PR removing "// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition"

It might be fine. Flagging this.

vinaysrao1 (approved):
2 nits:

1. this is removing the transitive dependency of lodash - not the lodash package itself, right?
2. removing // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition flag

server/routes/integration_logos/serveIntegrationLogo.ts
juanmrad: Yes. we still have a lot of warnings. This PR mostly took care of the low hanging ones.

PR #217, opened 2026-04-14
Override axios with v1.15.0+ which hardened its CI/NPM publishing process

### LINKED ISSUE #201: Fully remove axios
Transitive dependencies: Check if googlemaps, sendgrid have an update that removes axios

Right now we have overrides to pin specific versions; @pawiecz will update to latest-available and verify that tests pass.

pawiecz:

Fixes: #201

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -162,6 +162,10 @@
     "@types/express": "$@types/express",
     "@googlemaps/google-maps-services-js@^3.3.16": {
       "retry-axios": "npm:@ethanresnick/retry-axios@2.6.1"
-    }
+    },
+    "axios": "^1.15.0"
+  },
+  "resolutions": {
+    "axios": "^1.15.0"
   }
 }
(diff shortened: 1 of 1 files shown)


server/package.json
```diff
@@ -157,14 +157,17 @@
...
     },
     "eslint-plugin-better-mutation": {
       "lodash": "^4.18.1"
-    }
+    },
+    "axios": "^1.15.0"
+  },
+  "resolutions": {
+    "axios": "^1.15.0"
   }
```
juanmrad: I'm unsure what this is changing if anything.

pawiecz: @juanmrad Fair point - the dependency has already been updated and locked on `coop` side. This snippet only blocks any potential transitive dependency that could have locked a version of Axios prior to CI hardening. This way we prevent pulling pre-hardened versions (even if not affected by the supply chain issue)

pawiecz: @juanmrad I'll close this PR and mark #201 as complete if you find this precaution unnecessary - let me know what you think

pawiecz:
Closing stale request: Reviewer expressed doubts about the need for this change in the first place.

PR #225, opened 2026-04-15
[Kysely] migrate rule-engine queries and related jobs to Kysely (phase 1)

juanmrad:

This is phase 1 of the Sequelize -> Kysely migration: we move a vertical slice of the POST / content rule hot path and closely related logic off Sequelize models onto Kysely and `ModerationConfigService`, this is reduced to keep ORM change small piece.

This PR mostly touches IOC, Jobs, Rules Engine injected queries.

There will be a handful of follow up prs after this to continue the migration.


server/services/policyActionPenalties.ts
```diff
@@ -0,0 +1,62 @@
...
+  orgId: string,
+): Promise<PolicyActionPenalties[]> {
+  const [actions, policies] = await Promise.all([
+    moderationConfigService.getActions({ orgId, readFromReplica: true }),
+    moderationConfigService.getPolicies({ orgId, readFromReplica: true }),
+  ]);
+
+  return policies.flatMap((policy) =>
+    actions.map((action) => {
+      const actionPenalty = userPenalty(action as PenaltyRow);
```
vinaysrao1: I might be missing something. We don't have penalty in #dbResultToAction in ActionOperations.ts: 
https://github.com/roostorg/coop/pull/225/changes#diff-4d8001777029ca40dc33c74d613ef1b93363fa8c95df32f4a1c079708c996b40L108

This might cause the casting "as PenaltyRow" in this line to hide this error

vinaysrao1 (changes requested):

server/services/policyActionPenalties.ts
juanmrad: fixed

server/models/rules/ruleTypes.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export enum ConditionCompletionOutcome {
+  PASSED = 'PASSED',
+  FAILED = 'FAILED',
+  INAPPLICABLE = 'INAPPLICABLE',
+}
+
+export enum ConditionFailureOutcome {
+  ERRORED = 'ERRORED',
+}
```
vinaysrao1, line 28: This seems like repeated declaration of enums (RuleModes.ts)

vinaysrao1 (approved):
Nit about repeated enum definition.

PR #226, opened 2026-04-15
[Kysely] Remove knex migrate backtest pagination and takeLast to Kysely

juanmrad:

This change removes the direct `knex` dependency from the server and drops `Knex` from the IOC container replacing its uses with kysely.

Knex was only used as a query builder for backtest pagination while execution already went through the warehouse Kysely client. Consolidating on Kysely reduces dependencies, avoids raw SQL string round-trips, and keeps typing and compilation aligned with the rest of the warehouse stack.


server/graphql/datasources/RuleApi.ts
```diff
@@ -764,17 +757,20 @@ class RuleAPI {
...
+    const results = await finalQuery.execute();
 
-    return results.map((it: any) => ({
-      node: { ...it, result: it.result ? jsonParse(it.result) : null },
-      cursor: { ts: new Date(it.ts).valueOf() },
-    }));
+    return results.map((it) => ({
+      node: {
+        ...it,
+        result: it.result ? jsonParse(it.result as never) : null,
```
vinaysrao1: casting as never, can we just cast to string?

server/graphql/datasources/RuleApi.ts
```diff
@@ -764,17 +757,20 @@ class RuleAPI {
...
+      },
+      cursor: { ts: warehouseDateToDate(it.ts).valueOf() },
+    })) as unknown as Edge<RuleExecutionResult, { ts: number }>[];
```
vinaysrao1: double casting?

vinaysrao1 (approved):
Nits

PR #229, opened 2026-04-16
[Vulnerabilities] Override ajv to patch vulnerability

juanmrad:

This will fix the last vuln alert from dependabot for now. sadly umzug is not in active development so there isn't a way for us to upgrade as such we are doing an override. 

Will explore options to move out of umzug. potentially build out more of the db-migrator package to handle the cases we currently use umzug for.

### DIFF
--- a/db/package-lock.json
+++ b/db/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "coop-data-migrator",
   "version": "1.0.0",
-  "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


pawiecz (approved):
There were no feature breaking changes between previously used 8.13.0 and overridden 8.18.0 releases - 3 of them were fixups (according to [`ajv` changelog](https://github.com/ajv-validator/ajv/releases)).

PR #260, opened 2026-04-17
[Kysely] Migrate MRT action/policy id lookups off Sequelize

juanmrad:

This change moves the cached “load by id list” helpers used by the manual review tool (and shared consumers) from Sequelize models to `ModerationConfigService`, so reads go through the same Kysely-backed path as the rest of moderation config.

### DIFF
--- a/server/services/manualReviewToolService/manualReviewToolQueries.ts
+++ b/server/services/manualReviewToolService/manualReviewToolQueries.ts
@@ -1,5 +1,3 @@
-import { Op } from 'sequelize';
-
 import { inject } from '../../iocContainer/index.js';
 import { cached } from '../../utils/caching.js';
 import { jsonParse, jsonStringify } from '../../utils/encoding.js';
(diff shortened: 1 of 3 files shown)


server/services/manualReviewToolService/manualReviewToolQueries.ts
```diff
@@ -37,12 +36,19 @@ export type GetActionsByIdEventuallyConsistent = ReturnType<
...
     cached({
+      keyGeneration: {
+        toString: (it: PolicyKey) =>
+          jsonStringify({ ...it, ids: [...it.ids].sort() }),
+        fromString: (it) => jsonParse(it),
+      },
       async producer(key: PolicyKey) {
-        return Policy.findAll({
-          where: { id: { [Op.in]: key.ids }, orgId: key.orgId },
+        return moderationConfigService.getPoliciesByIds({
```
vinaysrao1, line 51: Like in line 20 above for actions, should we return [] if "key.ids.length === 0"?

vinaysrao1 (approved):
Small safety suggestion otherwise LGTM

server/services/manualReviewToolService/manualReviewToolQueries.ts
juanmrad: yes. this would help prevent a db call.

PR #261, opened 2026-04-17
[Kysely] Migrate location banks API out of sequelize

juanmrad:

Moves the GraphQL location bank datasource off Sequelize models and onto KyselyPg (`public.location_banks` / `public.location_bank_locations`).

Was not able to fully move due to other dependencies but that will be taken care of once the org model is moved ( which would be last ).

### DIFF
--- a/server/graphql/datasources/LocationBankApi.ts
+++ b/server/graphql/datasources/LocationBankApi.ts
@@ -1,88 +1,158 @@
 import { type Exception } from '@opentelemetry/api';
+import { type Kysely } from 'kysely';
 import { uid } from 'uid';
 import { v1 as uuidV1 } from 'uuid';
 
 import { inject, type Dependencies } from '../../iocContainer/index.js';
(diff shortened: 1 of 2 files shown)


server/graphql/datasources/LocationBankApi.ts
```diff
@@ -143,10 +237,21 @@ class LocationBankAPI {
     const { id, orgId } = opts;
 
     try {
-      const bank = await this.sequelize.LocationBank.findOne({
-        where: { id, orgId },
+      const result = await this.db.transaction().execute(async (trx) => {
```
vinaysrao1, line 240: Like the old code, should we check for orgId?

vinaysrao1 (changes requested):

server/graphql/datasources/LocationBankApi.ts
juanmrad: Done!

juanmrad:
@vinaysrao1 this is ready for review

vinaysrao1 (approved):

PR #271, opened 2026-04-18
[Kysely] Migrate rule mutations and backtest list/cancel to Kysely

juanmrad:

Phase 2 of the Sequelize → Kysely migration, continuing from #225. This slice moves the Rule create/update/delete mutations and the backtest list / running-check / cancel paths off Sequelize onto Kysely + `ModerationConfigService`. `backtest` and `retroaction` resolvers now fetch rules via the new `ModerationConfigService.getRuleByIdAndOrg`, and the GraphQL Rule parent is built from a plain row via a small helper instead of a Sequelize instance.

Also fixes a latent bug where `expiration_time` was being overwritten on every rule update even when the GraphQL input omitted `expirationTime`. The new update path diffs against the existing row and only writes columns that actually changed.

Added some tests as I was feeling that we did not have enough testing for the feature and to ensure kysely migration didn't break anything.

### DIFF
--- a/server/graphql/datasources/RuleApi.ts
+++ b/server/graphql/datasources/RuleApi.ts
@@ -4,18 +4,14 @@ import { type Exception } from '@opentelemetry/api';
 import { makeEnumLike } from '@roostorg/types';
 
 import { type Kysely } from 'kysely';
-import Sequelize from 'sequelize';
 import { uid } from 'uid';
 
 import { inject, type Dependencies } from '../../iocContainer/index.js';
-import {
-  isEmptyResultSetError,
-  isUniqueConstraintError,
-} from '../../models/errors.js';
 import { type User } from '../../models/UserModel.js';
 import { type ActionCountsInput } from '../../services/actionStatisticsService/index.js';
 import { type AggregationClause } from '../../services/aggregationsService/index.js';
 import { type ConditionSetWithResultAsLogged } from '../../services/analyticsLoggers/index.js';
+import { type CombinedPg } from '../../services/combinedDbTypes.js';
 import {
   RuleType,
   type Condition,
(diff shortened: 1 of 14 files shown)


server/graphql/datasources/ruleKyselyPersistence.ts
```diff
@@ -0,0 +1,399 @@
...
+      creator_id: input.creatorId,
+      expiration_time,
+      condition_set: input.conditionSet,
+      alarm_status: RuleAlarmStatus.INSUFFICIENT_DATA,
+      alarm_status_set_at: now,
+      rule_type: input.ruleType,
+      parent_id: input.parentId ?? null,
+      created_at: now,
+      updated_at: now,
+    } as never)
```
vinaysrao1: This might be risky because it disable the typechecker

server/graphql/datasources/RuleApi.ts
```diff
@@ -67,10 +57,39 @@ import {
 } from '../generated.js';
 import { oneOfInputToTaggedUnion } from '../utils/inputHelpers.js';
 import { type CursorInfo, type Edge } from '../utils/paginationHandler.js';
+import { buildGraphqlRuleParent } from './buildGraphqlRuleParent.js';
+import {
+  isPostgresUniqueViolation,
```
vinaysrao1: there is perhaps an exact implementation of this in utils/kysely.js

https://github.com/roostorg/coop/blob/435d80f37202bd7f0c152163eaefea391cad3d42/server/utils/kysely.ts#L11

all the isPostgresUniqueViolation calls can be replaced with the utils/kysely.js implementation

vinaysrao1 (changes requested):

server/graphql/datasources/ruleKyselyPersistence.ts
juanmrad: Updated it. it was mostly given the created_at updated_at are not typed for kysely, but changed it to add the types version with expansion to ensure we have type safety.

server/graphql/datasources/RuleApi.ts
juanmrad: yes. done

vinaysrao1 (approved):

vinaysrao1 (approved):

PR #275, opened 2026-04-20
[Kysely] Migrate Action mutations and lookups to Kysely

juanmrad:

Continues the `Sequelize` -> `Kysely` migration. 

Moved the Action mutations and the action lookup paths used by `bulkExecuteActions` and the resolvers off Sequelize and onto ModerationConfigService. `ActionApi` no longer touches the DB directly. Resolver mappers in `codegen.yaml` are repointed from ActionModel to the service-layer types.

Fixed a few issues with `custom_mrt_api_params` being silently dropped on kysely path prior. Like `ActionApi.deleteAction` always returned true even when nothing was deleted. Empty `updateCustomAction` patches no longer bump `updated_at`, and `ContentType.actions `.

Also added some tests to cover the gaps found during the migration.


server/services/moderationConfigService/modules/ActionOperations.ts
vinaysrao1, line 354: We should add description here:

https://github.com/roostorg/coop/pull/275/changes#diff-4d8001777029ca40dc33c74d613ef1b93363fa8c95df32f4a1c079708c996b40L21

server/services/moderationConfigService/types/actions.ts
vinaysrao1, line 36: Add "description" here.

server/services/moderationConfigService/modules/ActionOperations.ts
vinaysrao1: I should have found this with the userTypeId checks. I did check for all other discrepancies.

server/services/moderationConfigService/modules/ActionOperations.ts
```diff
@@ -67,40 +86,219 @@ export default class ActionOperations {
...
+  }
+
+  async getActionsForItemType(opts: {
+    orgId: string;
+    itemTypeId: string;
+    itemTypeKind: ItemTypeKind;
+    readFromReplica?: boolean;
+  }) {
+    const { orgId, itemTypeId, itemTypeKind, readFromReplica } = opts;
+    const pgQuery = this.#getPgQuery(readFromReplica ?? true);
```
vinaysrao1: Do we want to read from replica?

vinaysrao1 (changes requested):

server/services/moderationConfigService/modules/ActionOperations.ts
juanmrad: Done

server/services/moderationConfigService/types/actions.ts
juanmrad: Done

server/services/moderationConfigService/modules/ActionOperations.ts
juanmrad: defaulted to false

vinaysrao1 (approved):

PR #283, opened 2026-04-21
[Express] Upgrade to express 5

### LINKED ISSUE #203: Upgrade Express 4 to Express 5
Part of the [Coop Code Simplification Plan](https://github.com/roostorg/coop/discussions/123).

Express 5 has been stable since late 2024 and is a real but manageable migration. Unblocks future simplification work.

juanmrad:
Fixes #203 

Upgrade Express from v4 to v5 (5.2.1), along with the corresponding `@types/express` and Apollo middleware adapter.

updated some of the dependencies. had to add override as `@node-saml/passport-saml` still uses express 4 types. may open a pr on their side to see if we can get that upgraded.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -24,7 +24,7 @@
   "license": "ISC",
   "dependencies": {
     "@apollo/server": "^5.5.0",
-    "@as-integrations/express4": "^1.1.2",
+    "@as-integrations/express5": "^1.1.2",
     "@aws-sdk/client-s3": "^3.1017.0",
     "@aws-sdk/client-secrets-manager": "^3.1017.0",
     "@aws-sdk/client-ses": "^3.1017.0",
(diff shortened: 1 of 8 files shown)


server/package.json
```diff
@@ -156,9 +157,10 @@
     "classNameTemplate": "{classname}",
     "titleTemplate": "{title}"
   },
-  "//": "Use an override to remove the @types/restify dependency, b/c it conflicts w/ pino-http. And a fork of retry-axios that has fixed type defs for TS's new module resolution algorithm.",
+  "//": "Use an override to remove the @types/restify dependency, b/c it conflicts w/ pino-http. And a fork of retry-axios that has fixed type defs for TS's new module resolution algorithm. Also pin @types/express to v5 across the dep tree so transitive deps (e.g. @node-saml/passport-saml) don't drag in v4 types alongside our v5 types.",
```
vinaysrao1, line 160: I am assuming passport-saml has not been updated to use express 5. Is that right?

vinaysrao1 (approved):
Just questions.
One in the codebase and another:
Should we eventually move all @types to devDependencies?

server/package.json
juanmrad: Correct. I was planning to put a pr there to try and get this upgraded, but we'll see. the project does seem to be a bit more stale.

PR #284, opened 2026-04-21
[Bug Fix] Surface Thread-kind items in user submission history

juanmrad:
There was a bug where `Thread` kind items submitted with a `creatorId` schema field role were silently being dropped from creator-keyed surfaces. This was due to `getCreator` in `makeItemSubmission.ts` hard-coded `undefined` for `kind: THREAD`, so every Thread submission was written to Scylla with an empty `item_creator_identifier`. 

As a result the `item_submission_by_creator` materialized view never indexed Thread items by their actual author, and `latestItemsCreatedBy` returned nothing for users whose only submissions were Threads.

This is the root cause for orgs that model their platform as "top-level post = Thread, replies = Content with threadId pointing back at the post": those top-level posts never appeared on the user the user investigation surface.

Note: existing Thread rows in Scylla were written with a nil `item_creator_identifier` and will not be retroactively reindexed; only Thread submissions received after this change will appear in creator-keyed queries.

### DIFF
--- a/client/src/graphql/generated.ts
+++ b/client/src/graphql/generated.ts
@@ -16689,7 +16689,35 @@ export type GQLGetLatestUserSubmittedItemsQuery = {
             };
           };
         }
-      | { readonly __typename: 'ThreadItem' }
+      | {
+          readonly __typename: 'ThreadItem';
+          readonly id: string;
+          readonly submissionId: string;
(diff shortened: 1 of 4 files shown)


vinaysrao1 (approved):
Nice work. Very clean

PR #289, opened 2026-04-22
[Helmet][Simplification][Security] Upgrade Helmet from 4 -> 8

### LINKED ISSUE #204: Upgrade Helmet 4.x to 8.x
Part of the [Coop Code Simplification Plan](https://github.com/roostorg/coop/discussions/123).

Helmet has had four major versions since 4.x, with stricter default CSP/CORP headers in each. Upgrading brings improved security defaults.

juanmrad:
Closes #204

Migrate from Helmet 4 -> 8 for better security. Mostly this is a non code change beyond the integration logos to ensure they can be loaded.

### DIFF
--- a/server/routes/integration_logos/serveIntegrationLogo.ts
+++ b/server/routes/integration_logos/serveIntegrationLogo.ts
@@ -31,6 +31,10 @@ export default function serveIntegrationLogo(
     }
     // Path was validated at plugin load (under package root); safe to send.
     res.setHeader('Cache-Control', 'public, max-age=86400');
+    // Public plugin asset; opt out of helmet's strict same-origin CORP default
+    // so the SPA can load it via <img src> when deployed on a different origin
+    // than the API.
+    res.setHeader('Cross-Origin-Resource-Policy', 'cross-origin');
     res.sendFile(filePath, (err) => {
       if (err != null && !res.headersSent) {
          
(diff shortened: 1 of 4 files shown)


server/routes/integration_logos/serveIntegrationLogo.ts
```diff
@@ -31,6 +31,10 @@ export default function serveIntegrationLogo(
     }
     // Path was validated at plugin load (under package root); safe to send.
     res.setHeader('Cache-Control', 'public, max-age=86400');
+    // Public plugin asset; opt out of helmet's strict same-origin CORP default
+    // so the SPA can load it via <img src> when deployed on a different origin
+    // than the API.
+    res.setHeader('Cross-Origin-Resource-Policy', 'cross-origin');
```
vinaysrao1, line 37: I am guessing the risk here is nil because of the static image, right?

vinaysrao1 (approved):
Just a question.

server/routes/integration_logos/serveIntegrationLogo.ts
juanmrad: Correct. The risk is very little we use `img` tag. We have this when we load from a remote location the image if the plugin does not ship the asset as part of the bundle.

PR #290, opened 2026-04-22
[Kysely] Migrate backtests and retroaction out of sequelize

juanmrad:

Moves createBacktest and runRetroaction in 837ff02081c24091ef911da363f5aa1d7c823586. These two are actually commented out. I did some manual testing to ensure migration work, but given we have them disabled and no way to run on the UI I am keeping them disabled for now.

May be worth having a a more complete e2e test on an env we can create rules to do backtest and retroaction with more data to ensure performance. 

My decision to move the code was to remove references in code for sequelize, and to keep disabled was because of low confidence right now to test. will come back to this after migration.


vinaysrao1 (approved):
Thanks Juan. This makes sense. Should we create an issue to come back to this?

juanmrad:
@vinaysrao1 yes adding a new issue for that.

PR #314, opened 2026-04-24
move CI checks into docker-compose services

vinaysrao1:
## Summary
- Every CI check is now a compose service 
  
## Why                                                                                     
AGENTS.md and docs that describe "how to run CI locally" had started to drift from the 
workflow YAML. With the checks defined in compose, there is one source of truth. 

## What reviewers should look at 
- `docker-compose.yaml` — three new services 
- `.github/workflows/apply_pr_checks.yaml` — simplified

## Commands                                                                       
  - `docker compose run --rm codegen-check`
  - `docker compose run --rm backend npm run lint`
  - `docker compose run --rm backend npm run build`
  - `docker compose run --rm client npm run lint`
  - `docker compose run --rm client npm run build`
  - `docker compose run --rm test`
 
## Out of scope   
- `check_migration_order` intentionally stays inline in the workflow as it is GitHub-specific

### DIFF
--- a/.github/workflows/apply_pr_checks.yaml
+++ b/.github/workflows/apply_pr_checks.yaml
@@ -29,21 +29,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v6
-        with:
-          node-version-file: '.nvmrc'
-          cache: 'npm'
-      - run: npm ci
-      - name: Generate graphql
-        run: |
-          npm run generate
-          if [[ -z "$(git status --porcelain)" ]]; then
-            echo "Generated GraphQL is up to date."
-          else
-            echo "GraphQL re-generation required!!"
-            echo "::error:: $(git status)"
-            exit 1
-          fi
+      - name: Check generated GraphQL is up to date
+        run: docker compose run --rm --quiet-pull codegen-check
 
   check_migration_order:
     runs-on: ubuntu-latest
@@ -135,14 +122,14 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Lint
-        run: docker compose run --quiet-pull lint
+        run: docker compose run --rm --quiet-pull lint
 
       - name: Build server
-        run: DOCKER_BUILDKIT=1 docker build . --target build_backend
+        run: docker compose build build-backend
 
       - name: Migrate CI DBs and Test
         run: |
-          docker compose run --quiet-pull test
+          docker compose run --rm --quiet-pull test
           echo "ATTENTION: If this step fails, look in the next step for the migration logs."
 
       - # Only output the migration logs if the prior step failed, as they
@@ -171,10 +158,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Lint client
-        run: docker compose run --quiet-pull lint-client
+        run: docker compose run --rm --quiet-pull lint-client
 
       - name: Build client
-        # Without DOCKER_BUILDKIT we'll attempt to build the backend as well
-        run: DOCKER_BUILDKIT=1 docker build client
+        run: docker compose build build-client
 
 
--- a/.github/workflows/apply_pr_checks.yaml
+++ b/.github/workflows/apply_pr_checks.yaml
@@ -125,7 +125,7 @@ jobs:
         run: docker compose run --rm --quiet-pull backend npm run lint
 
       - name: Build server
-        run: docker compose build backend
+        run: docker compose run --rm --quiet-pull backend npm run build
 
       - name: Migrate CI DBs and Test
         run: |
@@ -161,6 +161,6 @@ jobs:
         run: docker compose run --rm --quiet-pull client npm run lint
 
       - name: Build client
-        run: docker compose build client
+        run: docker compose run --rm --quiet-pull client npm run build
 
 
--- a/.github/workflows/apply_pr_checks.yaml
+++ b/.github/workflows/apply_pr_checks.yaml
@@ -122,10 +122,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Lint
-        run: docker compose run --rm --quiet-pull lint
+        run: docker compose run --rm --quiet-pull backend npm run lint
 
       - name: Build server
-        run: docker compose build build-backend
+        run: docker compose build backend
 
       - name: Migrate CI DBs and Test
         run: |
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -118,6 +118,39 @@ services:
...
+  codegen-check:
+    image: node:24.14.1-bullseye
+    working_dir: /src
+    command: bash -c 'set -e
+                      && npm ci
+                      && npm run generate
+                      && git diff --exit-code -- "*generated.ts"'
+    volumes:
+      - .:/src
+      - codegen_node_modules:/src/node_modules
```
juanmrad: Here we are bypassing dockerfile completely. This would actually cause more drift. This should instead use the dockerfile. 

Also it is using bullseye non slim version. which we use on the other places.

.github/workflows/apply_pr_checks.yaml
```diff
@@ -29,21 +29,8 @@ jobs:
...
-      - name: Generate graphql
-        run: |
-          npm run generate
-          if [[ -z "$(git status --porcelain)" ]]; then
-            echo "Generated GraphQL is up to date."
-          else
-            echo "GraphQL re-generation required!!"
-            echo "::error:: $(git status)"
-            exit 1
-          fi
```
juanmrad, line 46: This check here ensures newly generated files or deletions outside the glob are also catched which a simple `git diff --exit-code -- "*generated.ts".` does not catch properly. 

I do believe we only generate for front-end and back-end under `generated.ts` but also wonder if we may move to a better split approach later on which this won't catch.

docker-compose.yaml
```diff
@@ -118,6 +118,39 @@ services:
...
+    command: ['true']
+
+  # Builds the client image. Equivalent to `docker build client` in the
+  # `run_frontend_checks_if_changed` CI job. Run with
+  # `docker compose build build-client`.
+  build-client:
+    build:
+      context: client
+    command: ['true']
+
```
juanmrad: This is not normally how docker compose is used. specially since they are not services. I'd rather rename those to backend / client and use `docker compose run --rm <service> <command>` for each CI step, instead of adding per-step services.


```yaml
backend:
  build: { context: ., dockerfile: Dockerfile, target: server_base }

client:
  build: { context: client, target: client_base }
  volumes:
    - ./client/eslint:/app/eslint
    - ./client/.eslintrc.cjs:/app/.eslintrc.cjs
    - ./client/eslint.config.mjs:/app/eslint.config.mjs
```

docker-compose.yaml
vinaysrao1: Fixed with dockerfile and added it status --porcelain.

.github/workflows/apply_pr_checks.yaml
vinaysrao1: Fixed, in the codegen-check section.

docker-compose.yaml
vinaysrao1: Agreed — collapsed to two reusable services and pushed the per-step commands
 
 We can follow up for migrations and drop_dbs later which still use raw image instead of Dockerfile

.github/workflows/apply_pr_checks.yaml
```diff
@@ -135,14 +122,14 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Lint
-        run: docker compose run --quiet-pull lint
+        run: docker compose run --rm --quiet-pull backend npm run lint
 
       - name: Build server
-        run: DOCKER_BUILDKIT=1 docker build . --target build_backend
+        run: docker compose build backend
```
pawiecz: Previously `Lint` workflow step required only `server_base` build step instead of `build_backend` it uses now. With caching this makes `Build server` workflow step a no-op.

If you'd like to catch build and lint issues in separate workflow steps I'd suggest switching order of these two and build `backend` first, then use the Docker cache to run linter. If you're fine with having it in a single workflow step the `Build server` one can be dropped entirely.

.github/workflows/apply_pr_checks.yaml
```diff
@@ -171,10 +158,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Lint client
-        run: docker compose run --quiet-pull lint-client
+        run: docker compose run --rm --quiet-pull client npm run lint
 
       - name: Build client
-        # Without DOCKER_BUILDKIT we'll attempt to build the backend as well
-        run: DOCKER_BUILDKIT=1 docker build client
+        run: docker compose build client
```
pawiecz: Comment on lint/build order also applies here (`client_base` build step vs. default Docker build target for the `client` context)

docker-compose.yaml
```diff
@@ -100,19 +100,33 @@ services:
...
+                      && [[ -z "$(git status --porcelain)" ]]'
+    volumes:
+      - .:/src
+      - codegen_node_modules:/src/node_modules
+
+  backend:
+    build:
+      context: .
+      dockerfile: Dockerfile
+      target: build_backend
```
juanmrad: This should be `target: server_base`, not `build_backend`. `build_backend` adds 
```
RUN npm run build
```

on top of `server_base`, so `docker compose run --rm backend npm run lint` transitively runs a full TypeScript compile before lint ever executes. 

Basically we are now paying the build step when we are trying to run "lint" 

And any tsc error surfaces as a lint failure. 

We should instead be pointing at `server_base` and invoking the build as a runtime command: 
```
docker compose run --rm backend npm run build
```

So here:
```suggestion
      target: server_base
```

.github/workflows/apply_pr_checks.yaml
juanmrad:
```suggestion
        run: docker compose run --rm backend npm run build
```

based on my comment above

juanmrad (changes requested):

.github/workflows/apply_pr_checks.yaml
vinaysrao1: switched the backend compose service to target: server_base in the prev commit and made build server a runtime command (docker compose run --rm  --quiet-pull backend npm run build).

.github/workflows/apply_pr_checks.yaml
vinaysrao1: same fix applied client side.

docker-compose.yaml
vinaysrao1: Done

.github/workflows/apply_pr_checks.yaml
vinaysrao1: Done

juanmrad (approved):
LGTM! :shipit:

PR #328, opened 2026-04-27
Fix `create-org` and update readme

dom-notion:
I ran into a few bumps running `coop` locally, so this PR introduces the changes I need to make to unblock myself.
- `npm run create-org` wasn't working because the org wasn't created before trying to make the key, so I reordered that.
- `websiteUrl` for `create-org` needs to pass kysely's URL validation, so I added a step in the CLI to prepend `https://` if a scheme isn't provided by the caller
- I had to run `npm run db:create` for all the dbs before migrating, so I added that as a step to the README.

## Tests

ran `npm run create-org` locally successfully

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,12 @@ and then follow the steps below:
    (cd server && npm install)
    (cd db && npm install)
    ```
-3. Make sure the `.env` files for `/server` and `db` are populated (including ClickHouse credentials). Run database migrations:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -77,6 +81,8 @@ npm install \
 5. Create an organization and admin user:
 ```bash
    npm run create-org
+
+6. Run `cd server && npm run copy-assets`
```
juanmrad, line 85: I do wonder if we can make this copy assets part of some hook 🤔

juanmrad (approved):
👍🏻 Thank you!.

PR #346, opened 2026-04-28
Fix coop svg export

### LINKED ISSUE #345: Reduce Vite build work caused by broad SVG icon barrel exports
The client icon barrel at client/src/icons/index.tsx re-exported the full lnif icon tree with:

```ts
export * from './lnif';
```

That lnif barrel fans out across thousands of local SVG files, all loaded through vite-plugin-svgr. As a result, importing a small number of icons from @/icons caused Vite to transform many unnecessary modules during production builds.

### Impact
This increases build work and slows Vite production builds.

Observed comparison:

- Before: 11121 modules transformed, 16.50s
- After narrowing exports: 8403 modules transformed, 13.99s

### Proposed fix
Replace broad barrel exports with exact-file exports for only the icons actually used by the app.

serendipty01:
fixes #345 
This PR updates client/src/icons/index.tsx so @/icons only exposes the exact icons currently used by the app, instead of re-exporting the full lnif SVG tree.

It also simplifies those exports by using direct exact-file re-exports instead of wrapper components.

### Why
The previous barrel export caused Vite to walk and transform a very large local SVG icon set through vite-plugin-svgr, even when only a few icons were imported.

That unnecessary fan-out increased production build work.

### Results
Build comparison:

- Before: 11121 modules transformed, 16.50s
- After: 8403 modules transformed, 12.63s


juanmrad (approved):
👍🏻

PR #349, opened 2026-04-29
[kysely] Migrate User schema and auth out of sequelize

juanmrad:

Migrates `User` persistence and session/auth plumbing off Sequelize and onto Kysely, continuing the Sequelize -> Kysely migration.

Added new persistent layer on kysely to handle User operations. Adds app layer validations to match sequelize existing validation. 

Refactored UserAPI to use KyselyPgs and scopes some of the queries to tighten ownership.


server/graphql/datasources/UserApi.ts
```diff
@@ -138,57 +177,57 @@ class UserAPI {
...
       });
     }
 
-    // Hash and save new password
     const hashedNewPassword = await hashPassword(newPassword);
-    user.password = hashedNewPassword;
-    await user.save();
+    await kyselyUserUpdate(this.kyselyPg, user.id, {
+      password: hashedNewPassword,
+    });
```
vinaysrao1, line 237: Do we need a return value for potentially failed update?

server/graphql/datasources/UserApi.ts
```diff
@@ -212,60 +260,98 @@ class UserAPI {
...
     if (user.orgId !== invokerOrgId) {
       throw makeUnauthorizedError(
         'You can only approve users in your organization',
         { shouldErrorSpan: true },
       );
     }
 
-    user.approvedByAdmin = true;
-    await user.save();
+    await kyselyUserUpdate(this.kyselyPg, id, { approvedByAdmin: true });
```
vinaysrao1: Like above.

vinaysrao1 (approved):

PR #350, opened 2026-04-29
[Kysely] Migration of condition result types off sequelize

juanmrad:

Moves the nine condition-result types out of `server/models/rules/RuleModel.ts` (via `ruleTypes.ts`) into a new Sequelize-free home at `server/services/moderationConfigService/types/conditionResults.ts`, and flips the three associated GraphQL codegen mappers. This is pure prep work for the final Sequelize removal no real logic changes yet.

The reasoning behind this is that the codegen mapper for Rule, Backtest and ItemType would be blocked on the RuleModel.ts still being the top level home for these types even tho they do not depend on sequelize. Moving this would allow us to stop the transitive imports from the models. And keeping prs on the smaller side for review.

Follow up pr will do the actual flip on the codegen and the RuleModel.ts itself.

### DIFF
--- a/codegen.yaml
+++ b/codegen.yaml
@@ -114,11 +114,11 @@ generates:
         # `buildGraphqlRuleParent` casts to it at the boundary.
         Rule: ../models/rules/RuleModel.js#Rule
         Condition: ../services/moderationConfigService/index.js#Condition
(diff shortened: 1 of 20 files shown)


vinaysrao1 (approved):
Great PR. LGTM

PR #352, opened 2026-04-29
fix: disable demo request emails with env var

calebmcquaid:

Users are getting demo request emails from bots when setting an email address in their config. This enables them to turn that off.

## Tests

The requestDemo resolver now checks for ENABLE_DEMO_REQUEST=true before sending emails. Without the env var set (or set to  anything other than true), the endpoint returns false and no email is sent. No existing tests break since the env var defaults to disabled. Added the new var to .env.example and .env.githubci.


juanmrad (approved):
we should look into removing this feature altogether.

PR #354, opened 2026-04-29
Make Kysely work with NoOp data warehouse adapter

dom-notion:
When trying to deploy coop with a `NoOp` warehouse adapter, I ran into errors about there being no Kysely dialect for that adapter. This changes the behavior to instead only throw errors when queries are actually attempted, instead of on startup.

## Tests
Was able to successfully deploy this code to AWS. Also was able to run locally still.

### DIFF
--- a/server/storage/dataWarehouse/DataWarehouseFactory.ts
+++ b/server/storage/dataWarehouse/DataWarehouseFactory.ts
@@ -27,6 +27,7 @@ import {
   type DataWarehouseProvider as IDataWarehouseProvider,
   type TransactionFunction,
 } from './IDataWarehouse.js';
+import { type Kysely } from 'kysely';
 import type {
   AnalyticsSchema,
   BulkWriteConfig,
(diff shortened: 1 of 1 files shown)


server/storage/dataWarehouse/DataWarehouseFactory.ts
```diff
@@ -71,6 +72,22 @@ export type DataWarehouseConfig =
...
+    // allowing services to hold a reference without crashing at startup.
+    return new Proxy({} as Kysely<any>, {
+      get(_target, prop) {
+        if (prop === 'destroy') return async () => {};
+        return () => {
+          throw new Error('NoOp dialect: Kysely queries are not supported');
+        };
+      },
+    });
+  }
```
juanmrad, line 89:
```suggestion
    return new Proxy({} as Kysely<any>, {
      get(_target, prop) {
        if (prop === 'destroy') return async () => {};
        if (prop === 'then') return undefined; // not thenable
        If (typeof prop === 'symbol') return undefined;
        return () => {
          throw new Error('NoOp dialect: Kysely queries are not supported');
        };
      },
    });
  }
```

server/storage/dataWarehouse/DataWarehouseFactory.ts
```diff
@@ -71,6 +72,22 @@ export type DataWarehouseConfig =
       analyticsProvider?: AnalyticsProvider;
     };
 
+class NoOpKyselyDialect implements IDataWarehouseDialect {
+  getKyselyInstance(): Kysely<any> {
+    // Return a proxy that throws only when a query is actually executed,
+    // allowing services to hold a reference without crashing at startup.
```
juanmrad:
```suggestion
    // Return a proxy so services can hold a Kysely reference without
    // crashing at startup; any attempt to build or execute a query throws.
```

juanmrad (approved):
two minor comments here the comment is more about what would really happen and also safe return on then and other logger primitives.

PR #371, opened 2026-04-30
Update minimum memory requirement for instance deployment

### LINKED ISSUE #51: Clarifiy resource requirements in documentation
Right now it's not clear what the difference is between development versus production requirements. We should clarify this in the docs!

pawiecz:

FIxes: #51


cassidyjames (approved):
Thank you @pawiecz!

PR #334, opened 2026-04-28
fix: lazy-load dashboard routes

serendipty01:
## Context & Requests for Reviewers
fixes #333

PR #372, opened 2026-04-30
Fix job fragment circular dependency

### LINKED ISSUE #369: Manual review queue crash: `can't access lexical declaration 'JOB_FRAGMENT' before initialization`
I'm running Coop locally from main. When I open a manual review queue page (e.g. `/dashboard/manual_review/queues/review/<id>`), I get an an unhandled application error in the browser:

> Unexpected Application Error!
> can't access lexical declaration 'JOB_FRAGMENT' before initialization
>
> `@http://localhost:3000/src/webpages/dashboard/items/ItemActionHistory.tsx:82:11`

I did some digging with Claude, and it said it could be a circular module dependency. Here's what it said:

> This is a JavaScript Temporal Dead Zone (TDZ) violation caused by a circular module dependency. The cycle is:
>
> ```
> ManualReviewJobReview.tsx
>   → imports ManualReviewJobContentView.tsx
>     → imports ItemActionHistory.tsx
>       → imports JOB_FRAGMENT from ManualReviewJobReview.tsx  ← back to start
> ```
>
> `ManualReviewJobPrimaryUserComponent.tsx` and `ManualReviewJobRelatedUserComponent.tsx` (both also imported by `ManualReviewJobReview.tsx`) add two more paths through the same cycle.
>
> When the JS runtime starts evaluating `ManualReviewJobReview.tsx`, it pauses to evaluate its imports, which eventually leads back to `ItemActionHistory.tsx`. At that point, `ItemActionHistory.tsx` tries to use `JOB_FRAGMENT` in its module-level `gql` call (line 85) — but `ManualReviewJobReview.tsx` hasn't finished evaluating yet, so `JOB_FRAGMENT`'s `const` binding exists but has never been assigned. That's the TDZ error.
>
> The cycle appears to have existed since the initial release. It may have gone unnoticed because Webpack handles circular dependencies more permissively — it uses CommonJS-style evaluation where cycles result in `undefined` rather than a hard error. Vite uses native ES modules, which enforce strict TDZ semantics, so the Vite migration likely surfaced a latent bug rather than introducing a new one.

dom-notion:
Fixes #369 

On manual review pages, I started seeing `ReferenceError: Cannot access 'JOB_FRAGMENT' before initialization
    at http://localhost:3000/src/webpages/dashboard/items/ItemActionHistory.tsx?t=1777585623049:82:11`. It looks like this is a regression caused by https://github.com/roostorg/coop/pull/334. I fixed the error by splitting `JOB_FRAGMENT` out into its own file.

## Tests
manual review pages now load


cassidyjames (commented):
I can confirm that this fixes the issue for me locally, but I'm not comfortable enough with TypeScript to fully approve this. 😅 I'll let another maintainer take a look.

juanmrad (approved):
LGTM.

PR #376, opened 2026-05-01
[Dev] Adjustments to Postgres settings for development

juanmrad:

Based on debug when setting up Coop on a VPS we uncover some misses:

1) We did not pass down the SSL settings on the db migration.
2) We used default amount of pool connections which for some deployed services there may be connection limits as such we need to allow adjustments. 
3) pgdump schema was setting owner as `postgres` which not all databases will use this user. as such we should use `CURRENT_USER` which is supported in Postgres. Adjusted agents.md to also note when creating new migrations that this should be used. 

## Tests

manually tested this on VPS environment


juanmrad:
Ignoring migration comment. this is only to fix for future new users and has no effect on current users :)

ThatKoffe (approved):

PR #390, opened 2026-05-01
[Kysely] Migrate OrgAPI parent helpers off sequelize

juanmrad:

Continues the `Sequelize` -> `Kysely` migration. Removes the last two runtime Sequelize callers in `OrgAPI` so the datasource no longer depends on `Dependencies['Sequelize']`.

- `getSequelizeContentTypesForOrg(orgId)` is renamed to `getContentTypesForOrg(orgId)` and is now backed by `ModerationConfigService.getItemTypes({ orgId })`. Sequelize parity is preserved: returns every item-type kind for the org (USER + CONTENT + THREAD), matching the previous `findAll({ where: { orgId } })` shape.
- `getOrgUsersForGraphQL(orgId)` is now backed by `kyselyUserListByOrg`, returning `GraphQLUserParent[]`. The existing `Org.usersWhoCanReviewEveryQueue` filter (`u.getPermissions().includes('EDIT_MRT_QUEUES')`) keeps working unchanged.
- `Sequelize` is dropped from `OrgAPI`'s constructor and from the IoC `inject([...])` list.
The `ContentType` GraphQL codegen mapper is repointed from `models/rules/ItemTypeModel#ItemType` to `services/moderationConfigService/types/itemTypes#ItemType`. The two shapes differ on one field (the MCS type uses `schema` where the Sequelize model used `fields`), so `graphql/modules/contentType.ts` `baseFields` / `derivedFields` now read `contentType.schema`. No client query selects `Org.contentTypes` today, but the GraphQL surface is unchanged.


juanmrad:
@copilot review

cassidyjames (approved):
Smoke tested the Coop front-end with my atproto firehose locally similar to #420. Everything appears to work well here as well. :)

PR #391, opened 2026-05-01
[Husky] Wire up husky pre-commit with lint-staged

juanmrad:

We had Husky installed and recently bumped it to v9, but had no hooks wired up. After debating on leaving or removing. I've had suffered from not running lint and seen other prs also opened with no lint pass which made it obvious this would be a great addition to our flow.

I am adding the pre-commit hook for `lint-staged` which will make prettier and eslint fix any automatic lint errors before commit.

Considered also having a pre-push for running typecheck, but that may take around 30s or so. Would evaluate it after some time as it may be valuable as well.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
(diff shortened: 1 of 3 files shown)


cassidyjames (approved):
:shipit:

PR #400, opened 2026-05-05
Add parameterized actions to provide runtime values when executing

### LINKED ISSUE #377: Parameterized Actions
# Problem
Custom actions currently only support adding custom payload when the Action is created not when it is run.  

For example, you can only define the Action “Ban a user for 1 week”, but you cannot defined an action “Ban a user for N weeks” which would allow a moderator to ban the user for any number weeks, "N".

# Suggestion
Allow custom Actions to be defined with user-defined parameters similar to how item types can have user-defined data.  

Parameters are defined on a per-action basis.  Each parameter would have a key and some kind of data type associated with it (e.g. “numDaysBanned” - INT). A parameter can be required or optional.  When a moderator selects an action that has a parameter associated with it, they would be required to enter all required parameters.

Parameters provided by the moderator would then be passed to the action webhook associated with the Action.

Some suggested controls that would be useful:
- Select/dropdown of defined values
- Multiselect of defined values
- Free text input
- Numerical input (same as above, only numbers allowed)

juanmrad:
Adds parameterized actions and all related UI and audit logging to fix #377.

## Add typed Action parameter authoring

Introduce a typed parameter spec for Actions so moderators can be prompted for runtime values when executing an action (e.g. "Ban for N days"). This first half of addressing #377 is the authoring and configuration part definition, validation, persistence, and the create/edit UI from the actions form only.

Added new Actions model and added new fields to the existing actions model.

Screenshot 2026-05-04 at 7 49 59 PM
Screenshot 2026-05-04 at 7 49 42 PM
Screenshot 2026-05-04 at 7 49 19 PM

## Wire Action parameters through execution + persist to ClickHouse audit log

This is the second half of addressing #377 which has been merged into this branch and included in this PR.

Routes moderator-supplied parameter values from each run UI through the publisher into the action webhook payload, adding a separate "moderator note" field on every execution path, and recording both to the `analytics.ACTION_EXECUTIONS` audit log.

Added validation for the parameters on the action page and the actions form. Added form on the actioning with modal to add details on action to be taken and ability to edit those details.

Screenshot 2026-05-04 at 8 55 49 PM
Screenshot 2026-05-04 at 8 55 07 PM
Screenshot 2026-05-04 at 8 55 04 PM
Screenshot 2026-05-05 at 10 31 41 PM
Screenshot 2026-05-05 at 10 31 36 PM
Screenshot 2026-05-05 at 10 31 24 PM


cassidyjames (approved):
I'm no typescript expert but this looks like it does the thing. Thanks @juanmrad!

PR #407, opened 2026-05-05
Show comments from deleted users on manual reviews

dom-notion:
Right now, if a job has a comment from a deleted user, the manual review page fails to load. GraphQL expects all comments to have an associated user, which doesn't work if the user has been deleted.

This PR makes the comment author optional, and shows "Deleted User" in those cases in the list of comments.

## Tests
Manually verified that I could load a manual review page for a job with a comment from a deleted user.


juanmrad (approved):
LGTM

PR #417, opened 2026-05-08
mdbook: use versioned subdirectories for GitHub Pages

### LINKED ISSUE #415: Versioned docs
Right now the [docs site](https://roostorg.github.io/coop/) tracks the main branch; this means that any documentation reworking in between releases will break existing docs links. It also means that if someone is using Coop v1 and we theoretically released Coop v2 with a bunch of breaking changes and new UI, the docs would no longer be accurate for those Coop v1 users.

The solution: versioned docs, e.g. `/docs/$TAG/path`. We could either:

1. Keep the docs in the repo relevant to `main`, and handle versioning in GitHub Actions workflow when building and deploying to Pages, or

2. Maintain the versioned docs directly in the repo, e.g. under versioned sub-directories

I actually like the first option more, as it gives us the most flexibility and ensures the only docs in a checkout at any given time are relevant to the code in that checkout.

cassidyjames:
Fixes #415

Switches from the artifact-based deploy-pages approach (which replaces the entire site on each run) to deploy to a `gh-pages` branch scoped to a single target folder.

- Docs in `main` deploy to `/latest/`
- Tagged releases deploy that tag's docs to a matching subdirectory (e.g. `/0.1/`)
- A simple index is regenerated after each deployment to list available versions at `/`

This does bring a one-time break of existing docs URLs on the generated docs site since they'll be under `/latest/` now; however, going forward this should give us more stable URLs since they'll be scoped to the released versions. Links to Markdown files directly in the GitHub web UI will remain the same as before, including browsing the repo by tag or commit to view a snapshot of the docs.

After updating the Pages source setting to serve from the `gh-pages` branch, we'll need to backfill old releases:

```sh
gh workflow run mdbook.yml --repo roostorg/coop --field ref=0.0
gh workflow run mdbook.yml --repo roostorg/coop --field ref=0.1
```

## Non-code follow-up

- [x] switch GitHub Pages setting to deploy from `gh-pages` branch
- [x] backfill old releases
- [x] test and make sure it all looks right!

### DIFF
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -42,6 +42,8 @@ jobs:
     env:
       MDBOOK_VERSION: 0.5.2
     steps:
+      # inputs.ref is only set for manual backfills; push-to-main and tag-push events
+      # default to the triggering ref automatically
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref }}
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -33,7 +33,7 @@ permissions:
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
-  group: 'pages'
+  group: 'mdbook'
   cancel-in-progress: false
 
 jobs:
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -62,26 +62,37 @@ jobs:
         run: ${{ github.workspace }}/mdbook-bin/mdbook build docs
 
       - name: Deploy to GitHub Pages
-        uses: JamesIves/github-pages-deploy-action@v4
-        with:
-          folder: docs/book
-          target-folder: ${{ steps.deploy-path.outputs.path }}
-
-      - name: Checkout gh-pages for index generation
-        uses: actions/checkout@v4
-        with:
-          ref: gh-pages
-          path: gh-pages-content
-
-      - name: Generate version index
-        working-directory: gh-pages-content
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
         run: |
-          mkdir -p ../index-deploy
-          tree -d -L 1 -H . -T "Coop Documentation" -o ../index-deploy/index.html --noreport
+          DEPLOY_PATH="${{ steps.deploy-path.outputs.path }}"
 
-      - name: Deploy root index
-        uses: JamesIves/github-pages-deploy-action@v4
-        with:
-          folder: index-deploy
-          target-folder: .
-          clean: false
+          # Clone the gh-pages branch, or initialize it if this is the first deployment
+          if git ls-remote --exit-code --heads origin gh-pages > /dev/null 2>&1; then
+            git clone --depth=1 --single-branch --branch gh-pages \
+              "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" \
+              _gh-pages
+          else
+            mkdir _gh-pages
+            git -C _gh-pages init
+            git -C _gh-pages checkout --orphan gh-pages
+            git -C _gh-pages remote add origin \
+              "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+          fi
+
+          touch _gh-pages/.nojekyll
+
+          # Deploy the versioned docs
+          rm -rf "_gh-pages/${DEPLOY_PATH}"
+          cp -r docs/book "_gh-pages/${DEPLOY_PATH}"
+
+          # Regenerate the root version index
+          (cd _gh-pages && tree -d -L 1 -H . -T "Coop Documentation" -o index.html --noreport)
+
+          # Commit and push
+          git -C _gh-pages config user.name "github-actions[bot]"
+          git -C _gh-pages config user.email "github-actions[bot]@users.noreply.github.com"
+          git -C _gh-pages add --all
+          git -C _gh-pages diff --staged --quiet || \
+            git -C _gh-pages commit --message "Deploy docs: ${DEPLOY_PATH}"
+          git -C _gh-pages push origin gh-pages
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -3,34 +3,42 @@ name: Deploy mdBook site to Pages
 on:
   # Runs on pushes targeting the default branch, only when docs or this workflow change
   push:
-    branches: ["main"]
+    branches: ['main']
     paths:
-      - "docs/**"
-      - ".github/workflows/mdbook.yml"
+      - 'docs/**'
+      - '.github/workflows/mdbook.yml'
+    tags:
+      - '[0-9]*'
 
-  # Allows you to run this workflow manually from the Actions tab
+  # Allows you to run this workflow manually from the Actions tab (supports --ref for backfilling tags)
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


cassidyjames:
I'm testing this on a personal fork at [cassidyjames/coop](https://github.com/cassidyjames/coop/actions). Currently getting an error and looking into it:


>**deploy**
> Tag "0.0" is not allowed to deploy to github-pages due to environment protection rules.
> 
> **deploy**
> The deployment was rejected or didn't satisfy other protection rules.

cassidyjames:
Ah, it was an environment protection rule; I've fixed that in my fork and can do the same in this repo so tags are allowed to deploy to GitHub Pages.

cassidyjames:
Okay this works! Tested on [my fork](https://github.com/cassidyjames/coop).

- Output [docs site](https://cassidyjames.com/coop/) and [source branch](https://github.com/cassidyjames/coop/tree/gh-pages)
- Manual/backfill [workflow run](https://github.com/cassidyjames/coop/actions/runs/25576304691)
- New tag [workflow run](https://github.com/cassidyjames/coop/actions/runs/25580972150/job/75099568213)

.github/workflows/mdbook.yml
```diff
@@ -1,36 +1,64 @@
...
   push:
-    branches: ["main"]
+    branches: ['main']
     paths:
-      - "docs/**"
-      - ".github/workflows/mdbook.yml"
+      - 'docs/**'
+      - '.github/workflows/mdbook.yml'
+    tags:
+      - '[0-9]*'
```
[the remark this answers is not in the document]
cassidyjames: I'm actually not sure this concern is valid, as it hasn't shown up when testing on my fork at [cassidyjames/coop](https://github.com/cassidyjames/coop).

For example, see [this release](https://github.com/cassidyjames/coop/releases/tag/0.3); it's the same contents as the previous release (so no changes to any files, and no pushes to main), and yet it automatically triggered [this action run](https://github.com/cassidyjames/coop/actions/runs/25589705288) which built the 0.3 docs, which were then automatically [published by GitHub Pages](https://cassidyjames.com/coop/0.3/). Am I missing the actual issue?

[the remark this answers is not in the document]
cassidyjames: According to [GitHub's own docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore):

> Path filters are not evaluated for pushes of tags.

Which matches my experience in my fork. Soo, gonna consider this resolved.

.github/workflows/mdbook.yml
```diff
@@ -1,36 +1,64 @@
...
-  # Build job
   build:
     runs-on: ubuntu-latest
     env:
       MDBOOK_VERSION: 0.5.2
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+
```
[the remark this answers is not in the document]
cassidyjames: I think this is also not actually an issue.

Looking at the [`actions/checkout` source code](https://github.com/actions/checkout/blob/v4/src/input-helper.ts#L60-L61), the action reads the `ref` input via `core.getInput('ref')`, which returns an empty string when the input is unset. Then it checks `if (!result.ref)`, which is falsey for an empty string—the same as for an omitted input—so it falls back to `github.context.ref` and `github.context.sha` either way.

It should be the same whether `ref` is explicitly passed as `""` or just left out.

cassidyjames:
@juanmrad re: your question from Discord, the `ref` parameter in `actions/checkout` only controls what gets checked out into the workspace at that step, not which workflow file is run in the first place; the workflow file is set by the trigger event (or the branch selected in the "Run workflow" UI, which defaults to `main` for `workflow_dispatch`). By the time the checkout step runs, the workflow file is already loaded.

So for a manual backfill, the current workflow from `main` (or the manually-selected branch) runs, then the checkout step pulls the docs from the specified tag to build into the `gh-pages` branch.

.github/workflows/mdbook.yml
```diff
@@ -1,36 +1,66 @@
...
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
-  group: "pages"
+  group: 'pages'
   cancel-in-progress: false
 
 jobs:
-  # Build job
   build:
     runs-on: ubuntu-latest
```
juanmrad, line 41:
```suggestion
    runs-on: ubuntu-latest
    if: ${{ github.repository == 'roostorg/coop' }}
```

Avoid running this on Other people's forks of Coop

.github/workflows/mdbook.yml
```diff
@@ -1,36 +1,66 @@
...
     env:
       MDBOOK_VERSION: 0.5.2
     steps:
+      # inputs.ref is only set for manual backfills; push-to-main and tag-push events
+      # default to the triggering ref automatically
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
```
juanmrad, line 50: Nit: if someone uses the "Run workflow" UI from a non-main branch and leaves ref empty, this falls back to `github.context.ref` (the dispatcher branch), and then on line 59 `GITHUB_REF_TYPE` is branch and `DEPLOY_REF` doesn't start with a digit, so the docs from that feature branch get published to `/latest/`. 

We should guard the dispatch path to require either `inputs.ref` to be set, or `github.ref_name == 'main'.`

.github/workflows/mdbook.yml
```diff
@@ -1,36 +1,66 @@
...
-  contents: read
-  pages: write
-  id-token: write
+  contents: write
 
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
-  group: "pages"
+  group: 'pages'
```
juanmrad: Nit: the concurrency group name still says pages but we no longer deploy via the Pages environment. Something like `docs-gh-pages` or `docs-gh`

.github/workflows/mdbook.yml
```diff
@@ -1,36 +1,66 @@
...
+      - name: Determine deployment path
+        id: deploy-path
+        run: |
+          # For manual backfill, use the input ref; otherwise derive from the triggering ref
+          DEPLOY_REF="${{ inputs.ref }}"
+          if [[ -z "$DEPLOY_REF" ]]; then
+            DEPLOY_REF="$GITHUB_REF_NAME"
+          fi
+          if [[ "$GITHUB_REF_TYPE" == "tag" || "$DEPLOY_REF" =~ ^[0-9] ]]; then
+            echo "path=$DEPLOY_REF" >> "$GITHUB_OUTPUT"
```
juanmrad, line 69:
```suggestion
          if [[ "$GITHUB_REF_TYPE" == "tag" || "$DEPLOY_REF" =~ ^[0-9] ]]; then
            if [[ "$DEPLOY_REF" =~ [^A-Za-z0-9._-] ]]; then
               echo "Refusing unsafe deploy ref: $DEPLOY_REF" >&2
               exit 1
            fi
            echo "path=$DEPLOY_REF" >> "$GITHUB_OUTPUT"
```

We should add some sanitation on the refs just in case. This closes the path-traversal vector on line 108 for tags with `/` or `..`, and also validates workflow_dispatch input.

juanmrad (changes requested):

.github/workflows/mdbook.yml
cassidyjames: Makes sense; what about just matching the workflow name:

```suggestion
  group: 'mdbook'
```

.github/workflows/mdbook.yml
cassidyjames: Good catch, that would be annoying. 😅 Addressed in the latest commit.

juanmrad (approved):
👍🏻

PR #433, opened 2026-05-11
Build and serve client assets from the server container

dom-notion:
This adds client assets to the server container which makes it easier to deploy coop on ECS or something similar without needing a CDN.

## Tests
Deployed coop to ecs with these changes. Also verified that the app works locally.

### DIFF
--- a/server/server.ts
+++ b/server/server.ts
@@ -1,3 +1,5 @@
+import path from 'path';
+import { fileURLToPath } from 'url';
 import express from 'express';
 
 import makeApiApp from './api.js';
@@ -49,6 +51,18 @@ export default async function makeWWWServer(deps: Dependencies) {
   });
 
   app.use('/api/v1', apiApp);
+
+  const clientBuildPath = path.resolve(
+    path.dirname(fileURLToPath(import.meta.url)),
+    'client-build',
+  );
+  app.use(express.static(clientBuildPath));
+  app.get('{*path}', (_req, res, next) => {
+    const indexPath = path.join(clientBuildPath, 'index.html');
+    res.sendFile(indexPath, (err) => {
+      if (err) next();
+    });
+  });
 
   return { app, shutdown };
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# This is a fairly-standard multi-stage Dockerfile. We build 
+# This is a fairly-standard multi-stage Dockerfile. We build
 # backend in a Node image, and then we copy the built files
 # (but not the devDependencies [like typescript, etc.] or the raw source
 # files) to final images that we'll actually run. This makes the final image a
@@ -16,6 +16,13 @@ RUN npm ci
 COPY ["server", "./"]
 
 FROM server_base AS build_backend
+RUN npm run build
+
+FROM node:24.14.1-bullseye-slim AS build_client
+WORKDIR /app
+COPY ["client/package.json", "client/package-lock.json", "./"]
+RUN npm ci
+COPY ["client", "./"]
 RUN npm run build
 
 # make a shared layer that can be the base for worker and api images.
@@ -39,6 +46,7 @@ ENV BUILD_ID=$BUILD_ID
 # Expose 8080 because the backend will run on this port absent a
 # process.env.PORT to the contrary.
 FROM backend_base AS build_server
+COPY --from=build_client /app/build ./client-build
 EXPOSE 8080
 CMD ["node", "bin/www.js"]
 


server/server.ts
juanmrad, line 51: We should move the `helmet` set-up here.

juanmrad (changes requested):
There are several issues with this pr. 

`helmet`, `cors`, `express.json`, `sessions`, `passport`, and the 404/error handlers all live inside `makeApiServer` (server/api.ts).

Roughly, I'd want s`erver/server.ts` to end up something like:

```ts
const ONE_YEAR_SECONDS = 60 * 60 * 24 * 365;

export default async function makeWWWServer(deps: Dependencies) {
  const { app: apiApp, shutdown } = await makeApiApp(deps);
  const app = express();
  app.set('trust proxy', true);
  app.use((req, _res, next) => {
    Object.defineProperty(req, 'secure', { get: () => true });
    Object.defineProperty(req, 'protocol', { get: () => 'https' });
    req.headers['x-forwarded-proto'] = 'https';
    next();
  });
  // Security headers must cover both the API and any static HTML/JS we serve.
  app.use(helmet({ /* production CSP here */ }));
  app.use('/api/v1', apiApp);
  const clientBuildPath =
    process.env.CLIENT_BUILD_PATH ??
    path.resolve(path.dirname(fileURLToPath(import.meta.url)), 'client-build');
  if (existsSync(path.join(clientBuildPath, 'index.html'))) {
    app.use(
      express.static(clientBuildPath, {
        setHeaders(res, filePath) {
          res.setHeader(
            'Cache-Control',
            filePath.endsWith('index.html')
              ? 'no-cache'
              : `public, max-age=${ONE_YEAR_SECONDS}, immutable`,
          );
        },
      }),
    );
    app.get('/{*path}', (req, res, next) => {
      // Only return the SPA shell for navigation requests; let asset misses 404.
      if (path.extname(req.path) !== '' && req.path !== '/') return next();
      if (!req.accepts('html')) return next();
      res.setHeader('Cache-Control', 'no-cache');
      res.sendFile(path.join(clientBuildPath, 'index.html'), (err) => {
        if (err) next();
      });
    });
  }
  return { app, shutdown };
}
```

ThisIsMissEm:
@juanmrad and if we add in vite for development, then your above example becomes:

```ts
  // server/createServer.ts
  import express from 'express'
  import path from 'node:path'
  import { fileURLToPath } from 'node:url'
  import { createApi } from './createApi.js'

  export async function createServer(deps: Dependencies) {
    const { app: apiApp, shutdown } = await createApi(deps)
    const app = express()

    app.set('trust proxy', true)
    // [proxy-header overrides, helmet, etc — juanmrad's sketch unchanged]
    app.use('/api/v1', apiApp)

    if (process.env.NODE_ENV === 'development') {
      // Dynamic import: vite stays a devDependency, never loaded in prod
      const { createServer: createViteServer } = await import('vite')
      const vite = await createViteServer({
        server: { middlewareMode: true },
        appType: 'spa', // auto-handles SPA fallback
        root: path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../client'),
      })
      app.use(vite.middlewares)
    } else {
      // Production: serve built assets + SPA fallback (juanmrad's sketch unchanged)
    }
  
    return { app, shutdown }
  }
```

(I renamed makeApiApp to createApi, and makeWWWServer to createServer)

That gives you a single server to start in development, and fixes that issue with codespaces mentioned in the AGENTS.md file.

ThisIsMissEm:
Although, also instead of using `res.sendfile` we could do:

```ts
else {
  const shell = await fs.readFile(path.join(clientBuildPath, 'index.html'), { encoding: 'utf-8' })
  // ...
  app.get('/{*path}', (req, res, next) => {
    // ...
    res.write(shell)
  })
```

which avoids hitting disk for every request to the shell in production — it's not going to change so sendfile doesn't make sense.

ThisIsMissEm:
@dom-notion thanks for kicking this off. Happy to pick this up if you're not actively iterating — let me know either way?

juanmrad:
@ThisIsMissEm Makes sense we could also add some caching layer. I mean ideally some of all these should be served by a CDN.

ThisIsMissEm:
@juanmrad yeah, that's up to end-users of coop to configure, the docker image would just serve the assets so you can do that.

cassidyjames:
We briefly discussed this in today's working group; @ThisIsMissEm is going to look at filing an issue to track this and picking this PR up in a new branch.

PR #436, opened 2026-05-12
[Kysely] Switch lint enforcement from Sequelize to Kysely and added transaction wrapper enforcement

juanmrad:

Removes dead lint rule for `unmanagedSequelizeTransactionSelector` and adds a new one for Kysely `makeKyselyTransactionWithRetry`. As part of this I also moved all direct callers of kyely execute to use the wrapped Transaction query. 

This has no real behavior change, but moves towards better lint rules and code standards when working with kysely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deletions now report success only when exactly one row is removed.
  * Tests no longer generate invalid dates.

* **Improvements**
  * Write operations now use retry-aware transactions with optional isolation-level support to reduce transient failures.
  * Several delete APIs now require explicit organization scoping.

* **Tests**
  * Added tests validating retryable transaction behavior and retry scenarios.

* **Documentation**
  * Updated inline docs and examples for clarity.


julietshen (approved):
trusting the ai review for changes, logic looks good to me

PR #439, opened 2026-05-12
ci: harden GitHub Actions: SHA-pin all uses, --ignore-scripts

### LINKED ISSUE #447: Harden GitHub Actions workflows
In light of the recent [tanstack supply chain attack](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem), we should pin third-party actions to a specific release SHA to harden against similar compromises to the supply chain.

Full disclosure, @haileyok has an open PR at #439 (thank yoooou!) but I wanted to make sure we had an open issue on the v1 milestone to make it easier to track.

haileyok:
Fixes #447 

## Summary

Pin every action in `.github/workflows/` to a specific patch-release SHA (with version comments) to harden against the upstream-tag-hijack supply-chain pattern.
## Changes

| Action | Pinned to |
|---|---|
| `actions/checkout` | v4.3.1 |
| `actions/cache` | v4.3.0 |
| `actions/setup-node` | v6.4.0 |
| `dorny/test-reporter` | v1.9.1 |
| `ethanresnick/pr-checklists` | `v1` SHA — maintainer does not publish patch tags |

**`.github/dependabot.yml`** — added `github-actions` ecosystem (weekly, grouped) so the new SHA pins get bumped automatically and don't bitrot.

**`publish-db-migrator.yaml` / `publish-types.yaml`** — `npm ci` → `npm ci --ignore-scripts` as defense against postinstall-based credential theft in the publish chain. This blocks transitive-dep postinstall scripts from running while the OIDC token is in process env. `npm run build` and `npm publish` are unaffected — those run the project's own scripts.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflows to pin GitHub Actions to specific versions for improved security and stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/439)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -167,3 +167,12 @@ updates:
         applies-to: 'version-updates'
         patterns:
           - '*'
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+    groups:
+      github-actions:
+        patterns:
+          - '*'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -167,12 +167,3 @@ updates:
         applies-to: 'version-updates'
         patterns:
           - '*'
-
-  - package-ecosystem: 'github-actions'
-    directory: '/'
-    schedule:
-      interval: 'weekly'
-    groups:
-      github-actions:
-        patterns:
-          - '*'
--- a/.github/workflows/apply_pr_checks.yaml
+++ b/.github/workflows/apply_pr_checks.yaml
@@ -28,20 +28,20 @@ jobs:
     if: needs.check_for_server_changes.outputs.result == 'true' || needs.check_for_client_changes.outputs.result == 'true'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
       - name: Check generated GraphQL is up to date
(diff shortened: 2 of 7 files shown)


cassidyjames:
I actually wonder if we should use this as an opportunity to get rid of the checklist action while we're here; it doesn't seem super useful in practice and seems like a combination of a PR template plus actual ci tests make more sense.

.github/dependabot.yml
```diff
@@ -167,3 +167,12 @@ updates:
...
           - '*'
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+    groups:
+      github-actions:
+        patterns:
+          - '*'
```
juanmrad: we have specifically opted out of major bumps for now while we work on fixing all dependencies and start upgrades on packages. Dependabot it is still too noisy for us 😅

juanmrad (approved):
One thing only to not add dependabot updates on major for now. otherwise LGTM

.github/dependabot.yml
haileyok: oh lol yea, makes sense.

cassidyjames: Just capturing the suggestion to make it easier to commit. 😄 

```suggestion
```

cassidyjames:
Filed #448 as a follow-up so we don't lose track of it

haileyok:
I don't have merge perms on this repo, but should be good to merge now (thanks @cassidyjames for removing that!)

PR #444, opened 2026-05-13
Surface users in investigation tool even when no Submitted item available

### LINKED ISSUE #437: Allow User investigations with partial data
When on the investigation tab. If you are trying to investigate a user an all of their posts we do not display data if the user item itself has not been submitted.

We should allow construction of a user/creator from data currently available such that it can be investigated for all their reports even if the User/creator has not been submitted fully.

juanmrad:
Fixes #437 

The investigation tool returned **Item Not Found** for user IDs that had clearly been actioned through Coop. Root cause is a data‑model asymmetry: actions and `/reports` submissions can reference a user via a `creatorId` field on a content/thread item, but the investigation tool only looked up direct `UserItem` submissions in `item_submission_by_thread`. Users that have never been submitted as a top-level item, but are referenced as creators were invisible.

For most workflows on Coop we expect user items to also be submitted as to keep a record of the user, but this is not always the case.

This PR adds a fallback path: when no direct submission is found, we synthesize a minimal `UserItem` from indirect references and flag it with `isSynthetic: true` so the UI can clearly indicate it's a derived record.

## Tests

- `synthesizeUserItemFromCreatorReferences.test.ts`
- `investigation.resolver.test.ts`
- `ClickhouseActionExecutionsAdapter.test.ts` 
- `ClickhouseContentApiRequestsAdapter.test.ts`
- 
Screenshot 2026-05-11 at 9 57 07 PM

## Summary by CodeRabbit

* **New Features**
  * Synthetic item submissions are now created from creator references when direct data is unavailable, with alerts indicating these in investigations.
  * Enhanced user identity inference from action history and API requests to better support incomplete data scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/444)

### DIFF
--- a/client/eslint.config.mjs
+++ b/client/eslint.config.mjs
@@ -1,6 +1,6 @@
 import { createRequire } from 'node:module';
-import { FlatCompat } from '@eslint/eslintrc';
 import { fixupConfigRules } from '@eslint/compat';
+import { FlatCompat } from '@eslint/eslintrc';
 
 const require = createRequire(import.meta.url);
 const { ignorePatterns: _, ...legacyConfig } = require('./.eslintrc.cjs');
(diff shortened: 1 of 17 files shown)


juanmrad:
@coderabbitai review

ThatKoffe (approved):
Completely tested and works 👍

julietshen (approved):

PR #455, opened 2026-05-13
client: always use relative GraphQL URL

### LINKED ISSUE #452: Absolute GraphQL URL breaks Codespace demoing
Found this when trying to set up a Codespace demo. The client has `http://localhost:3000` hardcoded when not using the production environment. I'm not sure why that would be the case, instead of just always using the relative path?

https://github.com/roostorg/coop/blob/34e9c0df3e40b5844c431ce54084192191c1c784/client/src/index.tsx#L46-L51

cassidyjames:
The hardcoded localhost fallback breaks environments like Codespaces where the app isn't served from localhost. The relative path works in all environments.

Fixes #452

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved inconsistent API endpoint configuration to ensure reliable connectivity across all environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/455)

### DIFF
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -44,10 +44,7 @@ gql`
 `;
 
 const httpLink = new HttpLink({
-  uri:
-    import.meta.env.MODE === 'production'
-      ? '/api/v1/graphql'
-      : 'http://localhost:3000/api/v1/graphql',
+  uri: '/api/v1/graphql',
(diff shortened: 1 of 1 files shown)


juanmrad (approved):
👍🏻

PR #474, opened 2026-05-15
fix: Move NCMEC routing of reports to env config

juanmrad:

Fix bug coming from SaaS era. Move NCMEC routing of reports and display to an env variable to determine whether submissions go to NCMEC's sandbox (`exttest.cybertip.org`) or production (`report.cybertip.org`).

Previously, this was gated by a hardcoded allowlist of MRT queue UUIDs in `server/services/ncmecService/ncmecService.ts` (`ncmecProdQueues`). Adding a new tenant required a code change. That doesn't make sense for a self-hosted OSS project each deployment running Coop, not the maintainers, should decide whether their deployment talks to prod.

## Summary by CodeRabbit

* **Documentation**
  - Clarified NCMEC integration setup, credential distinctions, manual review queue configuration, and expanded access/roles and settings tables. Improved examples and response-contract guidance: responses must include entries for every requested user/media; if all media are marked missing the job is treated as a permanent error and will not be retried.

* **Configuration Enhancement**
  - Added NCMEC_ENV to control submission endpoint: set to "production" for live reporting; otherwise uses the sandbox/test endpoint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/474)

### DIFF
--- a/docs/NCMEC.md
+++ b/docs/NCMEC.md
@@ -29,7 +29,7 @@ Both sets of credentials must be obtained from NCMEC by [registering as an ESP](
 
 3. **NCMEC org settings configured**: set via **Settings → NCMEC** (Admin only). See [NCMEC Settings](#ncmec-settings) below.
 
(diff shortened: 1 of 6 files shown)


julietshen (approved):
good catch 😭

PR #475, opened 2026-05-15
[fix] Display report information on other reports table

### LINKED ISSUE #450: Number of reports missing information
Image

We are missing number of reports. We also duplicate show the same item on "additional items" field

juanmrad:

Fixes #450

When a job had multiple merged reports, the "N other reports" table on the manual review screen rendered column headers but no rows, even though the count above said reports existed.

Root cause: `MergedReportsComponent` filtered the merged report list by `reportedAt !== job.createdAt` to drop the primary report (which is shown separately above). That filter is fragile given that several reports share the same `reportedAt` (common in tests, scripted submissions, or bursty real traffic where multiple users report within the same second), every entry whose timestamp matched the original report's reportedAt got filtered out, leaving an empty table.

Some minor improvements:
- "Reported By" now shows System when a report has no reporterId (rule-engine, NCMEC, system enqueues) instead of a misleading Unknown. Falls back to the raw reporter id, then Unknown reporter, only when there was a reporterId we couldn't resolve. The investigation external-link icon is hidden for system reports.
- "Reported For" now renders the policy name as a link to /dashboard/policies/form/<policyId> (same destination PoliciesDashboard uses), opened in a new tab with the same external-link icon pattern as "Reported By".
- Empty Reported For / Reason render as — instead of blank cells.

 
The table now spans the full width of its container and the wrapper has symmetric padding so the Hide button isn't flush against the edge.

After:
Screenshot 2026-05-14 at 8 01 08 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report identification and filtering in the manual review interface by using a specific primary report identifier
  * Enhanced display of report metadata including associated policies with clearer policy links
  * Refined handling of missing reasons and system-generated reports with clearer labeling
  * Fixed reporter investigation link resolution so links appear only when reporter info is present
  * Improved table layout and container styling for more consistent display

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/475?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
@@ -831,7 +831,7 @@ function ManualReviewJobReviewImpl(props: {
   const otherReports =
     reportHistory.length > 1 ? (
       <MergedReportsComponent
-        primaryReportedAt={job.createdAt}
+        primaryReportId={reportHistory[0]?.reportId}
         reportHistory={reportHistory}
       />
     ) : null;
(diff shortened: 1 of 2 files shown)


julietshen (approved):

julietshen:
thank you! much better :) great catch

PR #479, opened 2026-05-15
[Fix] Provide recovery script on case of accidental queue cleanup

### LINKED ISSUE #146: Clearing out Redis orphans manual review jobs
MRT stores job metadata in both Postgres (`job_creations`) and Redis (BullMQ). 

**This might be a weird state I got into and most people might not ever do this,** but I flushed Redis to try and reset things when I accidentally overloaded Coop with too many tasks. If Redis is flushed, jobs exist in Postgres but not in the queue, and `pendingJobCount` returns 0 while `job_creations` still has rows. 

It leaves the manual review jobs in an orphaned state.  There's no consistency check between the two stores. A reconciliation mechanism or single source of truth would prevent this.

juanmrad:
Fixes #146

We want to provide a way to recover queue data in case of accidental clearing due to redis or user error. Pending MRT jobs only live in Redis, so once the queue is wiped the items are gone. Postgres records
*which* items were enqueued, but not the bodies, report history, or NCMEC media lists.

Deleting all jobs in a queue is now restricted to org admins instead of moderator managers, on both the server and the client. The confirmation modal also asks the admin to type the queue name before the Confirm button is enabled.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI recovery tool to restore manual-review queue items after Redis/job loss.

* **Improvements**
  * "Delete All Jobs" modal now requires typing the exact queue name, shows explicit irreversibility warnings, and resets input when changing targets.
  * Delete actions and delete-column visibility are now limited to org admins.
  * Queue ID cells now include a copy control for easier copying.

* **Documentation**
  * Added recovery tool documentation and usage guidance.

* **Tests**
  * Added authorization tests for the delete-all-jobs flow.

* **Infrastructure**
  * Updated Docker/Redis configuration for improved stability and resource limits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/479?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
+++ b/client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
@@ -356,10 +356,11 @@ export default function ManualReviewQueuesDashboard() {
     queues,
     deleteAllJobsModalInfo?.id,
   );
-  // Trim guards against trailing whitespace from copy/paste; comparison is exact.
+  // Trim both sides; the comparison is otherwise exact (case- and
+  // whitespace-sensitive in the middle of the string).
   const deleteAllJobsConfirmEnabled =
     deleteAllJobsTargetName != null &&
-    deleteAllJobsConfirmText.trim() === deleteAllJobsTargetName;
+    deleteAllJobsConfirmText.trim() === deleteAllJobsTargetName.trim();
   const deleteAllJobsModal = (
     <CoopModal
       title={
@@ -557,6 +558,10 @@ export default function ManualReviewQueuesDashboard() {
               canSort: false,
             }
           : undefined,
+        previewJobsViewEnabled &&
+        userHasPermissions(data?.me?.permissions, [
+          GQLUserPermission.EditMrtQueues,
+        ]) &&
         columnVisibility.previewJobs
           ? {
               Header: '',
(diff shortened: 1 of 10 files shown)


client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
```diff
@@ -348,8 +388,27 @@ export default function ManualReviewQueuesDashboard() {
...
-      Are you sure you want to delete all jobs from this queue? You can't undo
-      this action. For larger queues, this can take a few minutes.
+      <div className="space-y-3">
+        <p>
+          You are about to delete all jobs from{' '}
+          <strong>{deleteAllJobsTargetName ?? 'this queue'}</strong>. This wipes
+          the queue&apos;s pending items in Redis and{' '}
+          <strong>cannot be undone from the UI</strong>. Recovery is only
+          possible by re-running the items through{' '}
+          <code>npm run recover-mrt-queue</code> on the server.
```
cassidyjames, line 398: I don't _love_ having this level of back-end technicality in the front-end UI, but maybe it's appropriate here.

cassidyjames (approved):
Approving considering the recovery script has been used successfully. I left one comment on a front-end change but don't feel super strongly about it; maybe it's better as a follow-up issue.

client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
juanmrad: I was a bit torn on this. Part of me is that now cleaning up the queue is admin only. and may be good to show this is what we have.

PR #505, opened 2026-05-17
fix: NCMEC wellness permission check

### LINKED ISSUE #504: External Moderator can save NCMEC and org wellness settings
Steps to reproduce:
1. Create a new user in your org as "External Moderator"
2. Login as the new user
3. Click on the settings button
4. Click on NCMEC or Wellness settings
5. Able to view and save the settings

serendipty01:
fixes #504 

## Tests

Tested manually

# RCA: External Moderators Can Access Settings Pages (Claude)
---

## Validated Behavior per Settings Page (pre-fix)

| Page | Observed behavior | Root cause | Classification |
|------|------------------|------------|----------------|
| Item Types | Visible, create disabled | `canEditItemTypes` flag disables button; no redirect | ✅ Intentional (read-only) |
| Actions | Visible, create disabled | `canEditActions` flag disables button; no redirect | ✅ Intentional (read-only) |
| API Keys | Redirects to Item Types | Guard → `navigate('/settings')` → App.tsx `/settings` → `/dashboard/settings/item_types` (3-hop chain) | ✅ Working, indirect |
| Integrations | Visible, save disabled | `canEditConfig` flag disables save button; no redirect | ✅ Intentional (read-only) |
| Users | "Something Went Wrong" | Server rejects query; `throw error` fires **before** permission check → ErrorBoundary | ⚠️ Poor UX |
| Wellness | Fully accessible & writable | No permission guard in component | ❌ Security bug |
| Organization | Redirects to Item Types | Same 3-hop redirect chain as API Keys | ✅ Working, indirect |
| NCMEC Settings | Fully accessible & writable | No permission guard in component | ❌ Security bug |
| SSO Settings | Not in sidebar; blank page if accessed directly | Removed from `settingsMenuItems`; route still live; `if (error) return <div />` fires before permission check; missing `return null` after `navigate()` | ⚠️ Incomplete |

---

### Bug 1 — Sidebar shows ALL settings links regardless of permissions

`Sidebar.tsx` has two rendering paths:

**Main navigation** (`menuItems`, line 305) — uses `recursiveMenuItems` (line 133), which checks `requiredPermissions` at lines 138-143 and returns `null` for inaccessible items. ✅ Correct.

**Settings panel** (`settingsMenu`) — used a raw `.map()` on `settingsMenuItems[0]?.subItems` with **no permission check**. Every sub-item (NCMEC Settings, Wellness, SSO, etc.) was rendered for every user regardless of role.

Additionally, `isSettingsSelected` checked the **unfiltered** subItems list to decide whether to open the panel. This meant the panel would technically open (expanding to max-height) even when it contained zero visible items — producing a confusingly empty panel.

Finally, the Settings gear icon had no visibility condition tied to accessible items, so it was always shown even when no settings were accessible.

**Secondary issue — `requiredPermissions` mismatch:** Item Types, Actions, and Integrations were declared with `requiredPermissions: [ManageOrg]` in `settingsMenuItems`, but their components intentionally allow read access for all authenticated users (only editing is restricted). Once the sidebar filter was applied correctly, External Moderators lost sidebar access to these pages entirely.

---

### Bug 2 — `NCMECSettings` and `OrgSafetySettings` have no component-level guard

Neither component imported `userHasPermissions` nor checked permissions. Any authenticated user navigating directly to the URL got full read and write access. Both queries also lacked `me { permissions }`, making it impossible to check permissions even if the guard were added.

- `NCMECSettings.tsx` — queried only `ncmecOrgSettings` and `myOrg`
- `OrgSafetySettings.tsx` — queried only `myOrg`

---

### Bug 3 — `ManageUsers` throws before checking permissions → "Something Went Wrong"

The server-side resolver for `ManageUsers` rejects the query for non-`ManageOrg` users, setting `error`. The original component ordering was:

```
if (loading) return <FullScreenLoading />;
if (error) { throw error; }          ← server rejection lands here
const permissions = data!.me?.permissions;
if (!userHasPermissions(...)) { navigate('/settings') }  ← never reached
```

The `ErrorBoundary` in `Dashboard.tsx` caught the thrown error and showed "Something Went Wrong" rather than a clean redirect.

---

### Bug 4 — `SSOSettings` has two issues

`SSO Settings` was removed from `settingsMenuItems` on this branch (no sidebar link), but the route `/dashboard/settings/sso` remains registered and the page loads if navigated to directly.

**Issue A:** `if (error) return <div />` fired before the permission check. If the server rejects the query for non-admins, the component silently returned an empty `<div />` (blank page) before the guard ran.

**Issue B:** The permission check called `navigate('/settings')` with no `return` after it. React Router v6's `navigate()` schedules a redirect but does not halt the current render. Without `return null`, the full SSO form (org SSO URL, certificate, ACS callback URL) rendered and was briefly visible before the navigation fired.

---

### Note — `navigate('/settings')` works but is indirect

`App.tsx` (line 149-151) registers a top-level `/settings` route:
```typescript
{ path: 'settings', element: <Navigate replace to="/dashboard/settings" /> }
```
So the chain `navigate('/settings')` → `/dashboard/settings` → `/dashboard/settings/item_types` works, but requires three hops. Direct `navigate('/dashboard/settings')` is cleaner.

---

### Commit 1 — `fix: add ManageOrg permission guards to settings pages`

| File | Change |
|------|--------|
| `NCMECSettings.tsx` | Added `GQLUserPermission`, `userHasPermissions`, `useNavigate` imports; added `me { permissions }` to `NcmecOrgSettings` GQL query; added `ManageOrg` guard with `return null` after loading check |
| `OrgSafetySettings.tsx` | Same as above for `OrgDefaultSafetySettings` query |
| `ManageUsers.tsx` | Permission check moved **before** `throw error`; `data!` → `data?` (safe optional chaining); added `return null`; `navigate('/settings')` → `navigate('/dashboard/settings')` |
| `SSOSettings.tsx` | Permission check moved **before** `if (error) return <div />`; added `return null` after `navigate`; `navigate('/settings')` → `navigate('/dashboard/settings')` |
| `ApiAuthenticationSettings.tsx` | `navigate('/settings')` → `navigate('/dashboard/settings')` |
| `OrgSettings.tsx` | `navigate('/settings')` → `navigate('/dashboard/settings')` |
| `generated.ts` | Regenerated via `pnpm generate` to include `me { permissions }` in the two new queries |

### Commit 2 — `fix: hide restricted settings links from External Moderators in sidebar`

| File | Change |
|------|--------|
| `Sidebar.tsx` | Extracted `accessibleSettingsSubItems` useMemo (filters `subItems` by `requiredPermissions` against current user's permissions); used in three places: panel render, `isSettingsSelected`, and gear icon visibility condition. Settings gear now hidden entirely when no items are accessible. |

---

## Post-fix Expected Behavior (External Moderator)

| Page | Expected behavior |
|------|------------------|
| Settings gear | Hidden (no accessible settings sub-items for External Moderators) |
| Item Types | Not in sidebar; page still renders read-only if accessed directly |
| Actions | Not in sidebar; page still renders read-only if accessed directly |
| API Keys | Not in sidebar; redirects to `/dashboard/settings` if accessed directly |
| Integrations | Not in sidebar; page still renders read-only if accessed directly |
| Users | Not in sidebar; redirects to `/dashboard/settings` if accessed directly |
| Wellness | Not in sidebar; redirects to `/dashboard/settings` if accessed directly |
| Organization | Not in sidebar; redirects to `/dashboard/settings` if accessed directly |
| NCMEC Settings | Not in sidebar; redirects to `/dashboard/settings` if accessed directly |
| SSO Settings | Not in sidebar; redirects to `/dashboard/settings` if accessed directly |

---

## Design Note — Sidebar `requiredPermissions` vs Component Guards

These are two separate permission layers with different purposes:

- **`requiredPermissions` in `settingsMenuItems`** — controls **sidebar discoverability** (does the link appear?). Set to `[]` for pages that are intentionally read-accessible to all users.
- **Component-level `userHasPermissions` guard** — controls **route access** (can the page render?). Required for every settings page that performs org-admin operations, regardless of sidebar visibility. A user can always navigate directly to a URL.

---

### Commit 3 — `fix: enforce ManageOrg permission server-side for NCMEC and safety settings`

All four server resolvers previously only checked authentication (`if (!user)`) but not authorization (`ManageOrg`). Any authenticated user could read NCMEC credentials or overwrite org-wide safety settings via direct GraphQL calls, completely bypassing the frontend guards.

Pattern used (matching existing resolvers like `pendingInvites` and `updateOrgInfo`):
```typescript
if (!user.getPermissions().includes('MANAGE_ORG')) {
  throw forbiddenError('...');
}
```

| File | Resolver | Fix |
|------|----------|-----|
| `server/graphql/modules/ncmec.ts` | `ncmecOrgSettings` query | Added `MANAGE_ORG` check — External Moderators can no longer read NCMEC credentials |
| `server/graphql/modules/ncmec.ts` | `updateNcmecOrgSettings` mutation | Added `MANAGE_ORG` check |
| `server/graphql/modules/org.ts` | `defaultInterfacePreferences` Org field | Added `getUser()` call + `MANAGE_ORG` check (field resolver had no auth check at all) |
| `server/graphql/modules/org.ts` | `setOrgDefaultSafetySettings` mutation | Added `MANAGE_ORG` check |

---

## Open Items

- **SSO Settings route:** The route `/dashboard/settings/sso` is still registered but has no sidebar link. If SSO settings are being reworked, consider whether the route should be kept, removed, or redirected.

## Summary by CodeRabbit

* **Security & Permissions**
  * Org-management (ManageOrg) checks now block access on server and client; restricted settings operations and pages redirect to /dashboard/settings and won’t render without proper access. GraphQL queries now include the current user’s permissions to enforce these gates.

* **UI**
  * Sidebar and footer only show Settings when at least one settings sub-item is permitted; Settings sub-items are filtered to the user’s permissions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/505?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


serendipty01:
@juanmrad i have updated the PR

juanmrad (approved):
LGTM

PR #533, opened 2026-05-20
Server-side enforcement of requires_policy_for_decisions

### LINKED ISSUE #389: Require certain inputs for a decision
It would be useful if an organization could require that a moderator provide certain inputs before submitting a decision.  At Upscrolled, we require a moderator to select the specific policy that was violated and then add a comment explaining their rationale to the report.

Currently, there is no way to enforce this requirement through Coop so it is possible for decisions to be made outside of our moderation policy through human error.  Ideally, an org would have a way of specifying specific information that it wants a moderator to fill out.

julietshen:
- fixes #389

Closes the API-bypass gap for `requires_policy_for_decisions`. The flag already exists in `manual_review_tool.manual_review_tool_settings`, is exposed via `Org.requiresPolicyForDecisionsInMrt`, and the MRT UI already disables submit client-side when the flag is on and no policy is selected (`client/.../ManualReviewJobReview.tsx:564`). What was missing: nothing actually rejected a policy-less `CUSTOM_ACTION` decision if it reached the server via API or scripted client. This PR adds that one check.

What's new:

- `JobDecisioning.submitDecision` — after the existing action-id validation, reads `manual_review_tool.manual_review_tool_settings.requires_policy_for_decisions` for the org. If `true`, any `CUSTOM_ACTION` decision component with an empty `policies` array throws a new `MissingRequiredPolicyForDecisionError` (400, `InvalidUserInput`).
- Other decision kinds (`IGNORE`, `REJECT_APPEAL`, `ACCEPT_APPEAL`, `SUBMIT_NCMEC_REPORT`, `TRANSFORM_JOB_AND_RECREATE_IN_QUEUE`, `AUTOMATIC_CLOSE`) don't carry policies and are out of scope.
- Tests: two cases in `manualReviewToolService.test.ts` covering rejection and the happy path with the flag on, using the staging-seeded org/queue.

What's deliberately not in this PR:

- **No new migration, no new GraphQL field, no new IoC plumbing.** The settings column, the service read, and the GraphQL exposure all already existed — this PR just adds the missing server-side check.
- **No mutation/admin UI to flip the flag.** Today orgs toggle it via SQL, same as `allow_multiple_policies_per_action`. An admin settings page is a separate piece of work.
- `requires_decision_reason` enforcement (the comment-required flag Jess also asked about in #389) — that has its own column and is a clean follow-up with the same shape.

## Tests

Existing test file `services/manualReviewToolService/manualReviewToolService.test.ts` runs against the live staging-seeded stack and now includes:

```
requires_policy_for_decisions enforcement
  ✓ rejects a CUSTOM_ACTION decision with no policies when the flag is on
  ✓ allows a CUSTOM_ACTION decision with policies when the flag is on
```

Local run:

```bash
npm run up
npm run db:update -- --env staging
(cd server && npm run test:prepush -- services/manualReviewToolService/manualReviewToolService.test.ts -t requires_policy_for_decisions)
```

The describe block seeds its own queue row idempotently and resets the flag in `afterEach` so other tests aren't affected.

Reviewer checks:

- [ ] `(cd server && npx tsc --noEmit)` — exit 0
- [ ] `(cd server && npm run lint -- services/manualReviewToolService/)` — no new errors on `JobDecisioning.ts` / the test file
- [ ] `(cd server && npm run test:prepush -- services/manualReviewToolService/manualReviewToolService.test.ts -t requires_policy_for_decisions)` — 2 passing
- [ ] Behavior unchanged for orgs with the flag off (default)

## (Optional) Rollout Plan

No schema change, no GraphQL change, no client change. Existing orgs see no behavior difference unless they've already opted into `requires_policy_for_decisions`. Orgs that have opted in stop being able to bypass the policy requirement via the API.

## Summary by CodeRabbit

* **Tests**
  * Added integration tests covering enforcement of the "requires policy for decisions" org setting.

* **Bug Fixes**
  * When the org setting is enabled, custom-action decisions without policies now return a clear validation error; behavior unchanged when disabled.

* **New Features**
  * Submission flow now surfaces a modal prompting reviewers to choose at least one policy when required.

### DIFF
--- a/server/services/manualReviewToolService/modules/JobDecisioning.ts
+++ b/server/services/manualReviewToolService/modules/JobDecisioning.ts
@@ -196,6 +196,33 @@ export default class JobDecisioning {
       if (validActions.length === 0) {
         throw makeSubmittedJobActionNotFoundError({ shouldErrorSpan: true });
       }
+
+      // Issue #389: orgs that opt into `requires_policy_for_decisions` enforce
+      // that every CUSTOM_ACTION decision names at least one policy, so
+      // reviewers can't accidentally land actions outside the org's documented
+      // moderation policy. The flag is read via GraphQL as
+      // `Org.requiresPolicyForDecisionsInMrt`; the UI can disable submit on
+      // the client, but server-side enforcement is what makes the integrity
+      // guarantee real. Only CUSTOM_ACTION decisions carry policies —
+      // appeal/ignore/NCMEC/auto-close decisions are out of scope.
+      const settingsRow = await this.pgQuery
+        .selectFrom('manual_review_tool.manual_review_tool_settings')
+        .select(['requires_policy_for_decisions'])
+        .where('org_id', '=', orgId)
+        .executeTakeFirst();
+      if (settingsRow?.requires_policy_for_decisions) {
+        const customActionDecisionMissingPolicy = decisions.find(
+          (decision) =>
+            decision.type === 'CUSTOM_ACTION' && decision.policies.length === 0,
+        );
+        if (customActionDecisionMissingPolicy != null) {
+          throw makeMissingRequiredPolicyForDecisionError({
+            detail:
+              'This org requires every decision to include at least one policy.',
+            shouldErrorSpan: true,
+          });
+        }
+      }
     }
 
     const removeJob = async () =>
@@ -593,7 +620,8 @@ export type SubmitDecisionErrorType =
   | 'JobHasAlreadyBeenSubmittedError'
   | 'SubmittedJobActionNotFoundError'
   | 'NoJobWithIdInQueueError'
-  | 'RecordingJobDecisionFailedError';
+  | 'RecordingJobDecisionFailedError'
+  | 'MissingRequiredPolicyForDecisionError';
 
 export const makeJobHasAlreadyBeenSubmittedError = (data: ErrorInstanceData) =>
   new CoopError({
@@ -628,5 +656,17 @@ export const makeNoJobWithIdInQueueError = (data: ErrorInstanceData) =>
     type: [ErrorType.NotFound],
     title: String(data.detail),
     name: 'NoJobWithIdInQueueError',
+    ...data,
+  });
+
+export const makeMissingRequiredPolicyForDecisionError = (
+  data: ErrorInstanceData,
+) =>
+  new CoopError({
+    status: 400,
+    type: [ErrorType.InvalidUserInput],
+    title:
+      'This org requires every decision to include at least one policy. Pick a policy and resubmit.',
+    name: 'MissingRequiredPolicyForDecisionError',
     ...data,
   });
--- a/server/services/manualReviewToolService/modules/JobDecisioning.ts
+++ b/server/services/manualReviewToolService/modules/JobDecisioning.ts
@@ -197,14 +197,8 @@ export default class JobDecisioning {
         throw makeSubmittedJobActionNotFoundError({ shouldErrorSpan: true });
       }
 
-      // Issue #389: orgs that opt into `requires_policy_for_decisions` enforce
-      // that every CUSTOM_ACTION decision names at least one policy, so
-      // reviewers can't accidentally land actions outside the org's documented
-      // moderation policy. The flag is read via GraphQL as
-      // `Org.requiresPolicyForDecisionsInMrt`; the UI can disable submit on
-      // the client, but server-side enforcement is what makes the integrity
-      // guarantee real. Only CUSTOM_ACTION decisions carry policies —
-      // appeal/ignore/NCMEC/auto-close decisions are out of scope.
+      // Enforce `requires_policy_for_decisions` server-side. The MRT UI already
+      // disables submit when this is on, but API/script callers can bypass that.
       const settingsRow = await this.pgQuery
         .selectFrom('manual_review_tool.manual_review_tool_settings')
         .select(['requires_policy_for_decisions'])
--- a/server/services/manualReviewToolService/modules/JobDecisioning.ts
+++ b/server/services/manualReviewToolService/modules/JobDecisioning.ts
@@ -21,6 +21,7 @@ import {
   type ManualReviewJobEnqueueSourceInfo,
   type ReportHistory,
 } from '../manualReviewToolService.js';
+import type ManualReviewToolSettings from './ManualReviewToolSettings.js';
 import type QueueOperations from './QueueOperations.js';
 import { jobIdToGuid } from './QueueOperations.js';
 
@@ -141,6 +142,7 @@ export default class JobDecisioning {
     readonly onRecordDecision: (params: OnRecordDecisionInput) => Promise<void>,
     private readonly moderationConfigService: Dependencies['ModerationConfigService'],
     private readonly tracer: Dependencies['Tracer'],
+    private readonly manualReviewToolSettings: ManualReviewToolSettings,
   ) {}
 
   async submitDecision(opts: SubmitDecisionInput) {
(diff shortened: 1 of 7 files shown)


server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -196,6 +196,33 @@ export default class JobDecisioning {
...
       }
+
+      // Issue #389: orgs that opt into `requires_policy_for_decisions` enforce
+      // that every CUSTOM_ACTION decision names at least one policy, so
+      // reviewers can't accidentally land actions outside the org's documented
+      // moderation policy. The flag is read via GraphQL as
+      // `Org.requiresPolicyForDecisionsInMrt`; the UI can disable submit on
+      // the client, but server-side enforcement is what makes the integrity
+      // guarantee real. Only CUSTOM_ACTION decisions carry policies —
+      // appeal/ignore/NCMEC/auto-close decisions are out of scope.
```
juanmrad: I don't think we need this kind of detail here and will become irrelevant as the code can explain itself. we can reduce this to be more succinct and explicit. 

```suggestion
      // Enforce `requires_policy_for_decisions` server-side. The MRT UI already
      // disables submit when this is on, but API/script callers can bypass that.
```

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -630,3 +658,15 @@ export const makeNoJobWithIdInQueueError = (data: ErrorInstanceData) =>
     name: 'NoJobWithIdInQueueError',
     ...data,
   });
+
+export const makeMissingRequiredPolicyForDecisionError = (
```
juanmrad, line 657: This error is not wired on the graphql contract. we should update `SubmitDecisionResponse` with this new error.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -196,6 +196,33 @@ export default class JobDecisioning {
...
+      // `Org.requiresPolicyForDecisionsInMrt`; the UI can disable submit on
+      // the client, but server-side enforcement is what makes the integrity
+      // guarantee real. Only CUSTOM_ACTION decisions carry policies —
+      // appeal/ignore/NCMEC/auto-close decisions are out of scope.
+      const settingsRow = await this.pgQuery
+        .selectFrom('manual_review_tool.manual_review_tool_settings')
+        .select(['requires_policy_for_decisions'])
+        .where('org_id', '=', orgId)
+        .executeTakeFirst();
+      if (settingsRow?.requires_policy_for_decisions) {
```
juanmrad: This is already on `ManualReviewToolSettings.getRequiresPolicyForDecisions(orgId)` no need to have query here.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -196,6 +196,33 @@ export default class JobDecisioning {
...
+
+      // Issue #389: orgs that opt into `requires_policy_for_decisions` enforce
+      // that every CUSTOM_ACTION decision names at least one policy, so
+      // reviewers can't accidentally land actions outside the org's documented
+      // moderation policy. The flag is read via GraphQL as
+      // `Org.requiresPolicyForDecisionsInMrt`; the UI can disable submit on
+      // the client, but server-side enforcement is what makes the integrity
+      // guarantee real. Only CUSTOM_ACTION decisions carry policies —
+      // appeal/ignore/NCMEC/auto-close decisions are out of scope.
+      const settingsRow = await this.pgQuery
```
juanmrad: we shouldn't do a DB lookup if we don't need it. we can check if there are already policies. so we instead flip the check where we do:
```
const hasEmptyPolicyCustomAction = decisions.some(
  (d) => d.type === 'CUSTOM_ACTION' && d.policies.length === 0,
);
if (hasEmptyPolicyCustomAction) {
  const requiresPolicy =
    await this.manualReviewToolSettings.getRequiresPolicyForDecisions(orgId);
  if (requiresPolicy) {
    throw makeMissingRequiredPolicyForDecisionError({ ... });
  }
}
```

juanmrad (changes requested):

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: Trimmed to two-line suggestion

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: oop. Added `MissingRequiredPolicyForDecisionError` to the SDL, included it in the `SubmitDecisionResponse` union, and added it to the `isCoopErrorOfType` chain in the `submitDecision` resolver (AI Drafted)

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: Switched to `manualReviewToolSettings.getRequiresPolicyForDecisions(orgId)`. Had to thread `ManualReviewToolSettings` into `JobDecisioning` via the constructor and move its triggering above `JobDecisioning` in `manualReviewToolService.ts`. (AI Drafted)

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: Flipped to match your suggestion. check `hasEmptyPolicyCustomAction first`, only read the settings if true. So the common path (decisions already have a policy, or flag is off org-wide) does no extra DB work. (AI Drafted)

server/services/manualReviewToolService/modules/JobDecisioning.ts
juanmrad: It is still missing on `client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx` on the graphql mutation and on the case.

juanmrad (changes requested):

julietshen:
Wired both in 7defbae. added the `... on MissingRequiredPolicyForDecisionError` selection to the `SubmitManualReviewDecision` mutation and a matching `case 'MissingRequiredPolicyForDecisionError'` that surfaces the same message the server returns ("This org requires every decision to include at least one policy. Pick a policy and resubmit.").

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -196,6 +198,28 @@ export default class JobDecisioning {
       if (validActions.length === 0) {
         throw makeSubmittedJobActionNotFoundError({ shouldErrorSpan: true });
       }
+
+      // Enforce `requires_policy_for_decisions` server-side. The MRT UI already
+      // disables submit when this is on, but API/script callers can bypass that.
+      // Only check the flag when there's actually a policy-less decision to
+      // enforce against, so the common path avoids the extra DB hit.
```
juanmrad, line 206: There is something here to consider we are not guarding for which I think may be acceptable, but we are not validating the policies themselves. As your test below shows passing any id for policy gets accepted. Given the fact this is an internal tool the likelihood of a malicious intent or even being used here is pretty low.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -196,6 +198,28 @@ export default class JobDecisioning {
...
+      );
+      if (hasEmptyPolicyCustomAction) {
+        const requiresPolicy =
+          await this.manualReviewToolSettings.getRequiresPolicyForDecisions(
+            orgId,
+          );
+        if (requiresPolicy) {
+          throw makeMissingRequiredPolicyForDecisionError({
+            detail:
+              'This org requires every decision to include at least one policy.',
```
juanmrad: realized we have the same text 3 times 😅 may be good to move into a variable of sort.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -196,6 +198,28 @@ export default class JobDecisioning {
       if (validActions.length === 0) {
         throw makeSubmittedJobActionNotFoundError({ shouldErrorSpan: true });
       }
+
+      // Enforce `requires_policy_for_decisions` server-side. The MRT UI already
+      // disables submit when this is on, but API/script callers can bypass that.
+      // Only check the flag when there's actually a policy-less decision to
+      // enforce against, so the common path avoids the extra DB hit.
+      const hasEmptyPolicyCustomAction = decisions.some(
```
juanmrad: Just realized this whole code is inside:
```ts
    if (decisions.some((decision) => decision.type === 'CUSTOM_ACTION')) {
```

On line 187. so we do not need to compute the some here again. we can just check for policy length being 0.

juanmrad (approved):
approving to unblock as these are minor comments that can be addressed and then merged.

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: Tracked separately as #677, wanted to keep the all-resolve-vs-at-least-one semantics decision out of this PR

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: Dropped the duplicate detail in f591267, the factory's title is now the only source of the message

server/services/manualReviewToolService/modules/JobDecisioning.ts
julietshen: Extracted customActionDecisions once in f591267 and reused it for the empty-policy check

Issue #213, opened 2026-04-10
Work through SDLC checklist P0s and P1s


annebdh:
We now have overall [software development lifecycle guide for ROOST projects](https://github.com/roostorg/community/tree/main/software-development-practices)! 

Many of these Coop is already doing, but we should go through and address any gaps in P0s and P1s. Have listed below (details in guide on each area):

**Quality vision**
- #287

**Versioning**
- [x] Versioning
- [x] Branching

**Release cadence**
- [ ] Release cadence and roadmap planning 

**Project conventions**
- [x] README.md
- #206
- [ ] API format
- [ ] Changelog
- [x] Contributor guidance
- [x] Code of Conduct

**Testing**
- [ ] Tests for all commits
  - #288
- [ ] Release tests
- [x] CI checks

**Dependency handling**
- [x] Version alerts
- [x] Vulnerability alerts
- [ ] License scanning

**Security in CI/CD**
- [ ] Code analysis 

**Release security**
- [x] Security report intake
- [x] Incident response process

**PR triage and code review**
- [ ] Code review guidance

**Support**
- [ ] Breaking changes policy

pawiecz:
(Please let me know if it's ok to move comment below to the checklist above and I'll do so)

Items on the checklist selected sections should be considered "done" with following exceptions:

**Testing**
- Tests for all commits: (3) Integration tests might need some extension, otherwise this item is already covered
- Release tests: this item is covered only by CI checks running on tagged releases

**Dependency handling**
- License scanning: the only check currently not covered
- (P2 Vendoring policy: there are internal subpackages in `coop` but no dependency required vendoring so far)

cassidyjames:
@pawiecz if you have permissions, yes, please feel free to update the original comment above! I'm also breaking out to-be-done items as individual issues, but can wait to update the comment itself. :)

julietshen:
Changelog: https://github.com/roostorg/coop/blob/cassidyjames/changelog/CHANGELOG.md

julietshen:
removing from 1.0 milestone as we've added the prioritized requirements and will continue to add test coverage and tweak our SDLC requirements

taobojlen:
@cassidyjames / @julietshen this ticket is still marked as a P0. it has a pretty large set of nested sub-issues but i would argue that none of the open ones are actually all that important / urgent any longer!

julietshen:
Agree on lowering the priority!

ThisIsMissEm:
[...]

PR #534, opened 2026-05-21
Add OpenAI image moderation signals via omni-moderation-latest (#518)

### LINKED ISSUE #518: [OpenAI] Moderation api is missing image/media inputs
Signals dont yet support image/media inputs, only text for the OpenAI intergration

julietshen:
- fixes #518

The OpenAI moderation integration was text-only. OpenAI's `omni-moderation-latest` model scores both text and images, with overlapping category coverage. This PR migrates the lib to omni and adds image signal classes for the four categories that already had text counterparts (`violence`, `violence/graphic`, `self-harm`, `sexual`).

What's new:

- **Lib** (`openAIModerationUtils.ts`) — switched to `omni-moderation-latest`, rebuilt the request body as a multimodal array (`{ model, input: [{type: 'text', text} | {type: 'image_url', image_url: {url}}] }`), added `runOpenAiModerationImageImpl` parallel to the existing text impl. OpenAI fetches the image from the URL itself, so we pass the signal input's URL through with no buffering.
- **Four new image signal classes**: `OpenAi{Violence,GraphicViolence,SelfHarm,Sexual}ImageSignal` — mirror the text-signal pattern but take `ScalarTypes['IMAGE']` and route through the image impl.
- **Plumbing**: matching `SignalType` enum values, `integrationForSignalType` switch cases, `SignalArgsByType` / `RuntimeSignalArgsByType` entries, and IoC registration in `instantiateBuiltInSignals`.
- **Type guard**: `OpenAiImageModelName` is `Extract<OpenAiModelName, ...image-supported categories...>`, so callers of the image impl can't accidentally request a category OpenAI only scores against text.
- **Error classification** (per GitHub Action review prompt + project convention): the existing `'Missing API credentials'` checks in both impls and the new `'must include text or imageUrl'` check now throw `SignalPermanentError` rather than a regular `Error`, since none of those states can recover on retry within a request window. `'Empty OpenAI results'` and non-2xx HTTP responses keep their pre-existing classification (regular Error, retryable) — properly classifying 4xx/5xx is a wider cleanup not in scope here.

Notes worth flagging:

- **Existing text signals keep working.** The omni model still scores all the legacy text categories (`hate`, `hate/threatening`, `self-harm`, `sexual`, `sexual/minors`, `violence`, `violence/graphic`); `category_scores[name]` extraction is unchanged. Pricing is unaffected per OpenAI's posted rates.
- **No new text-only omni categories** (`harassment`, `harassment/threatening`, `illicit`, `illicit/violent`). Same pattern, separate ask.
- **`self-harm/intent` and `self-harm/instructions` (text + image) are in #535** — stacked PR. Split out so this one stays focused on resolving #518.
- **No GraphQL or client UI changes** — signals are registered server-side via `instantiateBuiltInSignals`. Per existing convention, they become available to rules without a separate codegen step.

## Tests

- **New**: `openAIModerationUtils.test.ts` pins the multimodal request body shape (one test for text-only, one for image-only). This util had no test prior; per SDLC issue #213 ("Tests for all commits"), new behavior should be covered. Kept narrow — just the body shape, which is the actual contract change with OpenAI's API.

I also tested in the UI and made sure the new image signals show up in the signal library

Screenshot 2026-05-20 at 11 22 22 PM
Screenshot 2026-05-20 at 11 22 27 PM

Local:

```bash
(cd server && npx tsc --noEmit)
(cd server && npm run lint -- services/signalsService/)
(cd server && npm run test:prepush -- services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.test.ts)
```

Reviewer checks:

- [ ] `(cd server && npx tsc --noEmit)` — exit 0
- [ ] `(cd server && npm run lint -- services/signalsService/)` — no errors on changed files
- [ ] `(cd server && npm run test:prepush -- services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.test.ts)` — 2 passing
- [ ] Spot-check that existing text signals still extract the right `category_scores[name]` (no test regression — same shape as before)

## (Optional) Rollout Plan

No schema change, no GraphQL change, no client change. Server starts registering 4 new built-in image signal types on next deploy. Orgs without OpenAI credentials still see them as disabled (existing `openAiModerationGetDisabledInfo` flow). Orgs with credentials can wire the new signals into their rules.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI image moderation signals (graphic violence, self-harm, sexual, violence), and expanded moderation utilities to handle images alongside text; signals are available for automated rules.

* **Tests**
  * Added tests validating multimodal OpenAI moderation request contracts (text/image) and GraphQL SignalType enum coverage.

* **Documentation**
  * Updated contributor guidance for adding new built-in signal types.

* **Chores**
  * Expanded signal type mappings and client/server GraphQL enums to include new OpenAI image model identifiers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/534?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


julietshen:
**Two follow-up commits to flag (`fc6af84` and `83fd504`)**

**`fc6af84` — Classify error cases as `SignalPermanentError`**

Per project convention (any case where the signal cannot determine a score for the input should be a `SignalPermanentError`, so the rejection gets cached and not retried), upgrade two error paths from `throw new Error(...)`:

- **Missing API credentials** (both `runOpenAiModerationImpl` and the new `runOpenAiModerationImageImpl`) — the org has no API key, retries on the same call same org always fail the same way.
- **`text` + `imageUrl` both nil** in `getOpenAiModerationScores` — caller bug; with neither input there is nothing to score and a retry yields the same failure. The typed entry points always pass one or the other, but the defensive check is now correctly classified.

Not changed in this commit: `'Empty OpenAI results'` and non-2xx HTTP responses keep the existing classification (regular `Error` → retryable without caching). Both pre-date this PR and could reasonably go either way; differentiating 4xx/5xx in particular is a wider cleanup.

**`83fd504` — Expose new signals to GraphQL + client, add coverage test**

The first commit registered the four new image signals server-side via `instantiateBuiltInSignals` and `SignalArgsByType`, but the GraphQL `SignalType` enum and the client-side `integrationForSignalType` switch are hand-maintained mirrors that were missed. Without these, the new signals register at runtime but never show up in the dashboard.

- `server/graphql/modules/signal.ts`: add the four `OPEN_AI_*_IMAGE_MODEL` types to the SDL enum.
- `client/src/models/signal.ts`: add the four types to the `OpenAi` case in `integrationForSignalType` so the UI tags them correctly.
- Regenerate `generated.ts` via `npm run generate`.

To prevent this regression class on future signal additions:

- New coverage test in `signal.test.ts` asserts every `BuiltInExternalSignalType` and `BuiltInThirdPartySignalType` value appears in the GraphQL `SignalType` enum string. CI fails when a new TS signal type isn't mirrored into the SDL.
- `AGENTS.md` gains an "Adding a new built-in `SignalType`" checklist listing the four files that must agree.
- Also export `BuiltInExternalSignalType` from the service barrel so the test can reach it through the public index (the restricted-paths rule disallows deep imports into service internals).

Follow-up issue #536 tracks tightening `integrationForSignalType` to use typed exhaustive switches — strongest guarantee but a wider refactor since the function currently accepts plugin signal types via a permissive `string` parameter.

**Stacked PR #535** rounds out the self-harm coverage on top of this — adds text + image signal classes for `self-harm/intent` and `self-harm/instructions` (four new classes), which omni-moderation scores in both modalities. Split out so this PR stays focused on resolving #518; the same GraphQL/client/coverage plumbing was applied there too so it doesn't reintroduce the regression class.

🤖 _Drafted by Claude Code._

AGENTS.md
```diff
@@ -186,9 +196,11 @@ Stop and get explicit human approval before:
 ## Commit attribution
 
 Agent-authored commits should include a `Co-Authored-By` trailer naming the agent, e.g.:
+
 ```
 Co-Authored-By: <agent-name>
 ```
```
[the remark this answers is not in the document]
julietshen: Addressed in `69a1543` — added `text` language identifier to the fenced block.

🤖 _Drafted by Claude Code._

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  it('throws a SignalPermanentError when neither text nor imageUrl is provided', async () => {
+    // Defensive guard — the typed entry points always pass one or the other,
+    // so this only fires on caller bugs. Classified as permanent because a
+    // retry with the same arguments yields the same failure.
+    const { fetchHTTP } = makeFetchHTTPCapturing();
+    await expect(
+      getOpenAiModerationScores(fetchHTTP, tracer, { apiKey: 'sk-test' }),
+    ).rejects.toThrow(/text.*imageUrl/);
+  });
```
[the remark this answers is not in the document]
julietshen: done

server/graphql/modules/signal.ts
```diff
@@ -202,23 +211,28 @@ const Signal: GQLSignalResolvers = {
   },
   integrationTitle(signal) {
     if (signal.integration == null) return null;
-    return getIntegrationRegistry().getManifest(signal.integration)?.title ?? null;
+    return (
+      getIntegrationRegistry().getManifest(signal.integration)?.title ?? null
```
ThisIsMissEm, line 215: Might be worth checking this is memoized as a query that asks for each property will be calling `getIntegrationRegistry()` at least three times, especially if that method is expensive.

ThisIsMissEm (commented):
There's a bunch of duplicated code between the various signal types, but maybe that's consistent with the rest of the codebase. LGTMO

server/services/signalsService/signals/third_party_signals/open_ai/moderation/OpenAiViolenceImageSignal.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  override get outputType() {
+    return { scalarType: ScalarTypes.NUMBER };
+  }
+
+  /**
+   * Placeholder estimate
+   */
+  override getCost() {
+    return 20;
```
juanmrad: Where are we bringing this cost from? 😅

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.test.ts
juanmrad, line 75: we should instead check against error match like:
```
toMatchObject({
  name: 'SignalPermanentError',
  title: expect.stringMatching(/text.*imageUrl/),
});
```

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
```diff
@@ -86,12 +100,66 @@ export async function runOpenAiModerationImpl(
...
 
   const response = await getOpenAiModerationScores({
+    apiKey: credential.apiKey,
     text: value.value,
+  });
+
+  if (response.length === 0) {
+    throw new Error('Empty OpenAI results');
+  }
+
```
juanmrad: we may want to instead use a `SignalPermanentError` over a generic error no?

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
```diff
@@ -86,12 +100,66 @@ export async function runOpenAiModerationImpl(
...
+  const scores = response[0];
```
juanmrad: we should validate this is a number before we do `Number(score)` just to confirm we do not return `NaN` as score

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
```diff
@@ -86,12 +100,66 @@ export async function runOpenAiModerationImpl(
...
+  const score = scores.category_scores[modelName];
+  return {
+    score: Number(score),
+    outputType: { scalarType: ScalarTypes.NUMBER },
+  };
+}
```
juanmrad: something like:
```ts
const score = scores.category_scores[modelName];
if (typeof score !== 'number' || Number.isNaN(score)) {
  throw makeSignalPermanentError(
    `Missing or non-numeric score for category "${modelName}"`,
    { shouldErrorSpan: true },
  );
}
return { score, outputType: { scalarType: ScalarTypes.NUMBER } };
```

juanmrad (commented):
Also agree with @ThisIsMissEm with this being too much boilerplate repeated, we should instead collapse these into a small factory.

Something like:
```
function makeOpenAiImageModerationSignal(config: {
  type: SignalType;
  displayName: string;
  description: string;
  modelName: OpenAiImageModelName;
}) { /* returns a SignalBase subclass */ }
```

Potentially.

server/services/signalsService/signals/third_party_signals/open_ai/moderation/OpenAiViolenceImageSignal.ts
julietshen: ooooh i think this is copied over from the original openai integration - let me clean this up

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
julietshen: you're so right, upgraded to SignalPermanentError in 0ef138f. Both text and image paths now route through a shared extractScoreOrThrow helper that throws SignalPermanentError('Empty OpenAI moderation results', blah blah)

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.test.ts
julietshen: Done in 0ef138f. Now asserts toMatchObject({ name: 'SignalPermanentError', title: expect.stringMatching(/text.*imageUrl/) }) so it pins the error class along with the message

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
julietshen: Done in 0ef138f

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
julietshen: done in 0ef138f

julietshen:
Addressed comments, and created new issues:
https://github.com/roostorg/coop/issues/589 on simplifying signal resolvers

https://github.com/roostorg/coop/issues/590 for making `cost` actually make sense. In this PR i just copied `20` from the existing OAI integration

juanmrad (approved):
👍🏻

server/services/signalsService/signals/third_party_signals/open_ai/moderation/openAIModerationUtils.ts
```diff
@@ -76,6 +91,33 @@ export function openAiModerationNeedsMatchingValues() {
...
+ * rather than spinning on retries.
+ */
+function extractScoreOrThrow(
+  response: ReadonlyDeep<OpenAiModerationResult[]>,
+  modelName: OpenAiModelName,
+): number {
+  if (response.length === 0) {
+    throw makeSignalPermanentError('Empty OpenAI moderation results', {
+      shouldErrorSpan: true,
+    });
```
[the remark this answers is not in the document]
julietshen: given our discussion above, going to ignore this suggestion cc @juanmrad

server/services/signalsService/signals/third_party_signals/open_ai/moderation/OpenAiSelfHarmImageSignal.ts
```diff
@@ -0,0 +1,19 @@
...
+ * disorders, etc.). Routes through omni-moderation-latest's multimodal
+ * endpoint and returns the `self-harm` category score (0..1).
+ */
+const OpenAiSelfHarmImageSignal = makeOpenAiImageModerationSignal({
+  type: SignalType.OPEN_AI_SELF_HARM_IMAGE_MODEL,
+  displayName: 'OpenAI Self Harm Image score',
+  description: `OpenAI's model that detects self-harm, which is defined as content that promotes, encourages, or depicts acts of self-harm, such as suicide, cutting, and eating disorders. Scored against the image.
+
+      This model produces a confidence score between 0 and 1, indicating the model's confidence that the image depicts self-harm.`,
+  modelName: 'self-harm',
```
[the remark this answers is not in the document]
julietshen: done

PR #560, opened 2026-05-22
[Permissions] Capability-based authz with DB-backed roles + static fallback

juanmrad:
Continuation for #406

## Context & Requests for Reviewers

Three pieces shipped together because they all touch the same authz surface:

1. **Capability checks everywhere.** All resolver gates use `user.getPermissions().includes(UserPermission.X)`. Zero role-string branches remain. Supersedes #547 
2. **DB-backed roles.** New `public.roles` / `public.role_permissions` tables, JOINed at user-load via correlated subquery. Migration seeds existing orgs and backfills `role_id` on `users` / `invite_user_tokens`.
3. **Static fallback.** When the JOIN is empty (orgs created after the migration), `getPermissions()` falls back to `UserPermissionsForRole` so fresh orgs work out of the box. The role-editor UI will persist edits in a follow-up PR, switching those orgs to the DB-backed path.

## Summary by CodeRabbit

* **New Features**
  * Add per-organization RBAC with roles and permissions; system roles auto-seeded and existing users/invites backfilled.

* **Bug Fixes**
  * Standardize permission checks to use shared permission constants; tighten authorization for user and org actions.

* **Chores**
  * DB schema and type definitions extended to support roles and role_permissions.

* **Other**
  * New permissions added: MANAGE_USERS, MANAGE_ROUTING_RULES; routing-rule mutations now require MANAGE_ROUTING_RULES.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/560?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #582, opened 2026-05-26
[Permissions] Add role editor API and UI

juanmrad:

Surfaces the capabilities #560 made DB-backed to admins via a role editor:

New GraphQL API endpoints to handle permission groups and modifying permissions as well as surfacing the permission available to those roles.

Materializes the `public.roles` row on first save, atomically replaces `role_permissions` in a transaction, and backfills `role_id` on the org's `users` / `invite_user_tokens` so the JOIN-based permission lookup picks up the new permissions on next load. Falls back to `SystemRoleDefaults` + `UserPermissionsForRole` for orgs that haven't customized.
 
New UI Tab on manage users that is capability gated. Adds UI components from figma to display user editing capabilities and dialogs.

Custom roles will be on a separate PR and work to another Issue, but this completes the work. 

Screenshot 2026-05-25 at 9 30 14 PM
Screenshot 2026-05-25 at 9 30 06 PM
Screenshot 2026-05-25 at 9 29 10 PM
Screenshot 2026-05-25 at 9 29 01 PM
Screenshot 2026-05-25 at 9 28 52 PM

### DIFF
--- a/server/graphql/modules/roles.ts
+++ b/server/graphql/modules/roles.ts
@@ -1,0 +1,171 @@
+import {
+  UserPermission,
+  UserRole,
+} from '../../services/userManagementService/index.js';
+import {
+  type GQLMutationResolvers,
+  type GQLQueryResolvers,
+} from '../generated.js';
+import { type Context } from '../resolvers.js';
+import {
+  forbiddenError,
+  unauthenticatedError,
+  userInputError,
+} from '../utils/errors.js';
+
+const typeDefs = /* GraphQL */ `
+  type Role {
+    "Persisted public.roles.id, or null when the row is materialized lazily on first save."
+    id: ID
+    "Stable role identifier (matches UserRole)."
+    key: UserRole!
+    displayName: String!
+    description: String
+    isSystem: Boolean!
+    permissions: [UserPermission!]!
+    "True when permissions/metadata come from the static fallback rather than public.roles."
+    isFallback: Boolean!
+    "Number of approved (non-rejected) users in the org assigned to this role."
+    userCount: Int!
+  }
+
+  type PermissionGroupItem {
+    permission: UserPermission!
+    label: String!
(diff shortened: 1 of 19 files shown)


julietshen:
I've been testing locally and had some trouble with the resolver. 

`server/graphql/schema.ts` imports `typeDefs` from `./modules/roles.js` but doesn't import its resolvers, so the `rolesForOrg` resolver is never registered, and graphql-js falls back to the default resolver which returns undefined, causing the non-null violation.

server/graphql/modules/roles.ts
```diff
@@ -0,0 +1,171 @@
...
+    throw userInputError('displayName must be at most 255 characters');
+  }
+}
+
+const resolvers = {
+  Query,
+  Mutation,
+};
+
+export { typeDefs, resolvers };
```
julietshen, line 171: as mentioned in the discussion, this module exports resolvers, but nothing imports it... `server/graphql/resolvers.ts` only registers the typeDefs (via schema.ts) and never merges these resolvers into the resolver map. As a result `Query.rolesForOrg` falls back to the default resolver and returns undefined, which graphql-js converts to Cannot return null for non-nullable field `Query.rolesForOrg` on every load of the Roles tab. Needs import { resolvers as rolesResolvers } from './modules/roles.js'; plus a rolesResolvers, entry in the mergeResolvers call in resolvers.ts.

julietshen (changes requested):

julietshen (approved):
added the resolver fix

julietshen:
approved - tested UI functionality locally and confirmed roles page shows up, i can manage and rename roles and edit the permissions for each role

Screenshot 2026-05-27 at 11 57 21 AM
Screenshot 2026-05-27 at 11 57 46 AM
Screenshot 2026-05-27 at 11 58 07 AM

PR #602, opened 2026-05-27
Rename published package to @roostorg/coop-types

### LINKED ISSUE #598: Rename package published for coop integration types
Currently we're publishing to npm as [`@roostorg/types`](https://npmx.dev/@roostorg/types), which isn't specific to coop, even though these types are specific to coop and it's ecosystem of integrations. Longer term, from within the codebase, this would be a local package via pnpm workspaces, the publication surface is for external integrators to consume.

As such, a name like `@roostorg/coop-integration-types` probably makes more sense.

As we use OIDC Trusted Publishing in this repository, the first publish of `@roostorg/coop-integration-types` would need to be done from a developer's machine (a minimal stub package @ version 0.0.0, no actual code), and then we'd do the release workflow to actually publish the package for real. This creates the package which allows us to configure the OIDC Trusted Publishing + any related workflow logic we want.

(We could also publish pre-releases from time-to-time, and/or publish on merge to main [as a separate tag](https://docs.npmjs.com/cli/v8/commands/npm-publish?v=true#tag), i.e., not `latest` which is the default and what you get when you do `pnpm install @roostorg/coop-integration-types`)

juanmrad:
Fixes #598 

Renames the types package in coop to a more appropriate `coop-types` already published a 2.1.0 so here we are bumping to 2.2.0 to confirm work of the release via CI. 

There will be a follow up pr after this to move to use 2.2.0 on codebase.

## Summary by CodeRabbit

* **Chores**
  * TypeScript types package renamed from `@roostorg/types` to `@roostorg/coop-types`.
  * Package version bumped to 2.2.0.
  * Publish process updated to derive the package name dynamically and skip publishing when the specific package@version already exists.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/602?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/publish-types.yaml
+++ b/.github/workflows/publish-types.yaml
@@ -28,13 +28,14 @@ jobs:
       - name: Check and publish types
         run: |
           cd types
+          PKG_NAME=$(node -p "require('./package.json').name")
           VERSION=$(node -p "require('./package.json').version")
-          echo "Checking @roostorg/types@$VERSION..."
-          
-          if npm view "@roostorg/types@$VERSION" version >/dev/null 2>&1; then
-            echo "⚠️  Version $VERSION of @roostorg/types already exists on npm - skipping"
+          echo "Checking $PKG_NAME@$VERSION..."
+
+          if npm view "$PKG_NAME@$VERSION" version >/dev/null 2>&1; then
+            echo "⚠️  Version $VERSION of $PKG_NAME already exists on npm - skipping"
           else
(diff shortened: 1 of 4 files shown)


julietshen (approved):
assuming coderabbit/copilot comments will be addressed, this LGTM based on our discussion in the coop working group. Let's make sure Kyodo and Notion are aware

ThisIsMissEm:
LGTM (fwiw, the release tooling will be able to simplify this a bunch too)

PR #606, opened 2026-05-27
[Coop-Types] Migrate to newly created package

juanmrad:

s/types/coop-types 

Replace all uses of types to coop types.


julietshen (approved):
thanks :)

PR #634, opened 2026-05-28
feat: add nginx serve stage to client Dockerfile

mbradley:
## Summary
- Adds `nginx:1.27.5-alpine` serve stage to the client Dockerfile for production SPA serving
- `nginx.conf` with SPA routing (`try_files`), security headers, and correct caching strategy
- Vite hashed assets (`/assets/`) get aggressive 1-year cache; everything else gets no-cache so deploys take effect immediately
- `.dockerignore` updated to allow `nginx.conf` through

## Context
The client Dockerfile only had a build stage with no production serving. The K8s `coop-client` deployment serves the built React SPA via nginx on port 80, with liveness/readiness probes on `GET /`.

## Test plan
- [ ] CI builds the client image successfully
- [ ] nginx serves the SPA on port 80
- [ ] `GET /` returns 200 (K8s probes will pass)
- [ ] Client-side routing works (deep links serve index.html)
- [ ] Static assets are cached, HTML is not

### DIFF
--- a/client/.dockerignore
+++ b/client/.dockerignore
@@ -9,3 +9,4 @@
 !postcss.config.js
 !tailwind.config.js
 !tsconfig.json
+!nginx.conf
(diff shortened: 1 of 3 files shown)


mbradley:
Opened against wrong remote -- moving to divinevideo/coop

Issue #288, opened 2026-04-21
Design (expanded/desired) integration tests


cassidyjames:
Related to #213 and as mentioned in https://github.com/roostorg/coop/discussions/277#discussioncomment-16657710, we need to determine what integration tests should cover and do some high-level design before implementation.

- [ ] @julietshen to think through other scenarios from a T&S perspective.
- [x] @cassidyjames to create sub-issues for each of the above

cassidyjames:
@vinaysrao1 shared the following:

> 1. End-to-end item submission: send an item through the queue and check it lands correctly in Scylla, ClickHouse.            
> 2. Rule changes take effect: when a rule is created or updated, confirm the cache clears and the next item is judged by the new rule.          
> 3. Report flow works across all systems: submit a report and verify it flows through Postgres, Scylla, the review queue, ClickHouse, and a “simulated” NCMEC service.  
> 4. Background jobs do the right thing: run the NCMEC retry, MRT refresh, and rule-anomaly jobs against real databases and check they don't double-fire or skip rows.                                  
> 5. System survives outages: simulate Redis, Scylla, Postgres, and HMA going down or slow; confirm Coop degrades gracefully without losing data.                                                          
> 6. Speed and throughput stay healthy: load-test the main endpoints to make sure response times and queue depth stay within agreed limits.

PR #640, opened 2026-05-29
Add #341 report-flow integration test

### LINKED ISSUE #341: Report flow integration test
Report flow works across all systems: submit a report and verify it flows through Postgres, Scylla, the review queue, ClickHouse, and a “simulated” NCMEC service.

julietshen:
## Summary

Closes #341 (part of the integration-test design in #288).

Submits a report via `POST /api/v1/report` against a real running stack and asserts the report flows through every system named in the issue:

- **Scylla** (`item_submission_by_thread`) — reported item submission
- **ClickHouse** (`REPORTING_SERVICE.REPORTS`) — analytics row written by `ReportingService.submitReport`
- **Postgres** (`manual_review_tool.job_creations`) — review queue membership
- **\"Simulated\" NCMEC**: when `reportedForReason.csam = true`, the report routes through `NcmecService.enqueueForHumanReviewIfApplicable`, which extracts the content's creator and enqueues the **user** (not the content). The `job_creations.item_id` flipping from content → creator is the observable difference from the standard path. The actual cybertip HTTP submission is downstream of a reviewer decision and intentionally out of scope.

Each test provisions its own content item type so item-type cache pollution from a prior test can't mask resolution problems.

Two new polling helpers in `server/test/integ/wait.ts`:
- `waitForReportInClickHouse` — polls `REPORTING_SERVICE.REPORTS`
- `waitForJobCreationInPostgres` — polls `manual_review_tool.job_creations` (the MRT row is the differentiator for the NCMEC vs. standard path — same table, different `item_id`)

The NCMEC scenario uses `ScalarTypes.IMAGE`. `isMediaType` returns true for IMAGE/AUDIO/VIDEO/MEDIA, and `NcmecEnqueueToMrt` uses the same generic `scalarGetValues` handler for all four, so any media-class scalar exercises the same code path. IMAGE keeps the test runnable today; once #632 (MEDIA wiring) lands, MEDIA would work the same way.

## Dependency note

Running the integ suite locally on today's `main` currently fails type-check at `services/itemProcessingService/fieldTypeHandlers.ts:83` because the `Handlers` type requires a MEDIA entry that #632 has not yet merged. This affects all integ tests, not just this one. Once #632 merges, this test should run cleanly via `npm run test:integ`.

## Test plan

- [x] `npm run up && npm run db:update` then `cd server && npm run test:integ -- report-flow.integ.test.ts` and confirm both scenarios pass
- [x] Verify the standard scenario lands a `job_creations` row with `item_id` = content item
- [x] Verify the CSAM scenario lands a `job_creations` row with `item_id` = content creator (the NCMEC enqueue path)

### DIFF
--- a/server/test/integ/report-flow.integ.test.ts
+++ b/server/test/integ/report-flow.integ.test.ts
@@ -1,0 +1,295 @@
+/**
+ * Integration test for #341: report flow.
+ *
+ * Submits a report via POST /api/v1/report against a real running stack and
+ * asserts it flows through every system named in the issue:
+ *
+ *   - Scylla (`item_submission_by_thread`) — reported item submission
+ *   - ClickHouse (`REPORTING_SERVICE.REPORTS`) — analytics row written by
+ *     `ReportingService.submitReport`
+ *   - Postgres (`manual_review_tool.job_creations`) — review queue membership
+ *   - "Simulated" NCMEC: when `reportedForReason.csam = true`, the report is
+ *     routed through `NcmecService.enqueueForHumanReviewIfApplicable`, which
+ *     extracts the content's creator and enqueues the USER (not the content).
+ *     The `job_creations` row's `item_id` flipping from the content item to
+ *     the creator is the observable difference from the standard path. The
+ *     actual NCMEC HTTP submission is a downstream reviewer-triggered flow
+ *     and is intentionally out of scope.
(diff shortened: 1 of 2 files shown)


calebmcquaid (approved):
Tested and verified. Looks good to me!

PR #641, opened 2026-05-29
Populate NCMEC reports with the `ipAddress` schema field role

### LINKED ISSUE #468: Add IP address as a role when creating Items
Per feedback from an adopter, adding IP address in the schema field for any items you create should ideally be a Role.

### LINKED ISSUE #592: Add IP address to NCMEC reports
Mentioned by @ThatKoffe in [#468](https://github.com/roostorg/coop/issues/468#issuecomment-4545645429): if we have an IP address role, it should automatically be added to related NCMEC reports.

juanmrad:
Fixes #468 
Fixes #592 

Adopters can already tag User and Content items with the new `ipAddress` schema field role from previous prs. This PR closes the loop by sending those IPs to NCMEC.

When `buildSubmitReportParamsFromDecision` assembles a CyberTip payload it now reads the `ipAddress` role off the reported user and each reported media item. `submitReport` appends those IPs to `<ipCaptureEvent>` on the user (`Unknown` event at the report's incident time) and on each file (`Upload` at the media's `createdAt`). 

The additional-info webhook stays the authoritative source. Its events come first, and the role IP is appended alongside, so adopters who haven't yet wired the webhook still get the reported IP into the submission. 

If neither source has data, the element is omitted entirely as usual.

## Summary by CodeRabbit

* **New Features**
  * IP address field mapping and synthesized IP-capture events are now included in NCMEC incident reports for reported users and media uploads.

* **Bug Fixes**
  * Media item timestamps now reliably fall back to the report time when an invalid media date is encountered, improving report accuracy.

* **Documentation**
  * NCMEC docs clarified to explain how mapped IP roles become specific IP-capture events (Unknown for users, Upload for media).

* **Tests**
  * Added tests validating IP propagation, event merging, and createdAt fallback behavior.

### DIFF
--- a/server/services/ncmecService/buildSubmitReportParamsFromDecision.ts
+++ b/server/services/ncmecService/buildSubmitReportParamsFromDecision.ts
@@ -104,6 +104,12 @@ export async function buildSubmitReportParamsFromDecision(
     'profileIcon',
     reportedUserData,
   );
+  const reportedUserIp = getFieldValueForRole(
+    reportedUserItemType.schema,
+    reportedUserItemType.schemaFieldRoles,
+    'ipAddress',
+    reportedUserData,
+  );
 
   // Pre-index allMediaItems by (itemId, typeId) so the per-decisionComponent
   // lookup below is O(1) instead of O(n) for every reportedMedia entry. The
@@ -147,6 +153,12 @@ export async function buildSubmitReportParamsFromDecision(
       if (createdAt === undefined) {
         throw new Error('No created at for reported media');
       }
+      const mediaIp = getFieldValueForRole(
+        mediaItemType.schema,
+        mediaItemType.schemaFieldRoles,
+        'ipAddress',
+        reportedItem.contentItem.data,
+      );
       return {
         id: it.id,
         typeId: it.typeId,
@@ -154,6 +166,7 @@ export async function buildSubmitReportParamsFromDecision(
         createdAt,
         industryClassification: it.industryClassification,
         fileAnnotations: it.fileAnnotations,
+        ...(mediaIp ? { ipAddress: mediaIp } : {}),
       };
     }),
   );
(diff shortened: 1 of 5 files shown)


server/services/ncmecService/ncmecReporting.ts
```diff
@@ -498,6 +507,40 @@ export function clampIncidentDateTimeToPast(
...
+        ? paramEvents
+        : [paramEvents];
+  const trimmedRoleIp =
+    typeof roleIpAddress === 'string' ? roleIpAddress.trim() : '';
+  const events: IPNCMECEvent[] = [
+    ...(webhookEvents ?? []),
+    ...paramEventsArray,
+  ];
+  if (trimmedRoleIp !== '') {
+    events.push({
```
julietshen: this might fail linting

The repo enforces functional/immutable-data.  I hit this rule on PR #643 and had to rewrite rows.push(...) to spread instead. Even though events is a fresh local array, the rule typically fires on the call site. Maybe:

```
  const events: IPNCMECEvent[] = [
    ...(webhookEvents ?? []),
    ...paramEventsArray,
    ...(trimmedRoleIp !== ''
      ? [{ ipAddress: trimmedRoleIp, eventName: synthesisedEvent.eventName, dateTime: synthesisedEvent.dateTime }]
      : []),
  ];
  return events.length > 0 ? events : undefined;
```

server/services/ncmecService/buildSubmitReportParamsFromDecision.ts
```diff
@@ -147,13 +153,20 @@ export async function buildSubmitReportParamsFromDecision(
       if (createdAt === undefined) {
```
julietshen, line 157: this is separate from your change but should we do format validation here? could add a `Date.parse(createdAt)` sanity check in a follow-up at some point

julietshen (approved):
looks fine, left one comment about lint likely to fail

PR #542, opened 2026-05-21
server: Survive Postgres idle-client errors without crashing the process

juanmrad:
Fixes #538. 

## Context & Requests for Reviewers

When Postgres restarts (provider maintenance, brief network blip), `pg.Pool` emits an `'error'` event for each dropped idle client. We had no `'error'` listener, so Node escalated to `uncaughtException` and `bin/www.ts` called `process.exit(1)`. `pg` already handles reconnect lazily on the next checkout the only thing missing was a log-only listener so the process stays alive long enough to recover.

This also centralizes kysely pg configs so the same pool configuration is used on the master pool. and added additional knobs to tune on env example.

## Summary by CodeRabbit

* **Documentation**
  * Added environment variable docs for DB pooling, TCP keepalive, statement/query timeouts, and pool lifecycle.

* **Refactor**
  * Centralized database pool creation and updated session store to use the shared pool for improved reliability and tuning.

* **Tests**
  * Added tests for pool error handling, config forwarding, and proper cleanup.

* **Bug Fixes**
  * Idle DB client errors are now logged and contained to prevent process crashes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/542?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #343, opened 2026-04-28
Outage integration test


cassidyjames:
System survives outages: simulate Redis, Scylla, Postgres, and HMA going down or slow; confirm Coop degrades gracefully without losing data.

julietshen:
## Plan

Splitting #343 into phased PRs because the four services need different machinery and some scenarios will fail-as-designed (surfacing real gaps that need a code decision, not just a test).

### Obstacles called out up front

1. **Service control from inside the test.** The test process needs to stop/start each docker-compose service. Using the host's docker socket: `docker compose stop <svc>` for "down," `docker compose pause <svc>` for "slow" (the container stops responding to TCP — close enough to a slow/hung dependency without standing up toxiproxy). The harness will need a small wrapper helper around these commands with timeout + automatic-restore-on-test-failure semantics.

2. **Some scenarios will fail as designed and surface real gaps.** Two known cases:
   - `ItemProcessingWorker` currently **swallows Scylla write errors** (`// swallow error for now if an item fails to make it into scylla`). A "Scylla outage doesn't lose data" assertion fails immediately — the worker silently drops the row. Decision needed: fail the test (document the gap and file a fix) or skip with a clear TODO.
   - `POST /api/v1/items/async` enqueues via BullMQ; BullMQ's behaviour when Redis is unreachable needs verifying empirically before I can write the assertion. If the API returns 202 anyway, that's a data-loss case worth flagging.

3. **Test-harness survival.** If Postgres goes down mid-test, `afterAll` cleanup hits the dead pool and fails. Pattern: an `afterEach` that brings the paused service back up *before* fixture cleanup runs. Manageable per scenario but worth being explicit about.

4. **"Without losing data" needs a concrete per-service contract.** For each scenario the assertion needs to know whether the system should:
   - Reject loudly (e.g. 503 with no side effects)
   - Enqueue for retry (durable state survives the outage and is replayed on recovery)
   - Accept-and-drop (data-loss case — assertion should fail and the code should change)

   I'll surface my read of each contract in the relevant PR description and the maintainer can correct it on review.

### Phasing

**PR 1 — HMA + Redis (this PR)**
- **HMA outage during report submission.** `submitReport`'s hash-lookup block is wrapped in a log-only `try/catch`. Test: pause HMA, send a report with images, assert the report still succeeds (201), no hashes attached to `data.images`. Bring HMA back, repeat, assert hashes present. This is a clean graceful-degradation check.
[...]

PR #646, opened 2026-05-29
Add outage integration test: Redis (phase 1 of #343)

julietshen:
First phase of #343 per the [planning comment](https://github.com/roostorg/coop/issues/343#issuecomment-4577676451) on that issue.

## What's in

- `server/test/integ/dockerCompose.ts` — thin wrapper around `docker compose stop/start/pause/unpause` plus `withServiceDown` / `withServicePaused` helpers that auto-restore in a `finally` so an assertion failure inside the callback can't leak a stopped/paused service into the next scenario (which would cascade-fail the rest of the suite).
- `server/test/integ/outage.integ.test.ts` — 4 scenarios, all green:
  1.  **Redis stopped** during `/api/v1/items/async` → 202 immediately (ioredis offline-queue path), item lands in Scylla after Redis is restored. **No data loss as long as the API process stays up.**
  4. **Redis paused** — same shape.

## Findings worth a follow-up (not in this PR)

- The Redis path returns **202 while Redis is unreachable**. Today's offline-queue buffering keeps the data durable until the buffer flushes on reconnect, but the buffer lives in the API process — a simultaneous Redis outage + API restart drops the submission. Worth either a pre-flight Redis health check before claiming 202, or `enableOfflineQueue: false` on the enqueue connection. I'll file this as a separate issue.

- How to handle HMA tests, as not all adopters will be using HMA or have configured it.

## What's deferred

Per the plan on #343, this PR covers Redis only. Scylla, Postgres, and explicit "slow" variants ship in follow-up PRs:
- **Phase 2 (Scylla)** ships either a "loud-error" code change or a documented swallow-error contract — design question, not just a test.
- **Phase 3 (Postgres)** builds on #542 for the active-request case.
- **Phase 4 (toxiproxy)** only if `docker pause` doesn't faithfully cover a needed "slow" scenario; current scenarios don't need it.

## Test plan

- [x] `(cd server && npm run test:integ -- outage.integ.test.ts)` against the real stack (`npm run up && npm run db:update`) — 4 passed, 4 total
- [ ] CI runs the full integ suite — confirm the docker-control primitives work in that environment (test process must have access to the host's docker socket; verify the CI runner config exposes it).

## AGENTS.md callouts

- The new tests use docker-compose service control via the host socket. CI may need the docker socket mounted into the test container or the test runner moved out-of-container. Flagging here so reviewers can confirm the CI runner exposes docker.
- `--runInBand` is required because docker-compose state is shared global mutable state. The existing `npm run test:integ` script does not pass `--runInBand`; if CI runs scenarios concurrently this PR's tests will stomp each other. Worth a short follow-up to the jest config to enforce serial execution for the integ suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience during temporary service outages, so the system returns promptly or continues to behave correctly when backend components (including Redis) are unavailable.
  * Added verification that item submissions do not persist when Redis is stopped or paused.

* **Tests**
  * Expanded integration coverage for Redis outage scenarios (stopped and paused), including robust restoration/cleanup to keep test runs reliable.
  * Refined Scylla polling logic used by integration tests to improve correctness during async assertions.

### DIFF
--- a/server/test/integ/dockerCompose.ts
+++ b/server/test/integ/dockerCompose.ts
@@ -1,0 +1,110 @@
+/**
+ * Docker-compose service-control primitives for outage integration tests
+ * (#343).
+ *
+ * The integ harness runs out-of-container alongside the docker-compose stack
+ * from `npm run up`, so we can `docker compose stop/start/pause/unpause` the
+ * individual services to simulate outages and slow dependencies.
+ *
+ * - `stop` / `start`: full container restart. TCP is rejected immediately
+ *   while stopped. Use this for "dependency is unreachable" scenarios.
+ * - `pause` / `unpause`: SIGSTOP / SIGCONT the container. TCP connections
+ *   hang instead of being rejected. Use this for "dependency is slow /
+ *   hung" scenarios, since it's closer to a network timeout than `stop`.
+ *
+ * Both wrappers (`withServiceDown`, `withServicePaused`) restore the
+ * service in a `finally` block so a test assertion failing inside the
+ * callback can't leak a stopped/paused service into a subsequent test
+ * — which would cascade-fail the rest of the suite.
+ *
+ * Tests using this helper MUST run with `--runInBand` (single worker)
+ * since the docker stack is shared state.
+ */
+import { exec } from 'node:child_process';
+import { promisify } from 'node:util';
+
+const execAsync = promisify(exec);
+
+const DEFAULT_COMPOSE_TIMEOUT_MS = 30_000;
+
+async function compose(
+  args: readonly string[],
+  opts: { timeoutMs?: number } = {},
+): Promise<void> {
+  const { timeoutMs = DEFAULT_COMPOSE_TIMEOUT_MS } = opts;
+  // `docker compose` (v2, space-separated) per the project's AGENTS.md.
+  // Shell-quoting the service name keeps multi-word names safe even though
+  // the compose services in this repo don't use any.
+  await execAsync(
+    `docker compose ${args.map((a) => `'${a.replace(/'/g, `'\\''`)}'`).join(' ')}`,
+    { timeout: timeoutMs },
+  );
+}
+
+export async function stopService(svc: string): Promise<void> {
+  await compose(['stop', svc]);
+}
+
+export async function startService(svc: string): Promise<void> {
+  // `start` (rather than `up`) reuses the existing container instead of
+  // recreating it — keeps volumes and data intact across the stop/start
+  // cycle so the next test sees the same DB state.
+  await compose(['start', svc]);
+}
+
+export async function pauseService(svc: string): Promise<void> {
+  await compose(['pause', svc]);
+}
+
+export async function unpauseService(svc: string): Promise<void> {
+  await compose(['unpause', svc]);
+}
+
+/**
+ * Run `fn` with `svc` stopped, then start it again no matter how `fn`
+ * exited. Errors from the restart are surfaced as an `AggregateError`
+ * alongside the original test error so neither failure mode is hidden.
+ */
+export async function withServiceDown<T>(
+  svc: string,
+  fn: () => Promise<T>,
+): Promise<T> {
+  await stopService(svc);
+  try {
+    return await fn();
+  } finally {
+    try {
+      await startService(svc);
+    } catch (restoreErr) {
+       
+      console.error(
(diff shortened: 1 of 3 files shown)


server/test/integ/outage.integ.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  // ioredis's offline buffer is in-process and goes with the restart.
+  // Worth a follow-up (durable enqueue ack, or pre-flight Redis health
+  // before claiming 202) but doesn't belong in this test PR.
+  // -------------------------------------------------------------------------
+
+  test('item survives a brief Redis outage and lands in Scylla after recovery', async () => {
+    const { res, itemId } = await withServiceDown('redis', async () => {
+      return submitItem();
+    });
+    expect(res.status).toBe(202);
```
juanmrad: I belive this is not the case anymore after #653 merged.

juanmrad (commented):
we may need to re-run this and validate assertions.

server/test/integ/dockerCompose.ts
```diff
@@ -0,0 +1,110 @@
+/**
```
taobojlen, line 1: It feels somewhat iffy to me to have test code manipulate infra like this. The main arguments against it are a) it's unusual and surprising, b) it couples our test suite to a specific Docker compose setup. For example, in GitHub Actions, you'd usually start the services you want via the `services:` block. Or perhaps some engineers prefer to use Podman over Docker.

To propose an alternative: could we instead mock the service in question, or use dependency injection to pass in a version of e.g. the Redis client that returns the error?

server/test/integ/outage.integ.test.ts
```diff
@@ -0,0 +1,332 @@
...
+          },
+        ],
+      },
+    });
+    itemTypeId = itemTypeFixture.itemTypes[0].id;
+    itemTypeCleanup = itemTypeFixture.cleanup;
+  }, 60_000);
+
+  afterAll(async () => {
+    // Belt-and-suspenders: even though `withServiceDown` / `withServicePaused`
```
taobojlen: I'm generally against this kind of fallback, though LLMs love to write them. IMO the code should work as we expect -- fallbacks can obscure issues that you'd like to know about.

server/test/integ/outage.integ.test.ts
```diff
@@ -0,0 +1,332 @@
...
+        reportedAt: new Date().toISOString(),
+        reportedItem: {
+          id: uid(),
+          typeId: itemTypeId,
+          data: { images: [imageUrl] },
+        },
+      });
+  };
+
+  test('report succeeds and images are not hash-wrapped when HMA is stopped', async () => {
```
taobojlen: Does this mean that if HMA is down, then reports get accepted and the hashing step silently fails? (IMO you *often* want loud failures in a case like this).

server/test/integ/outage.integ.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  //     accept-and-drop in the durable sense — ioredis has
+  //     `enableOfflineQueue: true` by default, so the `queue.addBulk`
+  //     command is buffered client-side and the DataLoader-batched write
+  //     resolves cleanly. On the wire from the caller's perspective the
+  //     submission has been accepted.
+  //   - Once Redis is restored, ioredis flushes the offline queue, BullMQ
+  //     enqueues the buffered job, and the inline `ItemProcessingWorker`
+  //     picks it up and writes Scylla as normal.
+  //
+  // Net "no data lost" outcome: holds as long as the API process stays up.
```
taobojlen: This is probably not a good idea. If the server returns a 202, it should actually have persisted the background job so that no data is lost even if the process stops. We should be able to assume that processes will stop and start constantly (e.g. if running in k8s pods) without impacting the system.

IMO, the service that runs background jobs (in our case Redis) is a core part of the system, just like Postgres. If it's down then the service should reject any requests that depend on Redis.

taobojlen (changes requested):
I'm just jumping in with some comments here! (I obviously don't have full context so feel free to correct me).

But broadly, a bit concerned about the direction this is going. Happy to open a PR if you'd like to see a concrete alternative proposal!

julietshen:
Update: I'm going to split HMA and Redis since not all Coop adopters will use HMA. If it's not configured at all, the report should succeed without calling HMA. If it is configured and is not reachable, that should indeed fail loudly

Issue #649, opened 2026-05-29
Phase 2 of #343: Scylla outage scenario (decide swallow-vs-loud first)


julietshen:
Phase 2 of #343 per the [planning comment](https://github.com/roostorg/coop/issues/343#issuecomment-4577676451). Phase 1 (HMA + Redis) ships in #646.

## What this needs

A code decision before the test can be meaningfully written. Today `ItemProcessingWorker` (`server/workers_jobs/ItemProcessingWorker.ts`) wraps the Scylla insert in a try/catch that **silently swallows** the error:

```ts
try {
  await insertWithRetries({ requestId, orgId, itemSubmission });
} catch (e: unknown) {
  // swallow error for now if an item fails to make it into
  // scylla; it shouldn't prevent processing
}
```

That means a Scylla outage during item processing causes silent row loss: the BullMQ job completes "successfully," the worker moves on, the row never lands in `item_submission_by_thread`.

## Two paths

1. **Loud-error**: re-throw on Scylla insert failure so BullMQ moves the job to the retry queue. The worker stays alive but the job is durably retried until Scylla recovers. Test asserts: Scylla stopped → item eventually lands after Scylla restored. Same shape as the Redis test in #646.
2. **Document the swallow**: keep the current behaviour but add an integ test that asserts "Scylla outage during item processing drops the row" so the contract is at least pinned and reviewers know what to expect. Add a TODO with this issue's link.

Path 1 is the right reliability bar but a real code change. Path 2 is cheap and pins the current contract.

## Acceptance (path 1)

- [ ] `ItemProcessingWorker` surfaces Scylla insert errors so BullMQ retries the job.
- [ ] Integ test in `outage.integ.test.ts`: stop Scylla, submit item, restore Scylla, assert item eventually lands in `item_submission_by_thread`.
- [ ] No regression: items still land happily on the happy path; rule evaluation still runs.

## Acceptance (path 2)

- [ ] Integ test in `outage.integ.test.ts` documents the current swallow-on-error behaviour with a clear `// TODO` linking to this issue.
- [ ] A separate decision recorded somewhere (this issue, an ADR, or the test comment) about whether to escalate to path 1 later.

References #343, #646.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

taobojlen:
i would very strongly recommend path 1 here! scylla stores item submissions; that's crucial data that should never be dropped. right now it's like redis/postgres, a piece of critical infrastructure for coop.

Issue #648, opened 2026-05-29
Integ test suite needs --runInBand for outage scenarios


julietshen:
Surfaced by #646 (phase 1 of #343).

## What's missing

The new outage integ tests in `server/test/integ/outage.integ.test.ts` mutate docker-compose state (`docker compose stop / pause / start / unpause`). That's shared global mutable state across the whole jest worker pool — if two outage scenarios run concurrently in different workers, they stomp each other's stop/start cycles.

The existing `npm run test:integ` script does not pass `--runInBand` (see `server/package.json`):

```
"test:integ": "NODE_OPTIONS=\"--loader ts-node/esm\" jest --ci --reporters=default --silent=false --reporters=jest-junit --detectOpenHandles --no-cache --forceExit --config jest.integ.config.cjs"
```

Locally this happens to work because the suite is small and jest-light-runner tends to schedule things sequentially when there's only one test file, but it's not guaranteed.

## Fix

Add `--runInBand` to the `test:integ` script (or set `maxWorkers: 1` in `jest.integ.config.cjs`). One-line change.

Tightening the integ suite's concurrency model isn't strictly an outage-test concern — items-submission and rule-change tests don't conflict — but it's the safe default once any scenario touches global state.

Discovered while building #646. References #343.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

taobojlen:
If at all possible, it'd be good to avoid doing this so that we can run tests in parallel (and in random order). This is especially relevant for integration tests since they can be slow!

taobojlen:
@julietshen can we close this? this would slow down our test suite and i don't think it's necessary to achieve the desired outcome!

julietshen:
sounds good!

PR #652, opened 2026-05-29
Add Scylla outage integ test (phase 2 of #343)

julietshen:
Phase 2 of #343. Stacks on #646 (phase 1: HMA + Redis) for the `dockerCompose.ts` harness.

## Summary

Pins today's known-lossy contract for #649: `ItemProcessingWorker` swallows Scylla insert errors, so a Scylla outage during processing drops the `item_submission_by_thread` row without retry. The job completes "successfully" from BullMQ's POV, the worker logs to ClickHouse, and the API has long since returned 202 to the client. Nothing surfaces the loss.

## Test design

1. Stop Scylla via `withServiceDown` (from phase 1).
2. POST item via `/api/v1/items/async` — expect 202.
3. Wait for `CONTENT_API_REQUESTS` row in ClickHouse — its presence proves the worker reached past `insertWithRetries`. If the worker were re-throwing on insert failure (the desired contract), this gate would time out instead.

## What's deliberately *not* asserted

"Row absent in Scylla after restore." The in-harness Cassandra driver doesn't recover cleanly within a reasonable window after a Scylla container restart — stale connections, prepared-statement re-prepare, etc. A direct `Scylla.select` after `withServiceDown` ends up testing the driver's reconnect timing rather than the worker's behaviour. The CH-row assertion alone is sufficient proof of the contract: that row can only get written if the worker reached past the swallowed insert.

When #649 flips the worker to re-throw-and-retry, the assertion shape becomes "row lands in Scylla after recovery" (same shape as the Redis-recovery test in `outage.integ.test.ts`) — and at that point the driver-recovery timing problem either dissolves into a presence poll (which tolerates transient errors as "not yet") or gets solved alongside.

## File placement

Lives in a separate file (`outage-scylla.integ.test.ts`) rather than appending to `outage.integ.test.ts` from #646 because #646 is still under review and merging this in the same file would require constant rebases. Once both land we can collapse the two files together — happy to do that follow-up.

## Test plan

- [x] `(cd server && npm run test:integ -- outage-scylla.integ.test.ts)` against the real stack (`npm run up && npm run db:update`) — 1 passed, 1 total
- [ ] Reviewer: confirm "CH-row landing as sufficient proof of swallow" framing is the right contract pin for #649. Alternative is to escalate to #649 path 1 (worker code change) in this same PR — happy to do that if you'd rather.

## AGENTS.md callouts

- Stacks on PR #646. If #646 changes base, this PR's base needs to follow.
- Same `--runInBand` requirement as phase 1 (see #648).

## Summary by CodeRabbit

* **Tests**
  * Added integration tests to validate system resilience during service outages, ensuring the API continues accepting requests and background workers persist with processing despite temporary service unavailability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/652?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/test/integ/outage-scylla.integ.test.ts
+++ b/server/test/integ/outage-scylla.integ.test.ts
@@ -1,0 +1,215 @@
+/**
+ * Integration test for #343 — phase 2: Scylla outage scenario.
+ *
+ * Lives in its own file rather than appending to `outage.integ.test.ts`
+ * because that file ships in #646 (still under review at the time of
+ * writing) and a parallel-branch PR would conflict. Once both land we
+ * can collapse the two files together.
+ *
+ * Pins today's known-lossy contract for #649:
+ *
+ *   `ItemProcessingWorker` wraps the Scylla insert in a try/catch that
+ *   silently swallows write errors (`// swallow error for now if an
+ *   item fails to make it into scylla`). So a Scylla outage during item
+ *   processing means the BullMQ job completes "successfully," the
+ *   worker moves on to log to ClickHouse, and the row never lands in
+ *   `item_submission_by_thread`.
+ *
+ * The test asserts:
+ *
+ *   1. The API still claims 202 (we never told the client there was a
+ *      problem).
+ *   2. The worker still reaches `CONTENT_API_REQUESTS` in ClickHouse
+ *      (the swallow let it continue past the insert).
+ *
+ * We deliberately don't assert "row absent in Scylla" after restore.
+ * The in-harness Cassandra driver doesn't recover cleanly within a
+ * reasonable window after a Scylla container restart (stale
+ * connections, prepared-statement re-prepare); a direct query there
+ * ends up testing the driver's reconnect timing rather than the
+ * worker's behaviour. The CH row landing in assertion 2 is sufficient
+ * proof of the contract — it can only get written if the worker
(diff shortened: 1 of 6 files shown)


server/test/integ/outage-scylla.integ.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    itemTypeId = itemTypeFixture.itemTypes[0].id;
+    itemTypeCleanup = itemTypeFixture.cleanup;
+  }, 60_000);
+
+  afterAll(async () => {
+    // Belt-and-suspenders restore: even though `withServiceDown` restores
+    // Scylla on the inner-fn exit, a harness crash could skip its finally
+    // block. Restoring here keeps the rest of the integ suite running.
+    try {
+      await startService('scylla');
```
juanmrad, line 115: `startService` returns when the container process starts, not when Scylla is CQL-ready (~10–30s). The suite runs `--runInBand` in one process, and the next test file's `scylla.connect()` fails fast if Scylla is unreachable. 

This restart could race a fresh harness and flake a later file. Can we add a readiness poll after restart, or confirm the suite tolerates it?

juanmrad (approved):
one comment, but not a huge deal. LGTM.

PR #653, opened 2026-05-29
items/async: fail fast on Redis outage (closes #647)

### LINKED ISSUE #647: items/async returns 202 while Redis is unreachable (narrow data-loss on API restart)
Surfaced by the integ tests in #646 (phase 1 of #343).

## What today's contract looks like

`POST /api/v1/items/async` returns **202 immediately** while Redis is stopped or paused. The call chain:

1. `submitItems.ts` → `itemSubmissionQueueBulkWrite(...)` → DataLoader batches for 500ms, then calls `queue.addBulk` on BullMQ.
2. ioredis has `enableOfflineQueue: true` by default, so when Redis is unreachable the command is buffered client-side and `addBulk`'s promise resolves immediately on enqueue (not on Redis ACK).
3. The DataLoader batch sees a clean resolve, every item slot returns `undefined`, `bulkWriteResponse.error` is `false`, and the handler falls through to `res.status(202).end()`.

## Why this is mostly fine, but not entirely

Once Redis recovers, ioredis flushes the offline queue, BullMQ delivers to the worker, and the item lands in Scylla as normal. The integ test (`outage.integ.test.ts` in #646) asserts exactly this — items survive a brief Redis outage as long as the API process stays up.

**Gap**: the offline buffer is in-process. A simultaneous Redis outage + API restart (deploy, crash, kubernetes-evict) drops every submission that's sitting in the buffer. The client got a 202; the data is gone.

## Suggested fixes

Either of these closes the gap:

1. **Pre-flight Redis health on the request path**: `PING` (cheap) or check the ioredis connection state before claiming 202. If unhealthy, return 5xx and let the client retry.
2. **`enableOfflineQueue: false` on the enqueue connection**: `queue.addBulk` would reject immediately when Redis is unreachable. The existing 500-path in `submitItems.ts` (`if (bulkWriteResponse.error) { res.status(500).end(); }`) would then fire correctly. May require a separate ioredis client for the enqueue path so it doesn't break unrelated read paths that benefit from buffering.

Option 2 is the smaller diff and matches the handler's apparent intent.

## Acceptance

- [ ] `POST /api/v1/items/async` returns 5xx (not 202) while Redis is stopped or paused.
- [ ] The integ test in `outage.integ.test.ts` flips from "lands after recovery" assertion to "5xx + no Scylla row" assertion.
- [ ] No regression in throughput for the happy path (DataLoader batching still applies).

Discovered while building #646. References #343.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

julietshen:
Closes #647.

## Summary

`POST /api/v1/items/async` used to return **202 even when Redis was unreachable**. ioredis's default `enableOfflineQueue: true` buffers commands client-side, so `queue.addBulk` resolved cleanly while Redis was down and the DataLoader-batched write reported success. The submission survived as long as the API process stayed up and Redis came back — but the buffer is in-process, so a simultaneous Redis outage + API restart silently dropped every submission acknowledged with 202.

## Fix

Three small changes:

1. **`server/iocContainer/index.ts`** — extract a `makeRedis(extraOptions)` helper so the existing IORedis connection wiring isn't copy-pasted, then add a second factory `IORedisEnqueueNoBuffer` that builds the same client with `enableOfflineQueue: false`. Wire `itemSubmissionQueueBulkWrite` to the new client. The shared `IORedis` keeps its default buffer for BullMQ workers (which legitimately need to ride through transient reconnects). Register the new client in `closeSharedResourcesForShutdown`.

2. **`server/routes/items/submitItems.ts`** — tighten the response flow. On `bulkWriteResponse.error`, short-circuit with 5xx and `return`. Before this fix the handler sent 500 *and* then fell through to a second `res.status(202).end()` call. Express swallowed the second one as "headers already sent" so clients got 500 in practice, but the code claimed both intents.

## Verification

Verified locally with a standalone smoke test against the dev `docker compose` stack:

```
initial PING: PONG
stopping redis...
PING-while-down REJECTED: Stream isn't writeable and enableOfflineQueue options is false (GOOD)
starting redis...
PING-after-restore: PONG
```

The DLQ path (`itemSubmissionRetryQueueBulkWrite`) stays on the shared `IORedis` — it's worker-internal and buffering through brief reconnects is fine there.

## What changed in observable behaviour

- Redis up: identical to before (202 on success).
- Redis briefly unreachable (stopped or paused): API now returns **5xx**, telling the client to retry. Previously returned 202 against a buffer.
- Redis intermittent (very brief disconnect during a single `addBulk` call): the new client rejects rather than buffering. Adopters who relied on the implicit buffering to ride through ms-scale blips will see more 5xx, but the trade is "honest 5xx" vs "false 202 + possible loss on restart." This is the intended contract per #647.

## Follow-ups

- Phase 1 of the outage integ suite (#646) currently pins the pre-fix lossy contract for the items/async path. Once both #646 and this PR land, the Redis scenarios in `outage.integ.test.ts` flip from "lands after recovery" to "5xx + no Scylla row." I'll do that follow-up in a rebase of #646 after this merges.

## Test plan

- [x] Standalone smoke test confirms `enableOfflineQueue: false` client rejects when Redis is stopped and recovers cleanly when restarted
- [x] `items-submission.integ.test.ts` (the happy-path integ test) still passes — confirms the new wiring doesn't regress submissions when Redis is healthy
- [ ] Reviewer: confirm the DLQ path staying on the shared `IORedis` is the right call (rationale: worker-internal, transient reconnects acceptable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced item submission error handling: Bulk write operations now correctly return HTTP 500 errors when the backend service is unavailable, preventing silent failures that previously appeared as successful submissions.


server/iocContainer/index.ts
```diff
@@ -337,6 +337,17 @@ export interface Dependencies {
...
+  /**
+   * Dedicated ioredis client for the items-async enqueue path. Same Redis
+   * as `IORedis`, but built with `enableOfflineQueue: false` so a
+   * `queue.addBulk` while Redis is unreachable rejects immediately
+   * instead of buffering in-process. The shared `IORedis` keeps its
+   * default offline-buffer behaviour because BullMQ workers and other
+   * read-paths rely on it for transient reconnects.
+   *
+   * Closes #647.
+   */
```
juanmrad, line 345:
```suggestion
   * Dedicated ioredis client for the items-async enqueue path. Same Redis
   * as `IORedis`, but built with `enableOfflineQueue: false` so a
   * `queue.addBulk` while Redis is unreachable rejects immediately
   * instead of buffering in-process
   */
```

server/iocContainer/index.ts
```diff
@@ -573,7 +584,14 @@ export default async function getBottle() {
       }),
   );
 
-  bottle.factory('IORedis', () =>
+  // Shared ioredis builder used by both the default `IORedis` client
+  // (BullMQ workers, general read/write) and the dedicated no-offline-buffer
+  // client used by the items-async enqueue path. The extra knobs let the
+  // enqueue path turn off the offline command queue without copy-pasting the
+  // connection wiring; see #647 for why the enqueue path needs that.
+  const makeRedis = (
```
juanmrad:
```suggestion
  const makeRedis = (
```

server/iocContainer/index.ts
```diff
@@ -605,7 +624,18 @@ export default async function getBottle() {
...
+          ...extraOptions,
+        });
+
+  bottle.factory('IORedis', () => makeRedis());
+  // The items-async enqueue path uses this client. With
+  // `enableOfflineQueue: false`, a `queue.addBulk` while Redis is
+  // unreachable rejects immediately instead of resolving against the
+  // in-process buffer. submitItems.ts's existing 5xx branch then fires
+  // and the API tells the client "couldn't enqueue" rather than
+  // claiming 202 against a buffer that vanishes on process restart.
```
juanmrad:
```suggestion
  // With `enableOfflineQueue: false`, a `queue.addBulk` while Redis is
  // unreachable rejects immediately instead of resolving against the
  // in-process buffer. fails enqueue early with "couldn't enqueue".
```

server/routes/items/submitItems.ts
```diff
@@ -327,11 +328,20 @@ Dependencies): RequestHandlerWithBodies<SubmitItemsInput, undefined> {
...
         },
       );
 
+      if (enqueueFailed) {
+        // Don't claim 202 if the enqueue rejected — the no-offline-buffer
+        // ioredis client used here (see #647) makes that rejection a real
+        // signal that the submission is *not* durable yet, and we'd rather
+        // tell the client to retry than imply success against a buffer that
+        // vanishes on process restart.
+        res.status(500).end();
```
juanmrad:
```suggestion
        res.status(500).end();
```

juanmrad (approved):
removal of excessive comments 😅. keeping code cleaner. Otherwise LGTM

Issue #611, opened 2026-05-27
License scanning


julietshen:
Coop is Apache-2.0. We should implement automated license checking to ensure we don't introduce dependencies with incompatible licenses.

julietshen:
Did a one-time license-check scan and created https://github.com/roostorg/coop/pull/642

We will likely set up a github action to trigger when changes are being made. cc @cassidyjames (lmk if you have bandwidth to set this up, or i can try to!)

cassidyjames:
Can do!

PR #691, opened 2026-06-03
package.json: Move root dependencies to devDependencies

cassidyjames:
The dependencies are only used at development time, and not deployed in any production package. For example, generated GraphQL files are committed to the repo, and the docker build doesn't re-run codegen but uses the in-repo generated files.

Moving to devDependencies makes it more clear that they're not included in the production build for license scanning (#611), especially for transitive deps (e.g. caniuse-lite is CC-BY-4.0).

## Summary by CodeRabbit

* **Chores**
  * Adjusted dependency configuration to mark a build/dev tool as development-only, clarifying which packages are required at runtime versus only for development and tooling.


package.json
```diff
@@ -25,19 +25,19 @@
   },
   "dependencies": {
     "@graphiql/create-fetcher": "^0.1.0",
```
juanmrad: This is also a dev dependency. Technically all in this file are since we do not deploy this to anywhere and it is used for local dev.

juanmrad (changes requested):
you need to run `npm i` to generate the new lock file with these changes.

PR #708, opened 2026-06-05
[docker] Create docker image for migrations scripts to run

juanmrad:

This pr adds a published `coop-migrations` Docker image so database migrations can be run as a one-shot task on deploy (e.g. an ECS `RunTask`/Kubernetes `Job`) instead of mounting source and installing deps at runtime. The image mirrors the root `npm run db:*` commands so usage matches local development.

Prod invocation is `npm run db:update -- --db <db> --env prod` only — `--env prod` skips the `*.seed.staging.sql` test data, and `create`/`clean`/`drop` are blocked in prod by the migrator.

## Summary by CodeRabbit

* **New Features**
  * Automated publishing of a prebuilt migrations Docker image for database operations; migrations service now uses that image.
  * Added npm scripts for database lifecycle management operations.

* **Bug Fixes**
  * Improved Scylla healthcheck to use a real CQL query so migrations wait for readiness.

* **Documentation**
  * Updated development docs with migration image details, usage examples, supported DB targets, and environment guidance.

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -6,6 +6,7 @@ on:
     paths:
       - 'server/**'
       - 'client/**'
+      - 'db/**'
       - 'Dockerfile'
       - 'client/Dockerfile'
       - '.github/workflows/publish-docker.yaml'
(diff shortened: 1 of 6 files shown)


calebmcquaid (approved):
This makes sense to me. The suggestions are fine but not 100% necessary IMO

PR #710, opened 2026-06-05
[NCMEC] Allow to configure min images to review for NCMEC report

### LINKED ISSUE #517: Improve NCMEC view/flow to handle image gallery
Instead of having a system where we have to review all media on an account, which can be upto 400 or more.. 

It'd be much better to have to select/classify 1 or more media to be able to submit the report. This would improve workflows a lot.

juanmrad:
Fixes #517

Adds an org-level policy for how much media a reviewer must classify before they can send an NCMEC CyberTip.

- **Review all media** (default): existing behavior — every piece of media on the account must get a decision before Send is enabled.
- **Require a minimum number of reviewed media**: Send is enabled after the configured minimum is reviewed, so reviewers aren’t blocked by hundreds of unrelated items on the same account.
In both modes, at least one piece of media must be classified with a reporting category (not `None`) before a report can be sent.

## Summary by CodeRabbit

* **New Features**
  * Admins can configure NCMEC "Media review requirement" (All vs Minimum) and a minimum count; reviewer UI now enables/disables the Send action based on that policy and shows a contextual tooltip. Keyboard shortcut behavior updated to respect the policy.

* **Documentation**
  * User guide updated to explain the two media-review modes and report eligibility.

* **Tests**
  * Added tests covering media-review policy validation and defaults.

* **Chores**
  * Database migration to store the new org settings.

### DIFF
--- a/client/src/webpages/settings/NCMECSettings.tsx
+++ b/client/src/webpages/settings/NCMECSettings.tsx
@@ -203,7 +203,7 @@ export default function NCMECSettings() {
     const isMinimumPolicy =
       settings.mediaReviewRequirement ===
       GQLNcmecMediaReviewRequirement.Minimum;
-    const parsedMinMedia = Number.parseInt(settings.minMediaToReview, 10);
+    const parsedMinMedia = Number(settings.minMediaToReview);
     if (
       isMinimumPolicy &&
       (!Number.isInteger(parsedMinMedia) || parsedMinMedia < 1)
@@ -499,13 +499,18 @@ export default function NCMECSettings() {
                   type="number"
                   min={1}
                   step={1}
+                  inputMode="numeric"
                   value={settings.minMediaToReview}
-                  onChange={(e) =>
-                    setSettings({
-                      ...settings,
-                      minMediaToReview: e.target.value,
-                    })
-                  }
+                  onChange={(e) => {
+                    const value = e.target.value;
+                    // Whole numbers only — fractional thresholds are invalid.
+                    if (value === '' || /^\d+$/.test(value)) {
+                      setSettings({
+                        ...settings,
+                        minMediaToReview: value,
+                      });
+                    }
+                  }}
                   placeholder="1"
                 />
                 <Text size="XS" className="text-gray-500">
--- a/client/src/webpages/settings/NCMECSettings.tsx
+++ b/client/src/webpages/settings/NCMECSettings.tsx
@@ -11,6 +11,7 @@ import {
 import { toast } from '@/coop-ui/Toast';
 import { Heading, Text } from '@/coop-ui/Typography';
 import {
+  GQLNcmecMediaReviewRequirement,
   GQLUserPermission,
   useGQLNcmecOrgSettingsQuery,
   useGQLUpdateNcmecOrgSettingsMutation,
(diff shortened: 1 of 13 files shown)


client/src/webpages/settings/NCMECSettings.tsx
```diff
@@ -416,6 +448,79 @@ export default function NCMECSettings() {
...
+                  onChange={(e) => {
+                    const value = e.target.value;
+                    // Whole numbers only — fractional thresholds are invalid.
+                    if (value === '' || /^\d+$/.test(value)) {
+                      setSettings({
+                        ...settings,
+                        minMediaToReview: value,
+                      });
+                    }
+                  }}
```
calebmcquaid, line 513: This might be simpler but up to you!

```suggestion
                  onChange={(e) => {
                    const parsed = parseInt(e.target.value, 10)
                    // Whole numbers only — fractional thresholds are invalid.
                    if (!Number.isNaN(parsed)) {
                      setSettings({
                        ...settings,
                        minMediaToReview: Math.max(1, parsed),
                      });
                    }
                  }}
```

db/src/scripts/api-server-pg/2026.06.03T04.19.05.add_media_review_requirement_to_ncmec_org_settings.sql
```diff
@@ -0,0 +1,26 @@
...
+-- classify hundreds of items to submit a report.
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD COLUMN IF NOT EXISTS media_review_requirement character varying(16) NOT NULL DEFAULT 'ALL';
+
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD CONSTRAINT ncmec_org_settings_media_review_requirement_check
+  CHECK (media_review_requirement IN ('ALL', 'MINIMUM'));
+
+ALTER TABLE ncmec_reporting.ncmec_org_settings
+  ADD COLUMN IF NOT EXISTS min_media_to_review integer NULL;
```
calebmcquaid, line 14: Should this be 
```suggestion
  ADD COLUMN IF NOT EXISTS min_media_to_review integer DEFAULT 1 NOT NULL;
```

calebmcquaid (approved):
Couple of nit suggestions but looks good!

client/src/webpages/settings/NCMECSettings.tsx
juanmrad: It does look simpler, but this would silently turn a typed 0 into 1 (`Math.max(1, 0)`), since 0 isn't `NaN`. My approach keeps the actual value the user entered and validates it against the guard on save, surfacing an explicit error for 0 instead of silently coercing it. The `/^\d+$/` check also rejects fractional input as you type, rather than truncating 1.5 → 1 the way parseInt would (the same silent-truncation Copilot flagged).

db/src/scripts/api-server-pg/2026.06.03T04.19.05.add_media_review_requirement_to_ncmec_org_settings.sql
juanmrad: I'd prefer to keep this nullable. min_media_to_review is only meaningful when `media_review_requirement = 'MINIMUM'`, so `NULL` is an intentional "not applicable" for ALL orgs rather than a missing value. defaulting it to 1 would store a phantom threshold the app never reads and could mislead anyone querying the table directly.

PR #713, opened 2026-06-05
docs: add deployment infrastructure requirements

### LINKED ISSUE #703: Document base infra requirements/recommendations for deployment
We should explicitly call out requirements (OS, CPU, RAM) for deploying Coop in the deployment guide.

JagadeeshChandra12:
Fixes #703

## Summary

Adds infrastructure requirements and recommendations to the deployment guide.

### Changes

* Added a new "Infrastructure requirements and recommendations" section.
* Documented supported operating system guidance.
* Added memory recommendations based on existing project prerequisites documented in AGENTS.md.
* Added guidance to monitor resource utilization and scale infrastructure as needed.

## Summary by CodeRabbit

* **Documentation**
  * Updated the self-hosting deployment guide with a new “Infrastructure requirements and recommendations” section, including baseline minimum vs. recommended sizing (OS, CPU, RAM).
  * Clarified how the documented development RAM relates to expected production memory usage, and recommended monitoring utilization and scaling accordingly.

### DIFF
--- a/docs/development/deployment.md
+++ b/docs/development/deployment.md
@@ -6,6 +6,19 @@ For historical reference, AWS infrastructure code (CDK, Helm charts, Pulumi, CDK
 
 ## Self-hosting checklist
 
+### Infrastructure requirements and recommendations
+
+The following recommendations are intended as a starting point for self-hosted deployments. Actual requirements will vary depending on workload, moderation volume, retention settings, and enabled integrations.
+
+| Resource         | Minimum                   | Recommended                             |
+| ---------------- | ------------------------- | --------------------------------------- |
+| Operating System | Modern Linux distribution | Latest supported LTS Linux distribution |
+| Memory (RAM)     | 4 GiB                     | 8 GiB                                   |
+
+The documented development prerequisites recommend 8 GiB of RAM, with approximately 4 GiB required to run a Coop instance and additional memory used by supporting services and tooling.
+
+Production deployments should monitor resource utilization and scale infrastructure as needed.
+
 Coop does not currently ship a single production deployment recipe, but the repository does include the configuration surface you need to stand up a self-hosted instance. Treat the example environment files in `db/.env.example`, `server/.env.example`, and `client/.env.example` as the starting point for your deployment-specific configuration.
 
 ### Required production configuration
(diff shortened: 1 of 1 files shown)


docs/development/deployment.md
```diff
@@ -1,28 +1,36 @@
...
 
+### Infrastructure requirements and recommendations
+
+The following recommendations are intended as a starting point for self-hosted deployments. Actual requirements will vary depending on workload, moderation volume, retention settings, and enabled integrations.
+
+| Resource         | Minimum                   | Recommended                             |
+| ---------------- | ------------------------- | --------------------------------------- |
+| Operating System | Modern Linux distribution | Latest supported LTS Linux distribution |
+| CPU              | 2 vCPUs                   | 4+ vCPUs                                |
+| Memory (RAM)     | 4 GiB                     | 8 GiB                                   |
```
taobojlen: Out of curiosity, where did these values come from?

julietshen:
Hi and thank you for your contribution! This looks fine to me, but maybe @calebmcquaid can chime in since he's more familiar with deploying

julietshen (approved):
lgtm pending Caleb's comments

calebmcquaid (changes requested):
I see the language in here around 

> additional memory used by supporting services and tooling.

But I think we should be more explicit around these numbers. 4-8GiB is purely to do a single instance _dev_ deployment while if this were to be used in prod, you are likely looking at upwards of >24GiB when all the services are launched.

JagadeeshChandra12:
Thanks for the clarification! I've updated the documentation to distinguish between development/single-instance deployments and production deployments. The table now explicitly calls out that production deployments running all supporting services will likely require substantially more memory (typically 24+ GiB RAM), while the lower requirements are intended for evaluation and single-instance setups.

docs/development/deployment.md
JagadeeshChandra12: Thanks for asking! The original values were intended as conservative estimates for a basic single-instance deployment rather than production sizing. I've updated the documentation to make that distinction explicit.

reitblatt:
@calebmcquaid I see @JagadeeshChandra12 has updated the documentation as you requested. Can you confirm it looks good, and if so then we should be good to merge.

taobojlen (changes requested):
@JagadeeshChandra12 have you run Coop in production?

your comments and this contribution all appear to be AI-generated, so i can't tell if this is based on your real experience or if you asked an LLM to come up with some numbers.

IMO we should not merge this if it's purely speculative and not based on real production experience.

Issue #409, opened 2026-05-07
Queues should have custom prioritization in the UI


julietshen:
I think right now, Coop's queues are first-in-first-out. Organizations might want their own ranking of tasks in queues and be able to configure this.

julietshen:
This might live in Settings, add a sort to the dashboard page, or change the filter button to be a task sorting

Sorting options should include:
- Oldest first
- # of User Reports
- Custom (open signal library to choose)

<img width="1457" height="667" alt="Image" src="https://github.com/user-attachments/assets/f6d03d51-7c5b-479a-a3e3-62bba7d6bfac" />

julietshen:
Let's split this into two tickets:
1. Order jobs in a queue by # of reports, for v1.

2. Make this more customizable in settings. This might need to be in a post-v1 release due to the complexity of the requirements. BullMQ pulls jobs from a sorted set, and we can reorder the set and do a recalculation based on a custom prioritization. We could batch backfill through a Settings configuration, but a live filter would be harder and could introduce race condition problems.

juanmrad:
## Server-side weighted priority

Priority model that feeds BullMQ's native priority at enqueue. This keeps the dequeue + lock path (`dequeueNextJobWithLock`) fully intact with no concurrency races, no 50-job ceiling, and it scales to arbitrarily large queues.

The idea
Let an admin define weights over an allowlisted set of job properties, e.g.:

| Property | Weight |
| ------------- | ------------- |
| numReports | 10 |
| numPostsByUser | 2 |
| userScore | 5 |

At enqueue, compute a weighted score `score = (weight × normalize(property))` and translate it into a [BullMQ priority](https://docs.bullmq.io/guide/jobs/prioritized). The single integration point is the existing `queue.add(...)` call:

```ts
const newJob = await queue.add(
  bullJobId,
  { orgId, payload, id: jobId, createdAt, policyIds, reenqueuedFrom, enqueueSourceInfo },
  { removeOnComplete: true, jobId: bullJobId, priority: toBullPriority(score) },
);
```

Priority direction: in BullMQ, lower priority value = dequeued first, so `toBullPriority` has to invert the score (e.g. `MAX_PRIORITY - clamp(score)`). Higher weighted score -> reviewed sooner.
Mutable properties: `numReports` keeps climbing after enqueue. To honor that, call `job.changePriority({ priority })` when a new report lands on an already-queued item. Otherwise priority is frozen at first enqueue and re-reports won't bump it. 


[...]

PR #718, opened 2026-06-07
feat: 670 order jobs by num reports

### LINKED ISSUE #670: Order jobs in a queue by # of reports, for v1

calebmcquaid:

Addresses #670 (sub-ticket of #409). First of a 3-PR stack (this → weighted priority → per-reviewer skip); the branch was rebuilt from main per review feedback, so the old commit history and inline threads are superseded.

Adds a per-queue **job sort mode** for manual review: `FIFO` (default, behavior unchanged) or `NUM_REPORTS` (most-reported first). Instead of sorting on the client, the sort mode feeds **BullMQ's native job priority at enqueue time**, so the locked dequeue path (`dequeueNextJobWithLock`) is untouched — no concurrency races, no fetch window cap, and it scales to arbitrarily large queues.

How it works:
- At enqueue, `NUM_REPORTS` stamps `priority = MAX − reportCount`; a re-report of an already-queued item restamps it forward.
- Priorities cap at 2^21 − 1: at exactly 2^21, BullMQ's float64 sort key loses its arrival-order tiebreaker (details in `JobPriority.ts`).
- Changing a queue's sort mode re-stamps every pending job in a **background sweep** (serialized per queue, traced via SafeTracer) rather than inside the mutation. The sweep snapshots `(id, createdAt)` tuples first so it can't skip jobs while mutating, holds no full payloads in memory, and re-stamps oldest-first so equal priorities resolve to true arrival order. Note: this means the new order lands moments *after* the mutation returns.
- `iteratePendingJobsForQueue` now includes BullMQ's `prioritized` state. Without this, admin sweeps (e.g. reporter invalidation) would silently miss priority-enqueued jobs.
- An unrecognized stored `job_sort_type` throws instead of silently falling back to FIFO (per review); missing values (pre-migration rows) still mean FIFO.
- UI: sort-mode selector on the queue form, a Sort Order column on the queues dashboard, and a report-count column (default-sorted) on the jobs preview.

## Tests

- `JobPriority.test.ts` - unit tests for priority math: FIFO/NUM_REPORTS dispatch, clamping, re-report advancement, throw-on-unknown sort type.
- `QueueOperations.priorities.test.ts` - integration: recompute reaches jobs in the `prioritized` state, a sort-mode change re-stamps queued jobs via the background sweep, and re-stamping to a constant priority restores arrival order (behavioral: dequeues and asserts order).
- `(cd server && npm test)` for the touched suites and `(cd client && npx vitest run)` pass; lint/build/codegen clean.
- Verified end-to-end in the browser: most-reported job surfaces first, sort-mode changes re-sort the queue live, FIFO queues behave unchanged.

## (Optional) Rollout Plan

- One additive migration (`job_sort_type` column, default `'FIFO'`). Existing queues keep exact current behavior until an admin opts in.
- The stacked PRs (weighted priority, per-reviewer skip) should merge after this one; their bases retarget automatically.

## Summary by CodeRabbit

* **New Features**
  * Added a new Job Priorities settings page for managing queue weighting.
  * Queues can now be configured to sort by FIFO, most reported first, or custom weighting.
  * Dashboard tables now show sort order and report counts, with default sorting for reported jobs.

* **Bug Fixes**
  * Improved job advancement flow to more reliably move to the next available review.
  * Existing queue jobs are now re-ordered when queue sorting settings change.


client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -359,26 +372,109 @@ function ManualReviewJobReviewImpl(props: {
...
+    (targetJobId: string) => {
+      navigate(buildSortedReviewUrl(queueId!, targetJobId, sortMode!), {
+        replace: true,
+      });
+    },
+    [navigate, queueId, sortMode],
+  );
+
+  const getNextSortedJob = useCallback(async () => {
+    const result = await refetchJobInfo({ jobIds: null });
```
juanmrad: This assumes `reviewableQueues.jobs` returns the entire queue, but the server caps it. With `ids: null`, the resolver calls `getAllJobsForQueue`, which is hard-limited to 50. This means we are grabbing only top 50 jobs to do the sorting. 

Meaning we still use BullMQ's native order ( FIFO ) and picks the most-reported among those not the most-reported in the queue. For any real backlog (the case where prioritization matters), the actual top job usually isn't in that window, so the feature silently surfaces the wrong job.

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -903,8 +1031,10 @@ function ManualReviewJobReviewImpl(props: {
                     })),
                     queueId: queueId!,
                     jobId: job.id,
-                    // This is safe because we prevent both closedJob and lockToken from being null
-                    lockToken: lockToken!,
+                    // In FIFO mode lockToken comes from the URL (set during dequeue).
+                    // In sorted mode there's no dequeue lock — send empty string so
+                    // removeJob's fallback path removes the job without a lock.
+                    lockToken: lockToken ?? '',
```
juanmrad: This is bad. The lock is never acquired in sorted mode. FIFO review goes through `dequeueNextJobWithLock`, which hands out a `lockToken` so two reviewers can't grab the same job. Sorted mode skips dequeue entirely, so every reviewer in "# of reports" mode deterministically lands on the same top job and reviews it concurrently

juanmrad (changes requested):
While the logic is sane. The approach does not satisfy the requirement.

Rather than sorting on the client, I think the durable version of this is an queue-level priority model that feeds BullMQ's native priority at enqueue. This keeps the dequeue + lock path (`dequeueNextJobWithLock`) fully intact which provides the no concurrency races, no 50-job ceiling, and it scales to arbitrarily large queues.

I'll add more context on the issue itself. sorry I didn't before but was a bit busy.

calebmcquaid:
I made the suggested changes and moved the sorting to the server-side with BullMQ. There's now FIFO (BullMQ default) and NUM_REPORTS sorting options. I used the `priority` key in the BullMQ setup to sort the reportCounts to implement the NUM_REPORTS sorting.

An issue I'm running into: the skip behavior doesn't work with priority sorting. In FIFO mode, skipping a job works because releasing the lock puts it at the end of the queue and the next `getNextJob()` call returns a different job. With priority sorting, the skipped job still has the highest priority, so it gets dequeued again immediately and causes an infinite loop. I tried `moveToDelayed(Date.now() + {amount of time})` to move skipped jobs out of the waiting set, but when all remaining jobs in the queue are delayed, `worker.getNextJob()` blocks forever. 

Here are the options I see:

1. drainDelay: `1` on the worker makes `getNextJob` return null after 1s instead of blocking. But this changes polling behavior for all queues, not just priority ones.
2. Don't release the lock on skip: the job stays invisible for the `lockDuration` (10 min). Problem is the reviewer can't skip more jobs than the lock window allows.

What's the right approach here? The skip + priority code is committed but not completely functional.

@juanmrad

db/src/scripts/api-server-pg/2026.06.08T12.16.24.add_sort_type_to_mrt_queues.sql
```diff
@@ -0,0 +1,13 @@
...
+-- NUM_REPORTS: items with more reports get reviewed first.
+--
+-- Defaults to FIFO so nothing changes for existing queues.
+
+ALTER TABLE manual_review_tool.manual_review_queues
+    ADD COLUMN IF NOT EXISTS job_sort_type character varying(16) NOT NULL DEFAULT 'FIFO';
+
+ALTER TABLE manual_review_tool.manual_review_queues
+    ADD CONSTRAINT manual_review_queues_job_sort_type_check
+    CHECK (job_sort_type IN ('FIFO', 'NUM_REPORTS'));
```
juanmrad: This is not what we were expecting. Here https://github.com/roostorg/coop/issues/409#issuecomment-4645168411 I mentioned we wanted to allow users to provide some sort of weight to certain properties on the job that can be used for calculating the priority of tasks. Such that a weight can be calculated via operation to provide sorting to the queue. This only provides the ability to either sort by `FIFO` or based on `NUM_REPORTS` which is still static. 

It also has the same problem as we should be enqueuing with a priority set based on the value so BullMQ can calculate and keep the stack sorted and we can just dequeue from the top.

juanmrad (changes requested):
seems like the pr still not there with the request around having some normalizing calculation that is part of the settings. 

We want to be able to dynamically provide some sort of weight to the properties and have bullMQ handle the sorting internally as jobs are added to the queue.

calebmcquaid:
@coderabbit review

client/src/webpages/dashboard/components/table/Table.tsx
```diff
@@ -54,8 +56,17 @@ export default function Table(
...
 
   const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } =
-    /* @ts-ignore */
-    useTable({ columns, data, filterTypes }, useFilters, useSortBy);
+    useTable(
+      {
+        columns,
+        data,
+        filterTypes,
+        // @ts-expect-error react-table v7 types don't include sortBy on TableState
```
taobojlen: can you explain why we need this `ts-ignore-error`? are we using an incompatible version of a types package?

db/src/scripts/api-server-pg/2026.06.26T12.30.03.add_sort_type_to_mrt_queues.sql
```diff
@@ -0,0 +1,18 @@
...
+-- priority at enqueue so the dequeue + lock path is unchanged. Default is
+-- FIFO so existing queues keep behaving exactly like before. Admins opt
+-- in to NUM_REPORTS or WEIGHTED per queue from the queue form; WEIGHTED
+-- uses the org's job_priority_weights to compute the priority score.
+ALTER TABLE manual_review_tool.manual_review_queues
+    ADD COLUMN IF NOT EXISTS job_sort_type character varying(16) NOT NULL
+DEFAULT 'FIFO';
+
+-- Drop-then-add makes the constraint application idempotent: if the
+-- column was previously created with an older check (only FIFO/NUM_REPORTS),
```
taobojlen: nit: this comment doesn't really make sense in the context of this PR. the column would never have been created with the older check.

db/src/scripts/api-server-pg/2026.06.26T12.30.23.add_job_priority_weights_table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS manual_review_tool.job_priority_weights (
+    org_id     character varying(255) NOT NULL,
```
taobojlen: can we make this a proper foreign key?

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+  normalizeJobSortType,
+  toBullPriority,
+  type JobPropertyKey,
+} from './JobPriority.js';
+
+// The usable ceiling is 2^21 - 1, NOT BullMQ's PRIORITY_LIMIT of 2^21:
+// a priority of exactly 2^21 makes the float64 sort key (priority * 2^32 +
+// insertionCounter) hit 2^53, where precision loss rounds away the counter's
+// low bit and collapses FIFO ties. Hard-coded here so the tests fail loudly if
+// the constant is ever changed without intent (the original PR shipped with
```
taobojlen: This comment will be confusing for future readers since it references "the original PR" and so on.

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+const MAX_BULL_PRIORITY = 2_097_151;
+
+const orgId = 'org-1';
+
+function makeItem(opts?: {
+  itemId?: string;
+  itemTypeId?: string;
+  creator?: { id: string; typeId: string };
+}): ItemSubmissionWithTypeIdentifier {
+  // Opaque type from itemProcessingService; cast through unknown rather
```
taobojlen: Can you explain why this is better than just using the normal constructor? (asking because it took me a minute to grok what the comment was saying here)

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+    });
+
+    test('non-integer score is rounded before subtraction', () => {
+      // 100.4 rounds down to 100, 100.6 rounds up to 101.
+      expect(toBullPriority(100.4)).toBe(MAX_BULL_PRIORITY - 100);
+      expect(toBullPriority(100.6)).toBe(MAX_BULL_PRIORITY - 101);
+    });
+
+    test('output never exceeds BullMQ priority cap', () => {
+      // Guards against the bug we hit shipping the wrong ceiling.
```
taobojlen: again, this comment references something that does not exist in the code, so it is confusing!

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+        reports: 5,
+        weights: new Map([
+          ['numReports', 100],
+          ['userScore', 100],
+        ]),
+      });
+      expect(withWeights).toBe(noWeights);
+    });
+
+    test('re-report bumps the same item forward', async () => {
```
taobojlen: nit: this duplicates the `priority is driven by numReports alone` test above

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+        sortType: JobSortType.WEIGHTED,
+        reports: 5,
+        weights: new Map([['numReports', 1]]),
+      });
+      const heavy = await priorityFor({
+        sortType: JobSortType.WEIGHTED,
+        reports: 5,
+        weights: new Map([['numReports', 100]]),
+      });
+      // The user-visible knob v1.1 exposes: cranking weight makes that
```
taobojlen: a reader will wonder "what is v1.1?" here. let's remove this reference

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+    });
+
+    test('weight 0 turns the property off (priority lands at MAX)', async () => {
+      const priority = await priorityFor({
+        sortType: JobSortType.WEIGHTED,
+        reports: 42,
+        weights: new Map([['numReports', 0]]),
+      });
+      // weight 0 * reports 42 = 0 contribution => MAX. Confirms 0 is
+      // honoured as a real value, not coerced via `||` to a default.
```
taobojlen: i don't understand what "not coerced via `||` to a default" means here.

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+        sortType: JobSortType.WEIGHTED,
+        reports: 100,
+        userScore: 1,
+        weights: new Map(),
+      });
+      expect(priority).toBe(MAX_BULL_PRIORITY);
+    });
+
+    test('worse userScore => sooner dequeue (the inversion is wired)', async () => {
+      // UserScore 1..5 with 5 = neutral, 1 = worst. After normalize:
```
taobojlen: i think all the comments in this file need a rewrite! they are really hard to read.

for context, i am reading this test file first, because with LLMs the tests are really the most important part. but, a lot of these comments seem to reference the specific implementation (probably best avoided, LLMs or no).

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+      });
+      expect(getUserScore).toHaveBeenCalledWith(orgId, {
+        id: 'user-42',
+        typeId: 'user-type-2',
+      });
+    });
+
+    test('report count saturates: +10 reports matters far more at low counts than high', async () => {
+      // Normalized as reports / (reports + K), so equal increments have
+      // diminishing returns — a viral item can't infinitely outrank everything.
```
taobojlen: i think this functionality falls under YAGNI.

if a viral item gets reported a lot -- well, for one, you *want* it to outrank everything, but also it really doesn't matter that much. there's still just one job that gets reviewed quickly, no?

IMO this diminishing-returns-normalization adds unnecessary complexity!

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+        }
+      }
+    });
+  });
+
+  describe('normalizeJobSortType', () => {
+    test('passes known sort types through unchanged', () => {
+      expect(normalizeJobSortType('FIFO')).toBe(JobSortType.FIFO);
+      expect(normalizeJobSortType('NUM_REPORTS')).toBe(JobSortType.NUM_REPORTS);
+      // Regression: WEIGHTED used to be collapsed to FIFO at enqueue time, so
```
taobojlen: this is not a regression because we never merged whatever bug the comment references! i.e.

> WEIGHTED used to be collapsed to FIFO at enqueue time

no it didn't 😅

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,451 @@
...
+      // a weighted queue ignored its weights for newly enqueued jobs.
+      expect(normalizeJobSortType('WEIGHTED')).toBe(JobSortType.WEIGHTED);
+    });
+
+    test('defaults missing / unrecognized values to FIFO', () => {
```
taobojlen: i don't think this is desirable behaviour.

LLMs love writing fallbacks and whatnot but really, if something in your system is not as you expect, you usually *want* it to fail loudly. it means your mental model of the system is wrong!

server/services/manualReviewToolService/modules/JobPriorityWeights.test.ts
```diff
@@ -0,0 +1,190 @@
...
+
+import getBottle from '../../../iocContainer/index.js';
+import { makeTestWithFixture } from '../../../test/utils.js';
+import JobPriorityWeights from './JobPriorityWeights.js';
+
+// JobPriorityWeights talks only to manual_review_tool.job_priority_weights,
+// which has no foreign keys -- so the fixture just spins up the ops class
+// and a unique org id. No need to seed an org row.
+describe('JobPriorityWeights', () => {
+  const testWithFixtures = makeTestWithFixture(async () => {
```
taobojlen: please use `makeTransactionalTestWithFixture` instead (you might need to rebase to get it).

that test function runs the test inside a DB transaction, which means you don't need to manually clean up at the end of it! the transaction just rolls back.

server/services/manualReviewToolService/modules/JobPriorityWeights.test.ts
```diff
@@ -0,0 +1,190 @@
+import { v1 as uuidv1 } from 'uuid';
+
+import getBottle from '../../../iocContainer/index.js';
+import { makeTestWithFixture } from '../../../test/utils.js';
+import JobPriorityWeights from './JobPriorityWeights.js';
+
+// JobPriorityWeights talks only to manual_review_tool.job_priority_weights,
+// which has no foreign keys -- so the fixture just spins up the ops class
```
taobojlen: see my comment in the migration -- i think we want the table to have a FK :)

server/services/manualReviewToolService/modules/QueueOperations.test.ts
```diff
@@ -203,9 +235,8 @@ describe('QueueOperations', () => {
     },
   );
 
-  // Regression: `deleteAllJobsFromQueue` is irreversible and used to accept
-  // EDIT_MRT_QUEUES (held by moderator managers) -- that gap accidentally
-  // cleared a production queue. It now requires MANAGE_ORG.
+  // Regression: deleteAllJobsFromQueue now requires MANAGE_ORG (was the weaker
+  // EDIT_MRT_QUEUES, which accidentally let a manager clear a production queue).
```
taobojlen: nit: this diff seems unrelated to the PR; best to avoid unrelated changes/refactorings like this to make reviewing easier!

server/services/manualReviewToolService/modules/QueueOperations.test.ts
```diff
@@ -231,4 +262,236 @@ describe('QueueOperations', () => {
       ).resolves.toBeUndefined();
     },
   );
+
+  // Regression: recompute used to skip BullMQ's 'prioritized' state, so weight
```
taobojlen: not really a regression, i don't think?

server/services/manualReviewToolService/modules/QueueOperations.test.ts
```diff
@@ -231,4 +262,236 @@ describe('QueueOperations', () => {
...
+  testWithQueueAndActions()(
+    'recomputePrioritiesForQueue breaks priority ties by createdAt (oldest first)',
+    async ({ org, queue, mrtService }) => {
+      const queueOps = mrtService['queueOps'];
+      const itemTypeId = uid();
+
+      const payloadFor = makePayloadFor(itemTypeId);
+
+      const base = new Date('2026-01-01T00:00:00.000Z').getTime();
+      // Arrival order is the REVERSE of lexicographic id order, so a fallback to
```
taobojlen: this says that arrival order is EDCBA but a few lines down we insert in order CAEBD -- which is it? hard to understand this

server/services/manualReviewToolService/modules/QueueOperations.test.ts
```diff
@@ -231,4 +262,236 @@ describe('QueueOperations', () => {
...
+          priority,
+          jobPayload: {
+            createdAt: new Date(createdAtByItem[itemId]),
+            policyIds: [],
+            payload: payloadFor(itemId),
+          },
+        });
+      }
+
+      // Re-stamp through FIFO's real ceiling (toBullPriority(0)) so a regression
```
taobojlen: i don't understand this comment at all, tbh. can you explain it?

like, why are we setting scores to 0 -- what exactly are we trying to assert in this test?

it seems like it would be much clearer to enqueue some items with varying priorities, change the queue to FIFO, then dequeue them and assert that they did indeed come back in FIFO order.

right now it's reading scores from redis and so on; feels like we're testing implementation rather than behaviour

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -894,13 +911,60 @@ export default class QueueOperations {
...
+    const { orgId, queueId, getPriority } = opts;
+    const queue = await this.#getBullQueue(orgId, queueId);
+    const batchSize = 200;
+    let start = 0;
+
+    // Collect every pending job first, then re-stamp them oldest-first below.
+    const allJobs: Job<StoredManualReviewJob>[] = [];
+    while (true) {
+      // Priority-enqueued jobs live in BullMQ's 'prioritized' state, not
+      // 'waiting'. Omitting it meant weight changes never re-sorted queued
```
taobojlen: another comment referencing the development process as if it was a production bug -- makes it harder to understand the final code!

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -894,13 +911,60 @@ export default class QueueOperations {
...
+    queueId: string;
+    getPriority: (job: ManualReviewJob) => Promise<number>;
+  }) {
+    const { orgId, queueId, getPriority } = opts;
+    const queue = await this.#getBullQueue(orgId, queueId);
+    const batchSize = 200;
+    let start = 0;
+
+    // Collect every pending job first, then re-stamp them oldest-first below.
+    const allJobs: Job<StoredManualReviewJob>[] = [];
```
taobojlen: this pulls all jobs in the queue into memory -- that could lead to OOMs for large queues with 10s or 100s of thousands of jobs. can we do this in batches instead?

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -897,11 +996,48 @@ export class ManualReviewToolService {
...
+    // queue, not just affect future enqueues — otherwise the queue keeps its
+    // old ordering until every job churns out. setJobPriorityWeights does this
+    // for weight changes; mirror it here when the sort type actually changed.
+    if (
+      previous !== undefined &&
+      !updated.isAppealsQueue &&
+      previous.jobSortType !== updated.jobSortType
+    ) {
+      const weights = await this.jobPriorityWeights.loadForOrg(input.orgId);
+      await this.queueOps.recomputePrioritiesForQueue({
```
taobojlen: because this sets the new priorities one by one, it could take a long time (it's sort of like an N+1 but idk if that's avoidable with how bullMQ works).

anyway, i think this should be done in a background job rather than in the request itself since it might time out!

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -897,11 +996,48 @@ export class ManualReviewToolService {
...
+        orgId: input.orgId,
+        queueId: input.queueId,
+        getPriority: async (job) =>
+          getJobPriorityForItem({
+            orgId: input.orgId,
+            item: job.payload.item,
+            sortType: normalizeJobSortType(updated.jobSortType),
+            deps: {
+              getNumTimesReported: this.getNumTimesReported(),
```
taobojlen: this will also be an N+1, i think? i don't know if this is easily avoidable, but at the very least confirms that we should be doing this in a background job!

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -1466,6 +1604,16 @@ export class ManualReviewToolService {
     userId: string;
   }) {
     await this.skipOps.logSkip(opts);
+    // Per-reviewer skip: hide this job from THIS reviewer for the skip window
+    // (the dequeue path reads this back), while it stays available to everyone
+    // else. The reviewer separately releases their lock so it returns to the
+    // shared pool immediately.
+    await this.queueOps.recordReviewerSkip({
```
taobojlen: is this related to this PR? (if not, fixing it in a separate PR is preferable)

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1194,47 +1262,119 @@ export default class QueueOperations {
...
     const worker = await this.getBullWorker({ orgId, queueId });
 
-    let hasDecision = true;
-    while (hasDecision) {
-      const job = await worker.getNextJob(lockToken);
-
-      if (!job) {
-        return null;
-      }
+    // Jobs this reviewer skipped within their 30-min window (the lock token is
```
taobojlen: i kind of suspect that this is unrelated to the feature in this PR. before i review it, can you confirm? (if it really is unrelated then separate PR is preferable)

server/services/manualReviewToolService/modules/JobPriority.ts
```diff
@@ -0,0 +1,169 @@
...
+  FIFO: 'FIFO',
+  NUM_REPORTS: 'NUM_REPORTS',
+  WEIGHTED: 'WEIGHTED',
+} as const;
+export type JobSortType = (typeof JobSortType)[keyof typeof JobSortType];
+
+// Coerce a raw (DB-stored) `job_sort_type` string into a known JobSortType,
+// defaulting to FIFO for missing/unrecognized values. Callers that drive
+// priority off a queue's persisted sort type should route through this so an
+// unexpected value degrades to FIFO instead of throwing — and, critically, so
```
taobojlen: it should throw if there's an unexpected value!

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -44,29 +20,52 @@ import {
...
-import { ITEM_TYPE_FRAGMENT } from '../../rules/rule_form/RuleForm';
-import { JOB_FRAGMENT } from './jobFragment';
-import ManualReviewJobDequeueErrorComponent from './ManualReviewJobDequeueErrorComponent';
-import MergedReportsComponent from './MergedReportsComponent';
-import ReportInfoComponent from './ReportInfoComponent';
-import ManualReviewJobContentView from './v2/ManualReviewJobContentView';
-import ManualReviewJobEmptyQueue from './v2/ManualReviewJobEmptyQueue';
-import { ManualReviewJobOtherItemsComponent } from './v2/ManualReviewJobOtherItemsComponent';
+} from '@/graphql/generated';
+import Sidebar1 from '@/icons/lni/Design/sidebar-1.svg?react';
```
taobojlen: looks like an awful lot of import ordering changes here! can you rebase against main to get the already-formatted file? should reduce the diff a lot

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -360,38 +359,41 @@ function ManualReviewJobReviewImpl(props: {
...
+      jobIds: closedJob ? [closedJob.id] : jobId ? [jobId] : [],
+    },
     fetchPolicy: 'no-cache',
   });
 
   const [
     getNextJob,
     { data: jobData, loading: jobDataLoading, error: jobDataError },
   ] = useGQLDequeueManualReviewJobMutation({
-    variables: { queueId: queueId! },
```
taobojlen, line 387: why move this variable to the call site, rather than here?

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -909,7 +925,6 @@ function ManualReviewJobReviewImpl(props: {
                     })),
                     queueId: queueId!,
                     jobId: job.id,
-                    // This is safe because we prevent both closedJob and lockToken from being null
```
taobojlen, line 912: why remove this comment -- is it no longer true?

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -801,15 +815,17 @@ function ManualReviewJobReviewImpl(props: {
...
+    });
 
     // If there's no next job, redirect to the queues page
     if (result.data?.dequeueManualReviewJob == null) {
       navigate('/dashboard/manual_review/queues');
     }
   };
 
-  if (loading || jobDataLoading || (!closedJob && !lockToken)) {
+  if ((loading && !data) || jobDataLoading || (!closedJob && !lockToken)) {
```
taobojlen: can there be any state where `loading && data` is true? if so, why shouldn't that show a loading state, given that `loading` is true?

client/src/webpages/dashboard/mrt/ManualReviewQueueForm.tsx
```diff
@@ -628,6 +638,29 @@ export default function ManualReviewQueueForm() {
...
+        >
+          <Option value="FIFO">First in, first out</Option>
+          <Option value="NUM_REPORTS">Most reported first</Option>
+          <Option value="WEIGHTED">Custom (weighted)</Option>
+        </Select>
+        <div className="mt-2 text-sm text-slate-400">
+          {jobSortType === 'FIFO' &&
+            'Jobs are reviewed in the order they were received.'}
+          {jobSortType === 'NUM_REPORTS' &&
+            'Jobs with more user reports are surfaced to reviewers first. Report counts are determined at enqueue time and updated when new reports arrive.'}
```
taobojlen:
```suggestion
            'Jobs with more user reports are surfaced to reviewers first.'}
```

client/src/webpages/dashboard/mrt/ManualReviewQueueForm.tsx
```diff
@@ -628,6 +638,29 @@ export default function ManualReviewQueueForm() {
...
+          {jobSortType === 'WEIGHTED' &&
+            'Jobs are scored using the weights configured in Settings → Review Console → Job Priority Weights, and reviewed in priority order.'}
```
taobojlen:
```suggestion
            'Jobs are ordered using the weights configured in Settings → Review Console → Job Priority Weights.'}
```

client/src/webpages/settings/JobPrioritySettings.tsx
```diff
@@ -0,0 +1,208 @@
...
+const PROPERTY_LABELS: ReadonlyArray<{
+  property: GQLJobPriorityProperty;
+  label: string;
+  help: string;
+  example: (weight: number) => string;
+}> = [
+  {
+    property: GQLJobPriorityProperty.NumReports,
+    label: '# of User Reports',
+    help: "Items with more user reports are reviewed sooner, with diminishing returns so a single viral item can't completely crowd out everything else. Set to 0 to ignore.",
```
taobojlen:
```suggestion
    help: "Items with more user reports are reviewed sooner. Set to 0 to ignore.",
```

client/src/webpages/settings/jobPriorityWeightSummary.ts
```diff
@@ -0,0 +1,40 @@
...
+// Data-driven so it scales automatically as more properties are added.
+export function summarizeWeighting(
+  entries: ReadonlyArray<{ label: string; weight: number }>,
+): string {
+  const active = entries
+    .filter((e) => e.weight > 0)
+    .sort((a, b) => b.weight - a.weight);
+
+  if (active.length === 0) {
+    return 'All weights are 0, so these queues fall back to first-in, first-out order — nothing here changes review priority.';
```
taobojlen:
```suggestion
    return 'All weights are 0, so these queues fall back to first-in, first-out order.';
```

taobojlen (changes requested):
okay, apologies for all the comments here!

this is a very big PR -- it took me a long time to review. in general, splitting into smaller (ideally stacked) PRs is appreciated and makes life easier for reviewers!

i think it also groups unrelated changes + some refactoring -- i prefer one change per PR, not just for ease of review, but also for a more useful git history / changelogs.

server/services/manualReviewToolService/modules/QueueOperations.ts
taobojlen: do we need to add `prioritized` here?

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
taobojlen, line 758: please add `queueId` here, if we do have to move `queueId` to the callsite for some reason

wayjaywang:
Some quick comments/questions about the proposed UI
- When "Custom (weighted)" is selected under Job Sort Order (1:13 in the loom video), is it possible to have the explanation text link directed to the Job Priority Weights settings page?
- In other words, can we change the text to "Jobs are scored using the weights configured in Job Priority Weights and reviewed in priority order." with "Job Priority Weights" a clickable link to the settings page
- Also, in settings, Job Priority Weights should be part of the Review Console settings tab.
- Lastly, just to make sure, is Review Priority Score visible to users in the UI?

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -934,9 +955,66 @@ export default class QueueOperations {
...
+    }
+
+    pending.sort((a, b) => a.createdAtMs - b.createdAtMs);
+
+    for (const { bullId } of pending) {
+      const bullJob = await queue.getJob(bullId);
+      // Dequeued or removed since the snapshot — nothing to re-stamp.
+      if (!bullJob) continue;
+      const priority = await getPriority(bullJob.data as ManualReviewJob);
+      await bullJob.changePriority({ priority });
```
taobojlen, line 1103: did some more digging, sharing my findings here:

this is still an N+1 but i don't think bullMQ provides a bulk setter for priorities. their [docs](https://docs.bullmq.io/guide/jobs/prioritized) state that adding a prioritized job is `O(log(n))`, so assuming that changing a priority is the same, then recomputing priorities is `O(n*log(n))`. not great but perhaps unavoidable -- at least we're doing it in a backgroung job.

server/services/manualReviewToolService/dbTypes.ts
```diff
@@ -87,6 +87,7 @@ export type ManualReviewToolServicePg = {
     is_default_queue: boolean;
     is_appeals_queue: boolean;
     auto_close_jobs: boolean;
+    job_sort_type: ColumnType<string, string | undefined, string>;
```
taobojlen: can this really be undefined?

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -907,11 +931,96 @@ export class ManualReviewToolService {
...
+        queueId: input.queueId,
+      });
+    const updated = await this.queueOps.updateManualReviewQueue(input);
+
+    // Changing the sort mode has to re-sort the jobs already sitting in the
+    // queue, not just affect future enqueues — otherwise the queue keeps its
+    // old ordering until every job churns out.
+    if (
+      previous !== undefined &&
+      !updated.isAppealsQueue &&
```
taobojlen, line 959: how come we don't recompute priorities in appeals queues?

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -907,11 +931,96 @@ export class ManualReviewToolService {
...
+
+    return updated;
+  }
+
+  // In-process chains of background priority recomputes, keyed by queue.
+  // Serializes rapid consecutive sort-mode changes for the same queue so their
+  // per-job re-stamps can't interleave (the last change wins). In-process
+  // only: sweeps from two server instances can still race, in which case the
+  // slower sweep's ordering wins.
+  readonly #priorityRecomputes = new Map<string, Promise<void>>();
```
taobojlen: i'm afraid this won't really work, because a deployment might have multiple workers/processes handling background jobs.

the canonical way to handle this is probably with bullMQ's [group rate limiting](https://docs.bullmq.io/guide/rate-limiting#group-keys). that way we can ensure that only one of these jobs is running at a time, per queue.

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -907,11 +931,96 @@ export class ManualReviewToolService {
...
+          async () =>
+            this.queueOps.recomputePrioritiesForQueue({
+              orgId: opts.orgId,
+              queueId: opts.queueId,
+              getPriority: async (job) =>
+                getJobPriorityForItem({
+                  orgId: opts.orgId,
+                  item: job.payload.item,
+                  sortType: opts.sortType,
+                  deps: { getNumTimesReported: this.getNumTimesReported() },
```
taobojlen: this will still hit `getNumTimesReported()` one time for each job, right? given that this can execute a clickhouse/data warehouse query, i think that's an N+1 that's too expensive (as in, might knock over the DB).

a better alternative would be to get the number of times reported in one query or in batches, i.e. by using a `GROUP BY` query against the data warehouse, and then keeping a mapping like `jobId -> numTimesReported` in memory.

taobojlen (changes requested):
A few footguns worth fixing (global job deduplicating/rate limiting, and the N+1 against the data warehouse). But it's going in the right direction!

server/services/manualReviewToolService/manualReviewToolService.ts
juanmrad: +1 In an appeals queue switched to NUM_REPORTS will leave already-queued jobs on the old order until they churn. Either recompute for appeals too, or hide/disable sort mode on appeals and force FIFO.

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: Good call. I patched this with a Redis lock using `IORedis` which already existed. Groups looks like it's a BullMQ Pro feature and I think we're on the open source version.

They removed support for groups in 3.0 and we're on 5.0

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: I went ahead and hid the selector for the appeals queues and forced FIFO

server/services/manualReviewToolService/dbTypes.ts
calebmcquaid: Switched to a union. I believe it can be on insert as the column defaults to FIFO

juanmrad (commented):
A few things to address:

**1. `numReports` is a locale string but the column sorts it as an integer**

`ManualReviewQueueJobsPreview.tsx` runs `toLocaleString('en')` before storing `numReports` in the table data, which gives you strings like `"1,000"`. The column uses `integerSort` though, so it'll sort lexicographically instead of numerically. Store the raw number and only format it in the cell renderer.

**2. `getNumTimesReported` / `getNumTimesReportedForItems` are public on the class**

These constructor params are `readonly` but not `private readonly`, so they're exposed as public API on `ManualReviewToolService`. They're DI wiring internals — mark them `private readonly` like `tracer` and `moderationConfigService`.

**3. `priority: undefined` gets passed explicitly to BullMQ**

When the FIFO path returns `undefined`, that still lands as an explicit `priority: undefined` key in BullMQ's job options object. Today BullMQ ignores it, but if a future version treats a present-but-undefined key differently from a missing one, FIFO jobs would silently end up in the `prioritized` set. Easy fix — conditionally spread: `...(priority != null && { priority })`.

**4. No visibility when `#MAX_RECOMPUTE_PASSES` is exhausted**

If someone toggles the sort mode rapidly and the sweep hits the 5-pass cap, it exits silently. The queue stays stale until the next enqueue or mode change, and nobody knows. Add a span event or warning so operators can see when this happens.

server/services/manualReviewToolService/modules/PriorityRecomputeLock.ts
```diff
@@ -0,0 +1,105 @@
...
+  return redis.call("DEL", KEYS[1])
+else
+  return 0
+end
+`;
+
+/**
+ * How long a held lock survives without the holder releasing it.
+ *
+ * Deliberately generous rather than watchdog-extended: a sweep is O(pending
```
taobojlen, line 23: what does "watchdog-extended" mean in this context?

server/services/manualReviewToolService/modules/PriorityRecomputeLock.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+export const RECOMPUTE_LOCK_TTL_MS = 5 * 60 * 1000;
+
+/**
+ * Cross-instance coordination for manual review queue priority sweeps.
+ *
+ * Two pieces:
+ *
+ * - A **lock** per (org, queue), so only one sweep runs at a time no matter
+ *   how many API processes are deployed. An in-process Map can't do this.
```
taobojlen, line 38:
```suggestion
 *   how many API processes are deployed.
```

comments don't need to (and shouldn't) reference past implementations in general, though LLMs love to do this

server/services/manualReviewToolService/modules/PriorityRecomputeLock.ts
```diff
@@ -0,0 +1,105 @@
...
+`;
+
+/**
+ * How long a held lock survives without the holder releasing it.
+ *
+ * Deliberately generous rather than watchdog-extended: a sweep is O(pending
+ * jobs) Redis round-trips, and a lock that expires mid-sweep would let a
+ * second instance start one concurrently — the exact thing this prevents. The
+ * cost of erring long is that a process which dies mid-sweep leaves the queue
+ * unswept for up to this long. That degrades ordering, it doesn't break it:
```
taobojlen, line 27: i find these LLM comments really dense and hard to read. no comment is preferable to a confusing one!

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -904,11 +933,170 @@ export class ManualReviewToolService {
...
+  }
+
+  // In-flight sweeps, tracked only so tests can await them. Coordination
+  // between instances is the Redis lock's job, not this set's.
+  readonly #priorityRecomputes = new Set<Promise<void>>();
+
+  // A sweep that keeps re-sorting while the mode changes under it would spin
+  // forever if an admin toggled repeatedly. Bounded; the version stays bumped,
+  // so the next change picks up where this left off.
+  static readonly #MAX_RECOMPUTE_PASSES = 5;
```
taobojlen, line 986: i think this is unnecessary complexity, and leads to a weird failure mode if someone *does* re-sort a queue many times.

server/services/manualReviewToolService/manualReviewToolService.ts
taobojlen: OK! i find the current redis lock quite complex -- especially the handling where a job might take longer than the TTL.

i think we can avoid a lot of this by bullMQ's native rate limiting, even without groups. the way to do it would be to create a new bullMQ queue for these operations (probably up to one bullMQ queue per coop queue) with `limit: 1`.

server/services/manualReviewToolService/manualReviewToolService.ts
taobojlen: @juanmrad wrote in another comment:

> If someone toggles the sort mode rapidly and the sweep hits the 5-pass cap, it exits silently. The queue stays stale until the next enqueue or mode change, and nobody knows. Add a span event or warning so operators can see when this happens.

that's what i mean, but i don't think that a span event or warning is enough! all of this complexity is unnecessary, we can avoid it entirely.

taobojlen (changes requested):
i think the job locks introduce an awful lot of complexity and failure modes.

also, it'd be helpful for reviewers if you could ensure that CI is green before requesting a review, and also go through old comment threads (both from humans and also AI reviews) and resolve the ones that are fixed or no longer relevant. if you disagree with a suggestion (from a human or AI), then it's helpful to write a comment explaining why before resolving the thread.

taobojlen (changes requested):
@calebmcquaid i can see that you requested a review from the community group but there are a bunch of unaddressed comments from me, so i'm unsure of the status of this PR -- are you waiting for reviews from others, or still working on addressing those comments?

assuming you'd like a re-review:

a lot of my comments are on code that has now changed (i.e. it's marked outdated). when that happens it's very hard for me to know what changed -- did the code just change as a result of a rebase? did you address the comment, decide to go some other route, or decide that changing nothing is the best approach?

and please keep in mind that when i write PR comments, i'm always up for a discussion or disagreement! the goal is to merge working, clean code, not to block merging for the sake of purity.

server/services/manualReviewToolService/modules/PriorityRecomputeLock.ts
calebmcquaid: It's a lock a background timer keeps restarting while work is in progress

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: Sounds good. I'm going to queue the sweeps instead. I can't delete the loop as it's what catches sweeps that would be dropped

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Moved to the server

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Fixed

client/src/webpages/dashboard/components/table/Table.tsx
calebmcquaid: react-table v7's TableState typings don't include the fields we set

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: The loser now waits for the lock instead of giving up. I believe for OSS `limit: 1` is a rate limit. It looks like the cross instance concurrency is Pro BullMQ

server/services/manualReviewToolService/modules/JobPriority.test.ts
calebmcquaid: This is a good point. Adjusted

server/services/manualReviewToolService/modules/JobPriority.test.ts
calebmcquaid: Removed

server/services/manualReviewToolService/modules/JobPriority.test.ts
calebmcquaid: Dropped

server/services/manualReviewToolService/modules/JobPriority.test.ts
calebmcquaid: Unknown values throw now. null/undefined still fallback to FIFO

server/services/manualReviewToolService/modules/JobPriorityWeights.test.ts
calebmcquaid: Moved to #892

server/services/manualReviewToolService/modules/JobPriorityWeights.test.ts
calebmcquaid: Moved to #892

server/services/manualReviewToolService/modules/QueueOperations.test.ts
calebmcquaid: Fixed

server/services/manualReviewToolService/modules/QueueOperations.test.ts
calebmcquaid: This is fixed. The `arrivalOrder` drives both inserts and the assertion now

server/services/manualReviewToolService/modules/QueueOperations.ts
calebmcquaid: I reworked this. It doesn't hold Job objects, but it does still use one array for the whole queue. Batching would need an external sort or redis structure. Do you think the queue size would justify that architecture?

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: Fixed

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: Fixed

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: This has been removed

server/services/manualReviewToolService/modules/QueueOperations.ts
calebmcquaid: Unrelated and removed

server/services/manualReviewToolService/modules/JobPriority.ts
calebmcquaid: Fixed

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Removed

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Reverted

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Added back

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Reverted

server/services/manualReviewToolService/modules/QueueOperations.ts
calebmcquaid: Good catch. Fixed

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: This shouldn't be needed anymore. `queueId` went back on the hook rather than the call site

client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
```diff
@@ -830,12 +852,26 @@ export default function ManualReviewQueuesDashboard() {
...
+              <Tooltip delayDuration={0}>
+                <TooltipTrigger
+                  aria-label="Why is this unavailable?"
+                  className="flex items-center justify-between w-full text-gray-500 cursor-help"
+                >
+                  —
+                  <Info className="w-3.5 h-3.5 shrink-0" aria-hidden />
+                </TooltipTrigger>
+                <TooltipContent>
+                  Not tracked for queues with a custom sort order.
```
taobojlen, line 865: Hmm, @julietshen flagging this gotcha: we can only show the oldest job in a queue if the queue is sorted using FIFO. If it's sorted by anything else, we show this message instead.

From a product perspective it's pretty weird, IMO! It's a side effect of how we store jobs. https://github.com/roostorg/coop/issues/1200 would fix this.

server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
```diff
@@ -190,6 +190,46 @@ export class ClickhouseReportingAnalyticsAdapter implements IReportingAnalyticsA
...
+    const uniqueItemIds = [...new Set(itemIds)];
+
+    for (let i = 0; i < uniqueItemIds.length; i += chunkSize) {
+      const chunk = uniqueItemIds.slice(i, i + chunkSize);
+      const rows = await this.query<{
+        reported_item_id: string;
+        count: number;
+      }>(
+        `
+        SELECT
```
taobojlen, line 214: I'm concerned about this query. The Clickhouse table is partitioned by date, but we can't filter by date here. For each chunk, Clickhouse has to scan every single partition, so a queue with 100k jobs would execute 100 sequential queries -- meaning it has to scan the same data over and over again.

This is an architectural limitation of how we store reports, and honestly, I'd argue it's a blocker for this feature! A much better approach would be a new column, stored on jobs (dependent on #1200) that aggregates live as new reports come in.

If we had a real production-scale deployment, we could actually quantify this, which would be my preference. But in lieu of that we can only think architecturally.

I apologize for not catching this initially. It's only now that the earlier N+1 is fixed, and as I've become more familiar with Coop's data architecture, that I noticed this.

server/services/manualReviewToolService/modules/JobPriority.test.ts
```diff
@@ -0,0 +1,242 @@
...
+import type { ItemSubmissionWithTypeIdentifier } from '../../itemProcessingService/makeItemSubmissionWithTypeIdentifier.js';
+import {
+  getJobPrioritiesForItems,
+  getJobPriorityForItem,
+  JobSortType,
+  normalizeJobSortType,
+  toBullPriority,
+} from './JobPriority.js';
+
+// Deliberately hard-coded rather than imported: BullMQ silently breaks FIFO
```
taobojlen: IMO this comment is unnecessary. It's hard to parse:

> BullMQ silently breaks FIFO tie-breaking for priorities above 2^21 - 1 (the float64 sort key loses integer precision at 2^53)

Imagine being a reader just coming across this file -- it won't make sense! To understand this comment, you need to know:

1. how FIFO tie-breaking is
2. what float64 sort key is being referred to
3. why losing integer precision matters

I'm a reviewer and I can't answer these things! So it mostly makes me confused. LLMs do this a lot: they have all this context, and they write comments (or code, or text...) that assumes that the reader has all the same info they do. But that's not how humans work!

server/services/manualReviewToolService/modules/JobPriority.ts
```diff
@@ -0,0 +1,122 @@
...
+
+// Coerce a raw (DB-stored) `job_sort_type` string into a known JobSortType.
+// Missing values (queues created before the sort-type column existed, or a
+// queue that wasn't found) mean FIFO — the historical behavior. Anything else
+// is a bug somewhere upstream (the column has a CHECK constraint), so fail
+// loudly rather than silently mis-sorting the queue.
+export function normalizeJobSortType(
+  raw: string | null | undefined,
+): JobSortType {
+  if (raw == null) {
```
taobojlen: This functionality is unnecessary -- we already default queues to FIFO in the data migration. There's no need to add complexity to handle a scenario that'll never come up.

server/services/manualReviewToolService/modules/JobPriority.ts
```diff
@@ -0,0 +1,122 @@
...
+    return JobSortType.FIFO;
+  }
+  switch (raw) {
+    case JobSortType.FIFO:
+    case JobSortType.NUM_REPORTS:
```
taobojlen: is all of this necessary? instead of `normalizeJobSortType(someSortType)`, can we not just use `someSortType` directly? AFAICT, when we call this function in this PR, the input is already typed as `JobSortType`. e.g. in `manualReviewToolService.ts`, instead of

```ts
sortType: normalizeJobSortType(targetQueue?.jobSortType),
```

you could do

```ts
sortType: normalizeJobSortType(targetQueue?.jobSortType ?? JobSortType.FIFO),
```

server/services/manualReviewToolService/modules/PriorityRecomputeLock.ts
```diff
@@ -0,0 +1,103 @@
+import type IORedis from 'ioredis';
```
taobojlen, line 1: I have to be honest and say that I'm becoming increasingly convinced that #1200 is the better approach -- it would avoid all this complexity.

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1656,31 +1762,29 @@ export default class QueueOperations {
...
-    // each state to find the oldest overall. NB: the equivalent
-    // queue.getJobs([state], 0, 0) defaults to descending order and would
-    // return the *newest* job instead.
-    const [waitingJobs, delayedJobs] = await Promise.all([
+    // getWaiting/getDelayed return oldest-first, so their first entry is the
+    // oldest. `prioritized` is ordered by priority, so scan it instead.
+    const [waitingJobs, delayedJobs, prioritizedJobs] = await Promise.all([
       queue.getWaiting(0, 0),
       queue.getDelayed(0, 0),
+      queue.getPrioritized(0, OLDEST_JOB_PRIORITIZED_SCAN_LIMIT - 1),
```
taobojlen, line 1778: Wait, how do we use this data? I thought we didn't show the oldest job for queues sorted by NUM_REPORTS?

Anwyay, this cap means that the data can't be trusted by end users. Again, this is a limitation of how we store jobs, and would be solved by #1200 . But we definitely should not be showing the users this data.

My question is: if we aren't showing this data to the user, why are we fetching it here? Am I mixing up two things?

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -918,11 +947,131 @@ export class ManualReviewToolService {
...
+        queueId: input.queueId,
+      });
+    }
+
+    return updated;
+  }
+
+  // In-flight sweeps, tracked only so tests can await them. Coordination
+  // between instances is the Redis lock's job, not this set's.
+  readonly #priorityRecomputes = new Set<Promise<void>>();
```
taobojlen, line 986: In general, adding code only for tests is an antipattern. Can we avoid this somehow, e.g. with mocks/spys?

server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
```diff
@@ -190,6 +190,46 @@ export class ClickhouseReportingAnalyticsAdapter implements IReportingAnalyticsA
     return typeof value === 'number' ? value : Number(value);
   }
 
+  async getNumTimesReportedForItems(
```
taobojlen, line 193: What happens when we've re-sorted a queue by # of times reported, and then new reports come in for jobs in the queue? We don't re-sort things in that case, do we? IMO that's a crucial requirement for this feature.

taobojlen: Similarly, when a new job is added to the queue, how do we insert it into the right place in the queue? Does this currently happen in this PR?

taobojlen (changes requested):
Thanks for the updates! As I've gotten to know Coop's data architecture, and the lower-hanging fruit has been fixed, it's become more clear that Coop's architecture makes this feature challenging.

The fundamental challenges are:
- There is currently no efficient way to get the number of reports for every item in a given queue, so this PR has to resort to expensive Clickhouse queries.
- Similarly, there is no good avenue to *update* the order of jobs in a queue as new reports come in. This is crucial! AFAICT, right now the queue's ordering does not update when new reports come in (a blocker, IMO).
- The fact that jobs are stored in BullMQ means that this PR has to introduce a lot of unnecessary complexity that could otherwise be avoided. (Would be fixed by #1200).
- There are also some things we just cannot do with jobs-in-BullMQ, like showing the age of the oldest job in a queue if the queue is sorted by e.g. number of reports, or a custom ordering.

To be clear, these things are not criticisms of this PR, but they make me believe that we need to do some foundational work (like #1200!) before we can really support this. In fact, by cleaning up the data architecture, this feature could become nearly trivial to implement -- that, to me, is a sign of the right architecture!

I've tried to argue against myself here, but the strongest reason I can see to merge this is really the sunk cost fallacy. IMO even if we merged it we should start fixing the above ASAP and then it's just churn.

server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
juanmrad: The batched getNumTimesReportedForItems only runs on sort-mode changes (admin switches a queue from `FIFO` to `NUM_REPORTS`), not on every request. Per-enqueue uses the existing single item `getNumTimesReported` which is already in the codebase. So this is a one-off operation that runs when someone changes queue settings, not something that scales with traffic. 

Long term a column on jobs that aggregates live as reports come in would be better, and that can come with #1200 as we think over a better way to do queue management and storage.

server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
juanmrad: It does update on new reports. Every enqueue (including re-enqueue from a new report) goes through `manualReviewToolService.ts` which fetches the target queue's sort type, gets the current report count, and stamps the BullMQ priority via `addJob` or `updateJobForQueue` -> `changePriority`. So if an item already has a job in the queue and gets reported again, its priority is recomputed with the new count. New jobs get their priority computed at enqueue time the same way.

server/services/manualReviewToolService/modules/QueueOperations.ts
juanmrad: We do show it. The dashboard shows oldest task age for all queues including NUM_REPORTS. The tooltip fallback only appears when oldestJobCreatedAt comes back null, which would only happen if a prioritized queue had more than 10k jobs (the scan cap). For the current scale this works.
I have a follow-up planned to cache the oldest job in Redis with a mini-heap approach so this limitation goes away entirely and we don't need the scan at all.

but will do a separate pr just to keep things scoped here

server/services/manualReviewToolService/modules/PriorityRecomputeLock.ts
juanmrad: We want to ship this feature now, #1200 is a separate effort with its own scope and timeline.

server/services/manualReviewToolService/modules/JobPriority.ts
juanmrad: Agreed, removed it. Call sites now use `?? JobSortType.FIFO` directly. The `DB CHECK` constraint prevents bad values so the runtime validation was redundant.

server/services/manualReviewToolService/modules/JobPriority.test.ts
juanmrad: Agreed, simplified to just explain why it's hard-coded.

client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
juanmrad: We actually do show it when available (see reply on the QueueOperations comment). The follow-up to cache oldest job in Redis will make this reliable at any queue size. For now the scan works up to 10k jobs which covers a good amount of cases. the fast follow will make this more efficient.

server/services/manualReviewToolService/manualReviewToolService.ts
juanmrad: These are integration tests running against real Redis the point is to verify priorities actually got re-stamped. A spy can confirm the sweep was called but can't tell us when the async Redis work is done, which is what we need before asserting. The alternative would be polling Redis state in a loop, which is flakier. This just exposes the in-flight promise so the test can await it cleanly.

client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
taobojlen: OK! IMO from a product perspective it is better to show no data than to show data that might be wrong, but I will let you decide!

PR #720, opened 2026-06-07
chore: switch postgres to official postgres:15.18 image

### LINKED ISSUE #548: Upgrade postgres/pgvector docker image
We are currently using ankane/pgvector:v0.5.1. Should we migrate to the official pgvector/pgvector [images](https://hub.docker.com/r/pgvector/pgvector)?
Additionally, since we do not appear to be utilizing any pgvector extension functionality, should we revert to a base PostgreSQL image? And should we also take this opportunity to upgrade the PostgreSQL version?

serendipty01:
Fixes #548 

Replace the unmaintained third-party `ankane/pgvector:v0.5.1` image with the official `postgres:15.18` image. Same PostgreSQL 15 version, no data migration needed — existing `pg_data` volumes are compatible.

## Summary by CodeRabbit

* **Chores**
  * Updated the PostgreSQL service image to a newer Postgres 15 release.
  * tightened Postgres port exposure so the database is bound to localhost by default.

* **Documentation**
  * Simplified development quick-start to reference Postgres more generally (removed explicit pgvector mention).

### DIFF
--- a/docker-compose.images.yaml
+++ b/docker-compose.images.yaml
@@ -1,6 +1,6 @@
 services:
   postgres:
-    image: ankane/pgvector:v0.5.1
+    image: postgres:15.18
     volumes:
       - pg_data:/var/lib/postgresql/data
     healthcheck:
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -58,7 +58,7 @@ services:
     volumes:
       - pg_data:/var/lib/postgresql/data
     ports:
-      - '5432:5432'
+      - '127.0.0.1:5432:5432'
     healthcheck:
       test: ['CMD', 'pg_isready', '-U', 'postgres', '-d', 'postgres']
       interval: 2s
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -54,8 +54,7 @@ services:
       - postgres
 
   postgres:
-    image: ankane/pgvector:v0.5.1
-    # image: postgres:15.3
+    image: postgres:15.18
     volumes:
       - pg_data:/var/lib/postgresql/data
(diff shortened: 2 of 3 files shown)


ThisIsMissEm:
We probably also want `command: postgres -c listen_addresses=localhost` here, such that postgres doesn't bind to a public port. Details: https://github.com/bluesky-social/ozone/pull/386

julietshen:
this looks good to me, were you able to test it and make sure everything still works?

juanmrad (approved):
LGTM. did some testing locally with postgres over pgvector. the move is sound given we are not using any features from pgvector.

This also moves us with new patches for security locally and adopters are using vanilla postgres over pgvector :)

juanmrad:
@ThisIsMissEm `listen_addresses` controls what addresses Postgres listens on inside the container's network namespace. Setting it to localhost means Postgres only accepts connections arriving at 127.0.0.1 within the container. Instead binding the port to `127.0.0.1` keeps 5432 reachable from your host and from other containers, but stops it being served on the machine's public interfaces

ThisIsMissEm:
@juanmrad all I know is that the -c setup was what @haileyok and others at bluesky ended up on after discovering several databases were getting exposed to the public internet, even with port binding to `127.0.0.1:5432:5432`

haileyok:
so in the ozone version of this, we were using `network_mode: host`, meaning that we actually did want to bind to the host's `127.0.0.1`. here it looks like we're using docker networking, so this PR looks right from my perspective. basically we just want to make sure that docker doesn't punch through iptables, which adding that `127.0.0.1` to `ports` achieves.

PR #721, opened 2026-06-08
ci: use zizmor to lint for CI security issues

taobojlen:

This PR introduces [zizmor](https://docs.zizmor.sh/), a static analysis tool for GitHub Actions.

I suggest we do this because it's an automated and low-friction way to ensure that our CI follows security best practices.

There are two parts to this:
1. I ran `zizmor` locally and fixed its findings. Most of these were automatic fixes, with the exception of:
    - moving `github.base_ref` into `env:` variables to avoid template injection
    - removing the `checks:write` permission from the `apply_pr_checks` workflow, and instead giving this permission to only the `check_api_server` job where it's actually needed.
2. I also added a new Github Actions workflow that runs zizmor in CI. This is a great way to ensure that we don't accidentally weaken our security posture in future PRs.
3. Finally, I removed the caching step from the `mdbook` workflow.
    - This is because zizmor flagged a [cache poisoning](https://docs.zizmor.sh/audits/#cache-poisoning) risk. In practice this couldn't occur, but I figured it was simpler to remove the caching entirely, because:
    - I can see from #14 that we used to build mdbook from source, which was slow. But now that we're downloading the built package, the benefit of also caching it is limited -- it might save a couple of seconds at most.

## Tests

Basically, ensure that CI continues to pass :-)

## (Optional) Rollout Plan

There's a slight challenge with this, in that some workflows (mdbook and publishing packages/images) only run on `main`. I can trigger these manually via `workflow_dispatch` once they are merged to `main`. But, they have side effects (like publishing a package!).

So I think the best approach there is just to be ready for potential issues the next time we run these, so we can fix them quickly.

## Summary by CodeRabbit

* **Chores**
  * Integrated automated security vulnerability detection into the CI/CD pipeline.
  * Configured dependency update cooldown periods to reduce update frequency and improve stability.
  * Enhanced CI/CD security practices with improved credential handling.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: ['**']
 
+concurrency:
+  group: zizmor-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions: {}
 
 jobs:
@@ -24,3 +28,11 @@ jobs:
 
       - name: Run zizmor
         uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
+        with:
+          version: "1.25.2"
+          # only upload to github advanced security on `main`.
+          # on PRs from forks (which do not have permission to push SARIF results)
+          # we surface findings as PR annotations instead.
+          advanced-security: ${{ github.event_name == 'push' }}
+          annotations: ${{ github.event_name == 'pull_request' }}
+
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,26 @@
+name: GitHub Actions security analysis
+
+on:
+  push:
+    branches: ['main']
+  pull_request:
+    branches: ['**']
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Run zizmor
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write  # to upload SARIF results to code scanning
+      contents: read
+      actions: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,7 +1,0 @@
-# zizmor configuration
-# https://docs.zizmor.sh/configuration/
-
-rules:
-  dependabot-cooldown:
-    ignore:
-      - dependabot.yml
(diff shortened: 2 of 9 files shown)


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,26 @@
...
+      contents: read
+      actions: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
```
juanmrad, line 30:
```suggestion
      - name: Run zizmor
        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
        with:
          version: "1.25.2"
          advanced-security: ${{ github.event_name == 'push' }}
          annotations: ${{ github.event_name == 'pull_request' }}

```

Pin the zizmor binary, not just the action. The action is pinned to a SHA, but its version input defaults to latest, so each run installs whatever zizmor release exists at that moment.

Also update such that we only upload SARIF on push, use annotations on PRs. Gating SARIF upload to `push` (where the token is privileged) and switching PRs to inline annotations keeps the Security tab populated from main while still surfacing findings on PRs

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  push:
+    branches: ['main']
+  pull_request:
+    branches: ['**']
+
+permissions: {}
+
+jobs:
```
juanmrad, line 15:
```suggestion
permissions: {}

concurrency:
  group: zizmor-${{ github.ref }}
  cancel-in-progress: true

jobs:
```

Add concurrency to cancel superseded runs. Saves a bit of Actions time on rapid pushes.

juanmrad (changes requested):

.github/workflows/zizmor.yml
taobojlen: great suggestions, thank you! fixed :)

taobojlen:
zizmor failed in CI because we don't use dependabot cooldowns. i added a `.github/zizmor.yml` file to ignore this for now.

.github/zizmor.yml
```diff
@@ -0,0 +1,7 @@
+# zizmor configuration
```
juanmrad: we do not run dependabot consistently, and we also don't do merges of dependabot without proper review. Technically right now we are not merging any dependabot pr until we go over package updates with care and manual testing. 

so to ask: could we resolve this finding in `dependabot.yml` instead of suppressing it here? Adding a cooldown directly means we fix the actual finding rather than mute it, and we can then delete this `zizmor.yml` entirely (no standing exception to maintain).

Since cooldown doesn't apply to security updates and we batch-review Dependabot PRs manually anyway, a short default is essentially free and even trims PR churn:

```
    schedule:
      interval: 'weekly'
    cooldown:
      default-days: 7
```

taobojlen: Sure thing! That's done now.

juanmrad (approved):

PR #726, opened 2026-06-08
client: stop rendering User Score in MRT and Investigation (part 1 of #596)

### LINKED ISSUE #596: Remove user scoring in UI
- this exists as well but is deprecated, which is what the screenshot above shows.
- this was a much simpler but invisible calculation for users with a score 1-5 (with no explainability on how the score is computed)
- new users start at 5 (initialUserScore = 5) and degrade as they accumulate penalized actions
- calculated based on Actions
- the score would then just be manually compared in rules vs trigger any cascading actions
- lifetime computed

### How User Score works
The actions table has a penalty column of type UserPenaltySeverity (enum: NONE/LOW/MEDIUM/HIGH/SEVERE). Same column exists on policies. Both default to NONE per the schema. 

Theoretically, when an org admin creates an action, they pick its severity from that 5-value enum. HOWEVER, this was never built in the UI so right now, all actions default to a NONE severity.

Scores were calculated per-severity magnitudes (exponential 3ⁿ ladder)

NONE = 0
LOW = 1
MEDIUM = 3
HIGH = 9
SEVERE = 27

penalty = action.penalty_magnitude + policy.penalty_magnitude

With one carve-out: if the action's penalty is NONE, the combined penalty is 0 regardless of the policy — i.e., non-punitive actions like "Send to Moderation" or "Restore Content" never contribute to the score even if they reference a SEVERE policy. So actions are marked punitive purely by the action's own penalty field being anything other than NONE.

We probably want to finish User Strikes and enable in the UI and remove the User Score

julietshen:
## Summary

User Strikes covers the workflow User Score was built for, with explainable per-policy scoring. Stop rendering User Score in the Review Console user panel and the Investigation page.

## Scope (UI only)

This is part 1 of two PRs that close #596. Backend stays in place since removing it would break adopter API consumers and any existing rules that reference the `USER_SCORE` signal.

Removed:
- `userScore` selection on `NcmecManualReviewJobPayload`, `ContentManualReviewJobPayload`, `UserManualReviewJobPayload`, `ContentAppealManualReviewJobPayload` in `jobFragment.ts`.
- `userScore` prop on `ManualReviewJobPrimaryUserComponent` and the corresponding `userScore` argument to `convertRelatedItemToFieldData`.
- `userScore` selection on the `UserItem` GraphQL inline fragment in `ItemInvestigation.tsx` plus the prop pass-through.
- Regenerated `client/src/graphql/generated.ts`.

Kept (follow-up):
- `GET/POST /api/v1/user-scores` route and `docs/api/user-scores.md`.
- `USER_SCORE` signal type in `signalsService` plus the rule-form switch cases / TODOs that reference it.
- `userScore: Int` on the GraphQL `UserItem` type and the MRT payload types.
- `user_statistics_service.user_scores` table and adjacent service code.

## Test plan

- [x] `tsc --noEmit` clean (server + client).
- [ ] Manual smoke against Platform A:
  - User panel in MRT Review Console no longer shows the "User Score" field.
  - Investigation page user item no longer shows a User Score row.
  - Existing rules that reference the `USER_SCORE` signal still load in the rule form (deferred to follow-up; current behaviour preserved).

### DIFF
--- a/client/src/graphql/generated.ts
+++ b/client/src/graphql/generated.ts
@@ -7030,7 +7030,6 @@ export type GQLGetItemsWithIdQuery = {
           readonly data: JsonObject;
           readonly submissionId: string;
           readonly submissionTime?: Date | string | null;
-          readonly userScore: number;
           readonly type: {
             readonly __typename: 'UserItemType';
             readonly id: string;
@@ -8160,7 +8159,6 @@ export type GQLGetDecidedJobFromJobIdQuery = {
       readonly payload:
         | {
             readonly __typename: 'ContentAppealManualReviewJobPayload';
-            readonly userScore?: number | null;
             readonly appealReason?: string | null;
             readonly appealId: string;
             readonly actionsTaken: ReadonlyArray<string>;
@@ -8310,7 +8308,6 @@ export type GQLGetDecidedJobFromJobIdQuery = {
           }
         | {
             readonly __typename: 'ContentManualReviewJobPayload';
-            readonly userScore?: number | null;
             readonly reportHistory: ReadonlyArray<{
               readonly __typename: 'ReportHistoryEntry';
               readonly policyId?: string | null;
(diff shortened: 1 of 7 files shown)


julietshen:
Tested locally and user score does not show up in the UI anymore

juanmrad (approved):
lgtm

PR #730, opened 2026-06-09
Show related item's name instead of its raw ID and fix duplicate relatedItems

### LINKED ISSUE #716: Associated Account/User information duplicated
### What happened

The information for associated accounts/users is duplicated when using the partial items API.

### Version/commit

_No response_

### Affected browser(s)

_No response_

### Relevant logs/errors

```shell

```

### Anything else

Screenshots are sent to @juanmrad in direct messages on Discord.

### Screenshots

_No response_

### Checklist

juanmrad:
Fixes #716

Two small fixes for how related items show up in MRT.

1. Related items rendered inline only resolved a title when the item was a User. Meaning every other type (Content, Thread) fell through to showing the raw ID. The data was available if the Item Type is setup with the property roles, we just weren't reading the `displayName` role for non-user types. Now we resolve it for all item kinds, so e.g. a related "Item" shows its name instead of raw ids. `clwuq0a62...`.

2. The reference-to-fields helper rendered denormalized keys carried on the related item reference, which could duplicate fields already shown from the item's typed schema. It now dedupes those inlined keys against the schema-rendered field names (case/whitespace-insensitive), so the same field isn't shown twice.

## Summary by CodeRabbit

* **New Features**
  * Manual review job submissions now support displaying and submitting related items of multiple types (users, content, threads).
  * Related items display human-readable names and profile images appropriate to their type.

* **Improvements**
  * Enhanced field deduplication for related items to prevent duplicate information display.

* **Tests**
  * Added comprehensive test coverage for related item field data conversion and name deduplication.

### DIFF
--- a/client/src/graphql/generated.ts
+++ b/client/src/graphql/generated.ts
@@ -16558,6 +16558,10 @@ export type GQLGetRelatedItemsQuery = {
               readonly valueScalarType: GQLScalarType;
             } | null;
           }>;
+          readonly schemaFieldRoles: {
+            readonly __typename: 'ContentSchemaFieldRoles';
+            readonly displayName?: string | null;
+          };
         };
       }
     | {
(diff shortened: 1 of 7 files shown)


julietshen (approved):

julietshen:
lgtm

Issue #485, opened 2026-05-15
Bootstrap Playwright E2E suite for moderator-critical UI flows


julietshen:
Stand up a Playwright (or comparable) E2E test suite covering the moderator-critical UI flows in the Coop dashboard. Run nightly and as a release gate, not per-PR.

## Motivation

Coop currently has ~19 client tests, all unit-level. There is no automated coverage of the UI flows a moderator actually uses. A field-rendering regression (e.g. AUDIO field stops rendering an inline player in MRT jobs) would ship.

## In-scope flows for the initial suite

- Login + session
- Create an item type with mixed field types (STRING, AUDIO, IMAGE, USER_ID)
- Create a rule and route to a manual review queue
- MRT job: render a job with text + image + audio + video fields; play the audio; apply a decision; verify it lands in the dashboard
- View item details and signal results

## Out of scope (for the bootstrap)

- Settings pages beyond item types and rules
- NCMEC review UI (consider for a follow-up)
- Integration plugins UI

## Related

- Parent: #288

taobojlen:
hi @julietshen! can i take this one? 

> Run nightly and as a release gate, not per-PR.

FWIW i would recommend running these tests per-PR by default unless there's a strong reason not to. better to keep CI green on main up front rather than having to open lots of fix PRs when things break!

julietshen:
that sounds good for me! cc @cassidyjames @juanmrad

PR #732, opened 2026-06-09
test: isolate server tests via transaction rollback

taobojlen:

(I realize that this is a large PR, but mostly whitespace! I decided that splitting it up would only make it harder to review, but happy to talk about it! And as always, all my PRs are just proposals, we can decide to go in another direction).

As I started looking into #485, I noticed that the existing tests in `server/` require developers to manually clean up any data they create during the test. This is fragile -- it's easy to miss something! In fact, we already have tests that leave data behind, such that if you try running the test suite twice against the same DB, it fails the second time around.

So, this PR is the first step towards #485.

In general, it's much easier to write tests if they each run in their own database transaction (that is then rolled back at the end). This means that tests are fully independent, and leave the database in a clean state. It's also how many web frameworks like Django work by default.

This PR updates most existing tests to use a new harness that does just this. This is a step towards a world where:
- each test creates just the data it needs, and cannot implicitly depend on data created elsewhere (explicit > implicit)
- tests automatically perform their own cleanup (so there's no risk of DB cross-contamination)
- because tests are independent by default, we can easily run tests in parallel (i.e. faster!)

## Tests

You should be able to see that tests continue to pass. So the important thing here when reviewing is to check and ensure that the tests are still testing the same things as before.

Locally, this does not meaningfully change how long it takes to run the test suite (155.2s on `main`, 159s on this branch).

I recommend toggling on GitHub's "hide whitespace" in the diff:

CleanShot 2026-06-09 at 13 32 35

## Release Notes

* **Tests**
  * Improved test reliability by migrating many suites to a transactional database harness, reducing manual cleanup and tightening isolation between org/user scenarios.
  * Added stronger coverage for transaction rollback/isolation and safeguards against unsupported transaction-control statements.
* **New Features**
  * Added a transactional test DB harness utility for consistent per-test database state.
* **Chores**
  * Centralized PostgreSQL connection parameter construction and reused it for master connections.
  * Enabled optional SSL configuration when database SSL is enabled.

### DIFF
--- a/server/iocContainer/index.ts
+++ b/server/iocContainer/index.ts
@@ -441,6 +441,17 @@ export interface Dependencies {
 // treatment that TS gives to classes with private fields; see https://stackoverflow.com/questions/55281162/can-i-force-the-typescript-compiler-to-use-nominal-typing)
 export type PublicInterface<T extends object> = { [K in keyof T]: T[K] };
 
+export function getPgConnectionParams(): pg.ClientConfig {
+  return {
+    user: process.env.DATABASE_USER ?? 'postgres',
+    database: process.env.DATABASE_NAME ?? 'development',
+    password: safeGetEnvVar('DATABASE_PASSWORD'),
+    port: parseInt(process.env.DATABASE_PORT ?? '5432'),
+    host: safeGetEnvVar('DATABASE_HOST'),
+    ssl: isEnvTrue('DATABASE_SSL') ? { rejectUnauthorized: false } : undefined,
+  };
+}
+
 /**
  * A function for creating our service container, configured for production.
  * Services can be rebound in other contexts (namely, tests) as needed.
@@ -514,15 +525,10 @@ export default async function getBottle() {
   // that would defeat the ability of safeGetEnvVar to alert us in prod if a
   // worker that needs these vars is run without them.
   const getPgMasterConnectionInfo = () => ({
-    user: process.env.DATABASE_USER ?? 'postgres',
-    database: process.env.DATABASE_NAME ?? 'development',
-    password: safeGetEnvVar('DATABASE_PASSWORD'),
-    port: parseInt(process.env.DATABASE_PORT ?? '5432'),
-    host: safeGetEnvVar('DATABASE_HOST'),
+    ...getPgConnectionParams(),
     max: parseInt(process.env.DATABASE_POOL_MAX ?? '30'),
     application_name:
       getEnvVarOrWarn('OTEL_SERVICE_NAME') ?? 'unknown-coop-service',
-    ssl: isEnvTrue('DATABASE_SSL') ? { rejectUnauthorized: false } : undefined,
     ...getPgPoolTuning(),
   });
 
--- a/server/routes/reporting/ReportingRoutes.test.ts
+++ b/server/routes/reporting/ReportingRoutes.test.ts
@@ -1,38 +1,22 @@
 /* eslint-disable max-lines */
 import { faker } from '@faker-js/faker';
+import { uid } from 'uid';
 
 import createOrg from '../../test/fixtureHelpers/createOrg.js';
 import createUser from '../../test/fixtureHelpers/createUser.js';
-import { makeMockedServer } from '../../test/setupMockedServer.js';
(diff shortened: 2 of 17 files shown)


server/routes/reporting/ReportingRoutes.test.ts
```diff
@@ -1,38 +1,22 @@
...
 import { faker } from '@faker-js/faker';
+import { uid } from 'uid';
 
 import createOrg from '../../test/fixtureHelpers/createOrg.js';
 import createUser from '../../test/fixtureHelpers/createUser.js';
-import { makeMockedServer } from '../../test/setupMockedServer.js';
+import { makeTransactionalTestWithFixture } from '../../test/harness/transactionalTest.js';
 
 describe('POST Report', () => {
-  const orgId = '2a4634c81d5',
```
taobojlen, line 9: There was no reason to hard-code these IDs so I removed it.

server/iocContainer/index.ts
```diff
@@ -441,6 +441,23 @@ export interface Dependencies {
 // treatment that TS gives to classes with private fields; see https://stackoverflow.com/questions/55281162/can-i-force-the-typescript-compiler-to-use-nominal-typing)
 export type PublicInterface<T extends object> = { [K in keyof T]: T[K] };
 
+/**
+ * Postgres connection identity (host, port, credentials, database, SSL) from
+ * the `DATABASE_*` env vars. Shared by the app's pools (via
+ * `getPgMasterConnectionInfo`) and the test harness so both connect to the same
+ * place; pool/client tuning is layered on separately.
+ */
+export function getPgConnectionParams(): pg.ClientConfig {
```
taobojlen, line 444: Extracting this to a function so I can use it in the new transaction-test harness, too.

taobojlen:
Fixing coderabbit suggestions.

server/test/harness/transactionalPgPool.ts
```diff
@@ -0,0 +1,131 @@
...
+  // Shared by both the per-connection facade (`connect().query`) and the
+  // pool-level `query` (used by consumers like the express-session store).
+  const runQuery = async (
+    textOrConfig: string | pg.QueryConfig,
+    values?: unknown[],
+  ): Promise<unknown> => {
+    const text =
+      typeof textOrConfig === 'string' ? textOrConfig : textOrConfig.text;
+    const normalized =
+      typeof text === 'string' ? text.trim().toLowerCase() : '';
```
juanmrad: may be worth removing `;` just in case

server/test/harness/transactionalPgPool.ts
```diff
@@ -0,0 +1,131 @@
...
+        throw new Error('COMMIT without a matching application transaction');
+      }
+      const name = savepointName(savepointDepth);
+      const result = await client.query(`RELEASE SAVEPOINT ${name}`);
+      // Only drop the depth once the SQL succeeds, so a failed call can't
+      // leave the savepoint stack out of sync.
+      savepointDepth -= 1;
+      return result;
+    }
+    if (normalized === 'rollback') {
```
juanmrad, line 129: may be worth doing a switch case instead so after the known cases, detect unhandled transaction-control verbs and throw, rather than silently passing them to the real connection. 

If any code or library ever emits `END` or `ABORT`, it would pass through unrewritten and commit/abort the outer test transaction silently defeating isolation.

juanmrad (approved):
mostly nit. otherwise LGTM

server/test/harness/transactionalTest.ts
```diff
@@ -0,0 +1,39 @@
...
+    return {
+      deps: server.deps,
+      request: server.request,
+      ...fixtures,
+      async cleanup() {
+        await server.rollback();
+        await server.shutdown();
+      },
+    };
+  });
```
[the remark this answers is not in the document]
taobojlen: Not an issue. If `makeFixtures` fails, something is wrong with the test suite which needs to be fixed first.

server/test/harness/transactionalPgPool.ts
taobojlen: great catch, thank you!

PR #733, opened 2026-06-09
ci: standardize on dorny/paths-filter

taobojlen:

This is just a small cleanup. I noticed that in CI, we were using both `dorny/paths-filter` and also our own `check_if_directory_changed.yaml` job workflow to do the same thing.

This PR standardizes on `dorny/paths-filter` -- less stuff to maintain for us!

## Summary by CodeRabbit

* **Chores**
  * Improved CI/CD workflow change detection by implementing a more efficient path-filtering mechanism.
  * Removed obsolete workflow file and consolidated change-detection logic.


juanmrad (approved):
lgtm

PR #741, opened 2026-06-09
Fix release of docker image for tagged release and allow manual release

juanmrad:

Fix release of tagged versions and allow manual run of workflow to backfill the v1 version.

## Summary by CodeRabbit

* **Chores**
  * Added an optional version input with validation to the Docker publish workflow for more predictable releases.
  * Enabled per-ref concurrency to avoid redundant simultaneous builds while allowing independent refs to proceed.
  * Unified Docker tag generation at workflow level so all image jobs produce consistent tags.
  * Refined build cache usage to improve cache reuse across server, client, and migration image builds.

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'Optional version tag to apply (e.g. 1.0.0). Also updates :latest. Leave blank for an SHA-only build.'
+        description: 'Optional semver version tag to publish (e.g. 1.0.0), for backfilling a release. Leave blank for an SHA-only build; does not move :latest.'
         required: false
         type: string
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker.yaml
```diff
@@ -13,12 +13,32 @@ on:
...
 
+# Avoid redundant concurrent builds for the same ref (expensive multi-arch).
+# Keyed per-ref so releases (tag refs) never cancel main builds.
+concurrency:
+  group: publish-docker-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   REGISTRY: ghcr.io
+  # Shared tagging policy for every image. `latest` tracks the latest build on main
```
taobojlen: fwiw i do agree with you that `latest` usually points to the latest tagged release. i think you may be able to do that using the `flavor` setting from `docker/metadata-action`: https://github.com/docker/metadata-action#flavor-input

and then e.g. `edge` could point to whatever's on `main`.

juanmrad: Makes sense I am ok with using edge for main and latest for release latest :)

PR #745, opened 2026-06-11
Only run License Check when we should check licenses

juanmrad:

We are running CI to check licenses on each commit on all PRs, but we should really only run this when there is a new dependency change. as such we should filter by the package-lock.json changes instead. Once we move to pnpm we can update to the appropriate lock file. 

## Summary by CodeRabbit

* **Chores**
  * Optimized license check workflow to run only when dependency files are modified, reducing unnecessary CI/CD executions.


julietshen (approved):
lgtm

PR #747, opened 2026-06-11
Set redis user and pw in no-cluster mode

dom-notion:
The redis username/pw wasn't being passed when in a single-node configuration, which is ok for local dev but doesn't work if using a single node Elasticache for example. This passes the username/pw if present in the env vars.

## Tests
Existing automated tests. Also deployed this to our instance and it works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Redis connection initialization to properly handle authenticated Redis instances. The application now correctly passes authentication credentials only when configured, preventing connection failures and ensuring reliable command execution with auth-enabled Redis deployments.


julietshen (approved):
lgtm

PR #755, opened 2026-06-11
[Feature] Include creator in action webhook callback

juanmrad:

Action webhook callbacks (`CUSTOM_ACTION`) now include a creator field identifying the user the action concerns, so consumers don't have to look it up themselves.

Previously the body only had `item.id`, which for content (e.g. a message or post that received a strike) is the content's ID not the author's. Consumers had to do a separate backend lookup to find who actually authored it. 

The change adds the creator information when available.

## Summary by CodeRabbit

* **New Features**
  * Custom-action webhooks may include top-level creator, decisionReason, and userStrikeCount when resolvable; creator falls back to content creator and is omitted if unresolved.

* **Documentation**
  * API docs updated with presence rules and a “Custom object” subsection documenting reason → decisionReason mapping and reportHistory structure.

* **Tests**
  * Added tests for creator resolution, decisionReason emission, and userStrikeCount inclusion/omission.


ThatKoffe (approved):
tested & reviewed

PR #760, opened 2026-06-12
client: set up knip and remove unused dependencies

taobojlen:

Like https://github.com/roostorg/coop/pull/734, but for the `client/` directory.

This PR adds knip, and removes unused dependencies from the client. We can continue using knip ad-hoc to remove unused code (there's quite a lot to remove from `client/` actually!).

Remove dependencies with no references anywhere in the source tree:
- framer-motion, jsonpointer, papaparse, react-markdown, react-router-hash-link, validator
- @eslint/js (not imported by the flat config; bundled by eslint itself)
- orphaned/obsolete types: @types/papaparse, @types/react-router-hash-link, @types/validator (their runtime packages were removed) and @types/recharts (recharts ships its own type definitions)

## Tests

I am primarily relying on Knip's logic and our test suite here. I also ran this locally, signed in to Coop, clicked around, reviewed some jobs -- everything seems to work.


juanmrad (approved):
LGTM

PR #763, opened 2026-06-12
Allow ignoring a job when "require decision reason" is enabled

### LINKED ISSUE #757: Unable to ignore job when "require decision reason" is enabled
### What happened

I am unable to ignore a job when the "require decision reason" is enabled

Image

The button remains disabled.

Image

### Version/commit

_No response_

### Affected browser(s)

_No response_

### Relevant logs/errors

```shell

```

### Anything else

_No response_

### Screenshots

_No response_

### Checklist

taobojlen:
## Summary

Fixes #757. Before, when an org had **require decision reason** enabled, reviewers could not Ignore a job without giving a reason. As of this PR, the Ignore action is excluded from that requirement.

There is a risk here: there may be users who *want* to require a reason even for Ignores. So this change may break that hypothetical workflow.

## Testing

I wrote a test for this and also tested it manually on my machine.

## Summary by CodeRabbit

* **Bug Fixes**
  * Decision reason requirements are now correctly waived for ignore-only submissions in manual review jobs, allowing users to submit without providing a reason when only ignoring items.

* **Tests**
  * Added test coverage to validate decision reason enforcement behavior for ignore-only decisions.

### DIFF
--- a/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
@@ -679,10 +679,6 @@ function ManualReviewJobReviewImpl(props: {
       }
     }
 
-    // If the org requires a decision reason and none has been provided, block
-    // submission — unless the only decision is Ignore. Ignoring a job is a
-    // dismissal, not a reasoned moderation action, so it's exempt from the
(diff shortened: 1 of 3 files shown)


taobojlen:
Based on discussion in Discord: we won't merge this as it would break existing workflows, if any users *want* to require a decision reason for ignored jobs.

Instead, we'll create two separate config options:
- Require decision reason (for violating jobs)
- Require decision reason (when ignoring)

PR #764, opened 2026-06-12
Copy Lua assets when starting the dev server

taobojlen:
## Summary

This is a small devx improvement.

On a fresh clone of the repo, if you run `npm run up && npm run server:start`, you'll get an error because the server doesn't have some required Lua files where it expects.

This fixes that!

## Testing

- `npm start` from a clean repo no longer throws ENOENT and boots normally.


juanmrad (approved):
I was meaning to get to this lol. thanks!

PR #774, opened 2026-06-13
[Bug] Fix duplicate recent decisions on select latest

juanmrad:

When users selects a recent decision on the recent decision pages we duplicate the last decision when selected. as we did not filter the currently selected decision. this pr fixes that

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate decisions appearing in the recent decisions table. The system now checks if a decision ID already exists before adding it to the list, preventing duplicate entries from being displayed to users. Users will now see only unique, clean decision records in the recent decisions view without any duplicates.


julietshen (approved):
lgtm

Issue #768, opened 2026-06-12
"Unblur all media" toggle doesn't affect iframe content


cassidyjames:
The "Unblur all media" option in the job review view only applies to `<img>` and `<video>` elements but has no effect on content displayed in iframes (e.g. atproto posts via the content proxy).

I would imagine we want that toggle to affect media regardless of whether it's displayed directly in Coop or in an iframe preview.

taobojlen:
I am not sure that this is technically possible without using a content proxy -- AFAIK a parent document cannot run arbitrary JS against the contents of an iframe. At least not if they are on different origins.

taobojlen:
@julietshen continuing from https://github.com/roostorg/coop/pull/777#issuecomment-4720895196...

If blur/grayscale/etc. inside iframes is important, then we'd probably want to re-introduce the content proxy and default to routing iframes through that. I can't think of a way to do it that *doesn't* involve proxying the content.

But one question is why people use iframes at all. If it's mostly because Coop doesn't allow them to display the content the way they want, then that may be a more valuable thing to fix. I don't know!

julietshen:
let's leave it as is for now and clarify in the documentation around wellness features. Once we hear from adopters on why they're using iframes we can make a decision from there

cassidyjames:
Looking at this again, I think I _had_ implemented a content proxy in my atproto-demo branch with wellness integration, but wasn't seeing the unblur option actually work. I'll need to double-check that; it's possible it was just a bug in my content proxy.

taobojlen:
This is what Astra had to say about blurring in the content proxy (before we removed it):

> It included blur/grayscale support for images inside proxied pages, rather than just filtering the whole iframe:
> - Injected CSS supported 100% grayscale and blur radii of 0, 4, 8, 12, 16, 24, and 40 px, including combined grayscale + blur.
> - An injected script listened for customControl messages from the parent page, then applied filter classes to every <img>. The blur radius was blur × 4.
> - Hovering removed blur while retaining grayscale.
> 
> One caveat: the checked-in injected script had a syntax error: the addEventListener(...) call ended with }; instead of });. I confirmed the extracted script fails to parse. So the feature was documented and implemented in intent, but that exact repository version would not actually activate the filters without fixing the script.

So that may have been the bug you hit.

PR #777, opened 2026-06-15
manual review: fix iframe rendering when no content proxy is set

taobojlen:
## What

There was a bug in iframe rendering. When `VITE_CONTENT_PROXY_URL` was unset or empty, then if we showed an iframe, it would render Coop itself and not the actual URL it was intended to render! This is because the code fell back to rendering `localhost:4000` in dev or `window.location.origin` in prod, instead of the correct URL.

This PR fixes that.

## Notes

Note that blur/grayscale/translation only work when a content proxy is configured (see also https://github.com/roostorg/coop/issues/768).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of content proxy configuration to correctly disable translation features when proxy is not configured.
  * Fixed iframe content display logic to properly route content through the proxy when configured and directly when disabled.

### DIFF
--- a/client/src/webpages/dashboard/mrt/manual_review_job/IframeContentDisplayComponent.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/IframeContentDisplayComponent.tsx
@@ -14,9 +14,11 @@ export default function IframeContentDisplayComponent(props: {
 }) {
   const { contentUrl } = props;
 
+  // An empty or unset proxy URL means the content URL is rendered directly in
+  // the iframe rather than routed through a content proxy.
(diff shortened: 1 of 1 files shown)


julietshen (approved):
LGTM. @cassidyjames do we need to update docs to reflect that the blur features won't work without a content proxy?

taobojlen:
@julietshen AFAICT the content proxy was removed from this project, so right now there's actually no way to get blur/grayscale/translation working in iframes.

julietshen:
that's right, I remember now. What are our options?

taobojlen:
@julietshen let's continue in https://github.com/roostorg/coop/issues/768!

PR #778, opened 2026-06-15
(auth) Invalidate sessions on password change

taobojlen:
## Summary

This PR fixes **GHSA-g5xq-67g7-36r2**.

As of this PR:
- when a user changes their password, all *other* sessions are invalidated, and
- when a user *resets* their password (via a password reset email, while logged out), then *all* sessions are invalidated.

## Testing

- I added tests for this, and verified that it works manually.

## Summary by CodeRabbit

* **New Features**
  * Updating your password now keeps your current session active while logging out all other active sessions.
  * Password reset now invalidates all your active sessions for improved security consistency.
* **Tests**
  * Added/expanded Jest coverage to verify session invalidation behavior for password changes and password resets, including transactional execution paths.

### DIFF
--- a/server/services/userManagementService/sessionPersistence.ts
+++ b/server/services/userManagementService/sessionPersistence.ts
@@ -1,0 +1,28 @@
+import { sql, type Kysely } from 'kysely';
+
+/**
+ * Delete connect-pg-simple session rows for a user, to force re-authentication
+ * after a password change (GHSA-g5xq-67g7-36r2). Passport stores the user id at
+ * `sess -> 'passport' ->> 'user'` (see `passport.serializeUser` in
+ * `server/api.ts`, which serializes `user.id`).
+ *
+ * Pass `exceptSid` to keep one session alive — e.g. the caller's own session on
+ * a self-service password change, so they aren't logged out mid-request.
+ */
+export async function deleteSessionsForUser(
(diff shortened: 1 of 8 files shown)


server/services/userManagementService/sessionPersistence.ts
```diff
@@ -0,0 +1,28 @@
...
+ * `server/api.ts`, which serializes `user.id`).
+ *
+ * Pass `exceptSid` to keep one session alive — e.g. the caller's own session on
+ * a self-service password change, so they aren't logged out mid-request.
+ */
+export async function deleteSessionsForUser(
+  // `Kysely<any>` because the session table is managed by connect-pg-simple and
+  // is intentionally absent from our typed schema; injected instances are
+  // themselves `Kysely<any>`.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
juanmrad, line 16: Something here for future us is that if we get reports of slowness here. it is because `public.session` has only 2 indexes in `sid` and `expire`. 

Meaning this operation does a full table scan. Given this only has rows for each user that is fine. we could mitigate creating an index like:
```
CREATE INDEX idx_session_passport_user
    ON public.session ((sess -> 'passport' ->> 'user'));
```

juanmrad (approved):
LGTM

PR #780, opened 2026-06-15
(manual review) Split require-decision-reason into action vs ignore settings (#757)

### LINKED ISSUE #757: Unable to ignore job when "require decision reason" is enabled

taobojlen:
## Why

Closes #757.

Before this PR: we have a "Require decision reason" setting which applies both when actioning a job, and also when ignoring it.

After this PR: there are now two settings, "Require decision reason when actioning jobs" and "Require decision reason when ignoring jobs".

I could use some feedback on the copy here -- I'm not quite sure how to make it read easily!

Video demo: https://cleanshot.com/share/pXc8jbmg

Screenshot 2026-06-15 at 12-08-46 Settings

## Deployment considerations

I'm changing the names of the existing database column. I think it's better to be explicit. However, this means that when this new version is deployed, users might momentarily see errors until the deployment succeeds.

If this was a SaaS app, I'd do this more gradually to avoid those transient errors -- but I don't think that's worth it when this is mostly self-hosted.

## Summary by CodeRabbit

* **New Features**
  * Added a separate moderator requirement to require a decision reason specifically for **IGNORE** decisions, alongside the existing non-ignore requirement.
* **Settings**
  * Updated the Moderator Requirements UI with the “Require Decision Reason (when ignoring jobs)” toggle and ensured save/toggle state stays in sync.
* **Bug Fixes**
  * Decision submission now enforces the decision-reason rule using the correct setting for IGNORE vs non-IGNORE decisions.
* **Tests**
  * Expanded automated coverage to validate missing/required decision reasons for IGNORE decisions.
* **Chores**
  * Added a database migration to split the prior single setting into two action/ignore flags.

### DIFF
--- a/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
@@ -147,6 +147,7 @@ gql`
       hasNCMECReportingEnabled
       requiresPolicyForDecisionsInMrt
       requiresDecisionReasonInMrt
+      requiresDecisionReasonOnIgnoreInMrt
       allowMultiplePoliciesPerAction
       hideSkipButtonForNonAdmins
     }
@@ -680,11 +681,16 @@ function ManualReviewJobReviewImpl(props: {
     }
 
     // If the org requires a decision reason, and no decision reason has been
-    // provided, return false
-    if (
-      data?.myOrg?.requiresDecisionReasonInMrt &&
-      !isNonEmptyString(decisionReason)
-    ) {
+    // provided, return false. The requirement is split (see #757): ignoring a
+    // job (a decision made up solely of the built-in IGNORE action) is governed
+    // by its own setting, separate from acting on a violating job.
+    const isIgnoreDecision = selectedPrimaryActions.every(
+      (it) => 'type' in it.action && it.action.type === 'IGNORE',
+    );
+    const reasonRequired = isIgnoreDecision
+      ? data?.myOrg?.requiresDecisionReasonOnIgnoreInMrt
+      : data?.myOrg?.requiresDecisionReasonInMrt;
+    if (reasonRequired && !isNonEmptyString(decisionReason)) {
       return false;
     }
 
@@ -1669,14 +1675,13 @@ function ManualReviewJobReviewImpl(props: {
                 <div className="self-start my-2 text-lg font-bold">Policy</div>
                 {policiesSection}
               </div>
-              {org.requiresDecisionReasonInMrt ? (
-                <div className="flex flex-col mb-4">
-                  <div className="self-start my-2 text-lg font-bold">
-                    Reason
-                  </div>
-                  {decisionReasonSection}
-                </div>
-              ) : null}
+              {/* The reason field is always shown; whether it's required
+                  depends on the org's require-decision-reason settings and the
+                  decision type (see canBeSubmitted). */}
+              <div className="flex flex-col mb-4">
+                <div className="self-start my-2 text-lg font-bold">Reason</div>
+                {decisionReasonSection}
+              </div>
               <ManualReviewJobEnqueuedRelatedActions
                 actionsData={selectedRelatedActions.map((action) => ({
                   // NB: We don't include any iconUrl or otherImageUrls here yet, since we're still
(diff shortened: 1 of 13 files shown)


client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -1669,14 +1675,13 @@ function ManualReviewJobReviewImpl(props: {
                 <div className="self-start my-2 text-lg font-bold">Policy</div>
                 {policiesSection}
               </div>
-              {org.requiresDecisionReasonInMrt ? (
```
taobojlen, line 1672: Slightly unsure about this part. Previously the reason field was only shown in MRT when a reason was *required*. Now, when should we show the reason field?
1. If it's required for either actions/ignores
2. Always

I chose option 2 here, so the Reason field is always shown.

client/src/webpages/settings/tabs/ReviewConsoleTab.tsx
```diff
@@ -97,6 +111,16 @@ export default function ReviewConsoleTab() {
     }
   };
 
+  // The "Require Decision Reason" master toggle is derived from the two
+  // underlying flags: it's on when a reason is required for either action or
+  // ignore decisions. Turning it on defaults to requiring a reason for actions
+  // only; turning it off clears both.
+  const requireReasonAny = requireReason || requireReasonOnIgnore;
+  const handleRequireReasonToggle = (checked: boolean) => {
```
taobojlen, line 119: Here's a video demo of this behavior: https://cleanshot.com/share/bGzPPr0b

julietshen:
copying discussion from Discord:

if we have the ignore webhook url, do we also need to split this setting into action / ignore? from a UX perspective can we keep it at a reason level toggle with a checkbox or setting indented on the next line to apply for All actions including ignore?


and Tao's updated UI:
https://cleanshot.com/share/bGzPPr0b



note for later: I don't love "ignore" since there can be assumptions around it. maybe in future UI we can explore something like "No violation found" or something less wordy

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
julietshen: can you add a screenshot of what it looks like when it's not required?

julietshen (approved):
lgtm, pending screenshot of what it looks like to always show the reason. I think Policy has been used as a reason by others, and I'm thinking through potential confusion if we have both (ie. do User Strikes get more confusing?)

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -224,26 +224,38 @@ export default class JobDecisioning {
...
-      const requiresReason =
-        await this.manualReviewToolSettings.getRequiresDecisionReason(orgId);
+      const requiresReason = isIgnoreDecision
+        ? await this.manualReviewToolSettings.getRequiresDecisionReasonOnIgnore(
+            orgId,
+          )
+        : await this.manualReviewToolSettings.getRequiresDecisionReason(orgId);
       if (requiresReason) {
         throw makeMissingRequiredDecisionReasonError({
           detail: 'This org requires every decision to include a reason.',
```
juanmrad:
```suggestion
          detail: 'This org requires a decision reason for this decision.',
```

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
juanmrad: I agree decision reason being always shown it makes more sense.

juanmrad (approved):
minor nit otherwise LGTM

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
taobojlen: Yep, here's an example where I've selected Ignore but not submitted yet. A reason is not required here.

It just shows the text box!

Screenshot 2026-06-16 at 13-29-57 Review Job

PR #781, opened 2026-06-15
(manual review) Show decision reason when a decision is opened via direct link

taobojlen:
## Why

I noticed a small bug. The decision details page (accessible e.g. by going to the "recent decisions" tab and clicking into one) did not show the decision reason, even though the frontend is wired up to show it.

The issue was that the SQL query never actually selects this column. This PR fixes that, so now the decision reason appears as you'd expect.

Screenshot 2026-06-15 at 15-12-49 Review Job

## Summary by CodeRabbit

* **New Features**
  * Decision details now include the decision reason, providing greater transparency into decisions made.


julietshen (approved):
lgtm

taobojlen:
Never mind -- looks like https://github.com/roostorg/coop/pull/772 already implements this fix!

PR #782, opened 2026-06-15
ci: typecheck db and migrator in CI

taobojlen:
## What

Adds typecheck coverage for the `db` and `migrator` packages in CI. I noticed that `npm run build` would fail locally due to typechecking errors -- this PR fixes those and ensures it won't happen again.

## Migrator type errors

`migrator` had two pre-existing type errors in the `apply-scripts` command: its `name` positional was never declared, so yargs typed it `unknown`. Declared it as a string (mirroring the `add` command) and asserted non-null where the existing `check()` already guarantees presence.


juanmrad (approved):
LGTM

PR #783, opened 2026-06-15
auth: fix SAML cross-org authentication bypass

taobojlen:
## Summary

Fixes GHSA-2v93-383c-9fw2.

## Testing

I verified that SAML login still works locally.

## Release Notes

* **Bug Fixes**
  * Improved SAML sign-in to strictly scope user resolution to the organization from the callback route, preventing cross-organization authentication.
  * Added clearer handling for missing/invalid organization identifiers and missing/invalid email claims during SAML login.

* **Tests**
  * Added coverage to verify correct behavior when email and organization match, and when they don’t (including expected failure scenarios).

### DIFF
--- a/server/api.ts
+++ b/server/api.ts
@@ -21,10 +21,7 @@ import { GraphQLError, type GraphQLFormattedError } from 'graphql';
 import helmet from 'helmet';
 import passport from 'passport';
 
-import {
-  kyselyUserFindByEmailAndOrg,
-  kyselyUserFindById,
-} from './graphql/datasources/userKyselyPersistence.js';
+import { kyselyUserFindById } from './graphql/datasources/userKyselyPersistence.js';
 import resolvers, { type Context } from './graphql/resolvers.js';
 import typeDefs from './graphql/schema.js';
 import { authSchemaWrapper } from './graphql/utils/authorization.js';
@@ -194,19 +191,9 @@ export default async function makeApiServer(deps: Dependencies) {
       // assertion authenticating one org can never resolve a user from another
       // (GHSA-2v93-383c-9fw2). Same helper for the signon and logout verify.
       async (req, profile, done) =>
-        resolveSamlUser(
-          async (opts) => kyselyUserFindByEmailAndOrg(KyselyPg, opts),
-          req,
-          profile,
-          done,
-        ),
+        resolveSamlUser(KyselyPg, req, profile, done),
       async (req, profile, done) =>
-        resolveSamlUser(
-          async (opts) => kyselyUserFindByEmailAndOrg(KyselyPg, opts),
-          req,
-          profile,
-          done,
-        ),
+        resolveSamlUser(KyselyPg, req, profile, done),
     ),
   );
 
--- a/server/api.ts
+++ b/server/api.ts
@@ -6,7 +6,11 @@ import { ApolloServerPluginLandingPageDisabled } from '@apollo/server/plugin/dis
 import { expressMiddleware } from '@as-integrations/express5';
 import { makeExecutableSchema } from '@graphql-tools/schema';
 import { MapperKind, mapSchema } from '@graphql-tools/utils';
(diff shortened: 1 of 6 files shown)


server/graphql/utils/resolveSamlUser.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  try {
+    const user = await kyselyUserFindByEmailAndOrg(db, { email, orgId });
+    // we should have already checked for this, but couldn't hurt to check again
+    if (user == null) {
+      return done(makeLoginUserDoesNotExistError({ shouldErrorSpan: true }));
+    }
+
+    return done(null, user);
+  } catch {
```
juanmrad: This is a very broad catch. could we catch specific errors? Would be good to consider `catch (e)` + `tracer.logActiveSpanFailedIfAny(e)` (as verifyEmailPasswordCredentials does) so genuine DB outages during login are observable.

server/graphql/utils/resolveSamlUser.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Resolves the authenticated user for a SAML assertion, binding the lookup to
+ * the org named in the callback path (`/saml/login/:orgId/callback`).
+ *
+ * Security (GHSA-2v93-383c-9fw2): the assertion's signature is verified against
+ * the path org's IdP cert, but the user must also belong to that same org.
+ * Looking up by email alone would let an assertion signed by org A's IdP
+ * resolve a user who lives in org B (same email across tenants), creating a
+ * session as that cross-tenant user. Scoping the lookup to the path org closes
+ * the bypass: a user from another org simply isn't found, and login fails.
```
juanmrad: This is a bit too long we could scale down on the annotation here the security issue should have enough context for future reference.

server/graphql/utils/resolveSamlUser.ts
```diff
@@ -0,0 +1,61 @@
...
+  profile: Pick<Profile, 'email'> | null,
+  done: VerifiedCallback,
+): Promise<void> {
+  const rawOrgId = req.params['orgId'];
+  const orgId = typeof rawOrgId === 'string' ? rawOrgId : undefined;
+  if (!orgId) {
+    return done(
+      makeNotFoundError('orgId not found in path.', { shouldErrorSpan: true }),
+    );
+  }
```
juanmrad: We have a similar check on `getSamlOptions` and with it being in `resolveSamlUser`. A tiny shared `getOrgIdFromPath(req)` helper would help to DRY as other functions may use this similar check.

server/api.ts
```diff
@@ -190,52 +187,13 @@ export default async function makeApiServer(deps: Dependencies) {
...
-          );
-        }
-      },
+      // Bind the user lookup to the org named in the callback path so an
+      // assertion authenticating one org can never resolve a user from another
+      // (GHSA-2v93-383c-9fw2). Same helper for the signon and logout verify.
+      async (req, profile, done) =>
+        resolveSamlUser(KyselyPg, req, profile, done),
+      async (req, profile, done) =>
+        resolveSamlUser(KyselyPg, req, profile, done),
```
juanmrad: You can move this callback into a function:
```ts
const verify = (req, profile, done) => resolveSamlUser(KyselyPg, req, profile, done)
```

and call verify here twice. mostly organizational/cosmetic

juanmrad (approved):
mostly nits, but you can fix before merging

server/graphql/utils/resolveSamlUser.ts
taobojlen: Very fair, actually I'll remove the security section from this comment -- we have test cases to enforce the correct behavior and git history will have details should anyone need it!

server/api.ts
taobojlen: I tried this out but I found it made the code a bit harder to read/verbose when defining the typing of the `verify` helper -- so I left it out for now!

PR #785, opened 2026-06-15
[Easy] Add missing @parcel/watcher dev dependency

reitblatt:
## Context & Requests for Reviewers

`npm run generate:watch` (used by `npm run start`) fails without this dependency. 
## Tests

`npm run start` failed before. Now it succeeds.

Confirmed compatible license (MIT), confirmed no additional vulnerabilities with `npm audit`.

## Summary by CodeRabbit

* **Chores**
  * Updated development dependencies.

---

**Note:** This release contains no user-facing changes. Updates are internal development tools only.

PR #786, opened 2026-06-15
client: Validate email address format for invites

### LINKED ISSUE #411: Adding comma-separated email addresses in User Management creates one long "email address"
Image

reitblatt:

Adds UI side email validation for user invite flow. Forces a single valid email address per-invite.

Fixes #411
Depends on #785

## Tests

Had Claude Code add a regression test for user invite page. Seems fine at first glance. 

Screenshots show new behavior:

Invalid email: button disabled
Screenshot 2026-06-15 at 16 03 31

Valid email: button enabled
Screenshot 2026-06-15 at 16 00 19

Valid email with comma: button disabled
Screenshot 2026-06-15 at 16 00 27

Two valid emails: button disabled
Screenshot 2026-06-15 at 16 00 37

## Summary by CodeRabbit

* **New Features**
  * Added local email validation for user invites, blocking submission until the email is valid and a role is selected.
  * Invite button now disables with user-facing explanatory tooltips when inputs are incomplete or invalid.
  * Form input validation visuals were improved, including accessible invalid state indication.
* **Tests**
  * Added coverage for the invite flow: disabled/enabled states, invalid email scenarios, role selection gating, and validation styling changes.

### DIFF
--- a/client/src/webpages/dashboard/components/CoopInput.tsx
+++ b/client/src/webpages/dashboard/components/CoopInput.tsx
@@ -8,7 +8,7 @@ export default function CoopInput(props: {
   value?: string;
   error?: boolean;
 }) {
-  const { type, placeholder, onChange, disabled, error } = props;
+  const { type, placeholder, onChange, disabled, value, error } = props;
   return (
     <input
       className={`block w-full px-3 py-2 ring-2 text-base border-solid rounded placeholder:text-gray-500 disabled:opacity-50 disabled:pointer-events-none ${
@@ -20,7 +20,8 @@ export default function CoopInput(props: {
       placeholder={placeholder}
       onChange={onChange}
       disabled={disabled}
-      value={props.value ?? ''}
+      aria-invalid={error}
+      value={value ?? ''}
     />
   );
 }
--- a/client/src/webpages/dashboard/components/CoopInput.tsx
+++ b/client/src/webpages/dashboard/components/CoopInput.tsx
@@ -6,11 +6,17 @@ export default function CoopInput(props: {
   onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
   disabled?: boolean;
   value?: string;
+  error?: boolean;
 }) {
-  const { placeholder, onChange, disabled } = props;
+  const { type, placeholder, onChange, disabled, error } = props;
   return (
     <input
-      className="block w-full px-3 py-2 ring-2 text-base border-gray-200 border-solid rounded ring-gray-200 placeholder:text-gray-500 focus:border-primary focus:ring-primary/20 disabled:opacity-50 disabled:pointer-events-none"
+      className={`block w-full px-3 py-2 ring-2 text-base border-solid rounded placeholder:text-gray-500 disabled:opacity-50 disabled:pointer-events-none ${
+        error
+          ? 'border-red-400 ring-red-400 focus:border-red-400 focus:ring-red-400/20'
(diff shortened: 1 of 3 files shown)


client/src/webpages/dashboard/components/CoopInput.tsx
```diff
@@ -6,11 +6,17 @@ export default function CoopInput(props: {
...
   disabled?: boolean;
   value?: string;
+  error?: boolean;
 }) {
-  const { placeholder, onChange, disabled } = props;
+  const { type, placeholder, onChange, disabled, error } = props;
   return (
     <input
-      className="block w-full px-3 py-2 ring-2 text-base border-gray-200 border-solid rounded ring-gray-200 placeholder:text-gray-500 focus:border-primary focus:ring-primary/20 disabled:opacity-50 disabled:pointer-events-none"
+      className={`block w-full px-3 py-2 ring-2 text-base border-solid rounded placeholder:text-gray-500 disabled:opacity-50 disabled:pointer-events-none ${
```
taobojlen, line 14: nit: it could be nice to make `error` a string so we can show a user-friendly error here rather than just a red outline.

taobojlen (approved):
Tested locally, works! One tiny nit but it's really not blocking, your call whether it's worth it :)

(I'm not a code owner so I don't think my review allows you to merge, but hey, adding it anyway)

juanmrad:
@julietshen would love your input, but IMO we should instead of block the comma split it and treat it as separate emails so an admin could copy a list of emails to add. 

Basically turn the input into something that allows multiple values like our text inputs, and the endpoint to allow multiple users to then display all user's created. This is a slightly bigger refactor, but still scoped to this area, and a better UX.

client/src/webpages/dashboard/components/CoopInput.tsx
reitblatt: Since this component isn't doing the validation, I think it should be up to the parent to decide how to communicate the specific error back to the user. In that case, a boolean here is the simplest, and most consistent with '<input>' in React, which would do something similar by applying the [:valid or :invalid pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/email).

But I'm really not much of a front-end guy, so open to push back.

julietshen:
> I'm okay merging this change first and then creating a new issue to track that, I think this change is valuable on its own and we can keep working towards that state (because I agree adding multiple users would be handy)

I want to think through the UX and make sure multiple users work too. Would that invite link be valid?

client/src/webpages/dashboard/components/CoopInput.tsx
julietshen: tagging in @calebmcquaid but this looks good to go! (Also @taobojlen your reviews do count ;) )

juanmrad:
>I want to think through the UX and make sure multiple users work too. Would that invite link be valid?

Each user will get their own invite link. and we can return the same to the response a list of invite links.

client/src/webpages/dashboard/components/CoopInput.tsx
calebmcquaid: Thanks for doing this @reitblatt! Looks like this component isn't used a ton of places right now, so I'm fine with whichever API we land on. Would probably be worth documenting that we expect the parent to handle the errors if we go that route.

2 things: `value` is mentioned in the props and should probably be added to the destructure. And I'd add `aria-invalid={!!error}` while we're here for a11y

calebmcquaid (approved):

PR #792, opened 2026-06-16
Extend parameterized actions to proactive rules and user strikes

### LINKED ISSUE #791: Parameterized Actions do not work with Proactive rules and user strikes
### What happened

When you create actions that allow parameters there are no way to configure proactive rule action or user strikes with parameters for these actions.

### Version/commit

_No response_

### Affected browser(s)

_No response_

### Relevant logs/errors

```shell

```

### Anything else

_No response_

### Screenshots

_No response_

### Checklist

juanmrad:
Fixes #791 

On #377 We added parameterized actions as a way for platforms to define Custom actions that allow allow parameters to be sent as a way to configure things like "Ban User" to have some Time. These were exposed on MRT and allow moderators to provide values for those parameters, but for proactive rules and user strikes we did not provide this capabilities.

Proactive rules and strike thresholds fire with no human present, so the value cannot be defined during the request as we send the action and instead needs to be pre-defined as part of the flow. These should be configured when the action is attached in the rule editor and the strike-threshold settings.

Screenshot 2026-06-15 at 9 59 30 PM
Screenshot 2026-06-15 at 9 59 19 PM
Screenshot 2026-06-15 at 9 53 23 PM
Screenshot 2026-06-15 at 9 53 20 PM

## Release Notes

* **New Features**
  * Configure custom action parameters when creating or updating rules
  * Configure custom action parameters for user strike thresholds
  * UI displays which action parameters are required and which are already configured
  * Parameter values automatically persist with rules and thresholds
  * Missing required parameters are clearly highlighted during setup


server/graphql/modules/org.ts
```diff
@@ -243,6 +250,61 @@ export async function resolveOrgActions(
...
+        continue;
+      }
+      const spec = specByActionId.get(actionId) ?? [];
+      const values = validateActionParameterValues(spec, raw[actionId]);
+      if (Object.keys(values).length > 0) {
+        // Validated values are JSON by construction (the validator only emits
+        // coerced primitives/arrays).
+        validated[actionId] = values as JsonValue;
+      }
+    }
```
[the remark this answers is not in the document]
taobojlen: @juanmrad this seems like a real bug!

server/rule_engine/RuleEngine.ts
```diff
@@ -223,19 +224,57 @@ class RuleEngine {
...
+            await this.getRuleActionsEventuallyConsistent({
+              orgId: evaluationContext.org.id,
+              ruleId: rule.id,
+            });
+          return [rule, actionsWithParameters] as const;
         }),
       ),
     );
 
+    const actionableRulesToActions = new Map(
```
taobojlen, line 280: Can we add a test case for this? I.e. there's a rule with a parameterized action, it fires, the expected action + the correct parameter is taken?

server/rule_engine/RuleEngine.ts
```diff
@@ -223,19 +224,57 @@ class RuleEngine {
...
+            actionsWithParameters.map(
+              (it) => it.action,
+            ) satisfies readonly ReadonlyDeep<Action>[] as readonly Action[],
+          ] as const,
+      ),
+    );
+
+    // Map of actionId -> the parameter values configured on the rule that
+    // attached it. When several rules attach the same action (deduped below),
+    // the first non-empty configuration wins, mirroring the action dedup.
```
taobojlen: It may be good to have a test case for this behavior too

server/services/moderationConfigService/moderationConfigService.ts
```diff
@@ -296,6 +296,14 @@ export class ModerationConfigService implements ReturnsModerationConfigTypes {
     return this.actionOps.getActionsForRuleId(opts);
   }
 
+  async getActionsWithParametersForRuleId(opts: {
```
taobojlen: would it be preferable to instead modify `getActionsForRuleId` so that it includes the actions' parameters, rather than adding a new function for it? (i'm thinking that it could make it simpler for the caller -- no risk of accidentally calling `getActionsForRuleId` when you actually do need the parameters)

server/graphql/modules/rule.ts
```diff
@@ -57,6 +64,7 @@ const typeDefs = /* GraphQL */ `
     status: RuleStatus!
     conditionSet: ConditionSet!
     actions: [Action!]!
+    actionParameters: [RuleActionParameterValues!]!
```
taobojlen, line 67: why add this as a separate field, rather than adding `parameters: JsonObject` to the `Action` graphQL type?

server/services/moderationConfigService/modules/actionParameterValueValidation.ts
```diff
@@ -75,14 +78,48 @@ export function validateActionParameterValues(
...
+    return validateActionParameterValues(spec, opts.rawValues);
+  } catch (error) {
+    // This pure module has no tracer/span, so log directly rather than via one.
+    // eslint-disable-next-line no-restricted-syntax
+    logErrorJson({
+      message: `Configured action parameter values failed validation; falling back to defaults actionId=${opts.actionId}`,
+      error,
+    });
+    try {
+      return validateActionParameterValues(spec, null);
```
taobojlen, line 109: I'm a little unsure about this behavior.

Say that somehow, the DB ends up with an invalid action here -- as a user I think I'd rather see errors, instead of the system taking an unintended action (like using the default value).

For example, maybe I have a rule that bans users for 1 day, but due to some error it falls back to the default of banning them for 30 days. That could have a pretty significant impact!

What do you think of instead using strict validation logic here too? With sufficient testing I'd hope we can be pretty confident that we won't get invalid action parameters in the DB.

taobojlen (commented):
Thank you for this!

I have two main questions, about the GraphQL API shape and about the strictness of validation for proactive rules. Curious to hear how you think about these.

Regardless, I think it's probably a good idea to have tests for user strikes/proactive rules with parameterized actions -- since if there's a bug, it's not the kind of thing a user might immediately notice.

server/services/moderationConfigService/modules/actionParameterValueValidation.ts
juanmrad: That's why actions should have safe defaults. we do have validation on the parameters to ensure you don't put something invalid.

taobojlen: OK, I won't block on this! But to share my reasoning a bit further:

> we do have validation on the parameters to ensure you don't put something invalid.

If we trusted the validation completely, then we wouldn't need a try/catch block here. If we hit the catch block, it means that something in Coop is not working as expected -- I think we'd want to know about that quickly so we can fix the bug!

In general, I have a very strong bias against silent fallbacks since they can make a system harder to reason about. So you'll probably see me making similar comments often 😆

server/graphql/modules/rule.ts
taobojlen: @juanmrad friendly ping on this -- what do you think of the suggestion of changing the ActionType instead?

server/rule_engine/RuleEngine.ts
taobojlen: @juanmrad ty for adding the `resolveConfiguredParametersByActionId` tests!

For this comment I meant a test of the actual rule execution. Without something like an integration test it's a bit hard to have confidence that this all works as expected! But not blocking if you're confident in this code

taobojlen (commented):
Same question about the GraphQL API shape!

server/services/moderationConfigService/modules/actionParameterValueValidation.ts
juanmrad: I hear you on the general principle, silent fallbacks can absolutely mask bugs. In this case though, the tradeoff is different because this code path runs during automated rule execution (proactive rules and user-strike thresholds). If we throw here:

1. The entire action-publishing for that item gets aborted. That meaning no action is taken at all on content that matched a live rule.
2. There's no human in the loop to see or react to the error.
3. The adopter configured defaults specifically as a safety net for exactly these situations.

The logErrorJson call ensures this surfaces in monitoring, so we will know about it quickly and can fix the underlying validation bug. But in the meantime, an action with a safe default (configured by the adopter) still fires rather than silently dropping enforcement.

Think of it as: "log loud, fail gracefully" rather than "fail loud, enforce nothing." The risk of a silent drop (rule doesn't fire, bad content stays up) is worse than the risk of using a default parameter value.

server/graphql/modules/rule.ts
juanmrad: Done

taobojlen (approved):
🙏 thank you for this; it's very cool! and thank you for addressing the comments :)

taobojlen:
@juanmrad  oh i thought the failing CI was a false positive but think it might be real?

juanmrad:
@taobojlen Not a real issue I think we should remove the migration order check. makes no sense given migrations are not run sequentially, but based on what has not yet run.

PR #797, opened 2026-06-16
Add CLAUDE.md symlinks to AGENTS.md

taobojlen:
## Context

Claude Code doesn't read `AGENTS.md` files by default; it only reads `CLAUDE.md` (very annoying)!

This PR creates `CLAUDE.md`s that symlink to our `AGENTS.md`s.

This is the recommendation in the global ROOST `AGENTS.md`: https://github.com/roostorg/community/blob/main/software-development-practices/agents.md

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


julietshen (approved):
that's so annoying! I had no idea. thanks!

julietshen (approved):
that's so annoying! I had no idea. thanks!

CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
juanmrad, line 1:
```suggestion
@AGENTS.md
```

I think you are supposed to do this with markdown import.

.github/CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
juanmrad, line 1: same here

juanmrad (changes requested):

CLAUDE.md
taobojlen: it's not a markdown import! the files are symlinks, so whenever an agent (or anything else) reads `CLAUDE.md` it returns the contents of `AGENTS.md`.

verified this by asking claude code about the contents of CLAUDE.md; it answers without using any tools because it's automatically loaded into the context.

juanmrad (approved):
👍🏻

PR #807, opened 2026-06-18
Add persistent MRT queue recovery

Perkeloni:

Core MRT recovery implementation: Redis-backed recovery state, the scheduled recovery job, admin reset mutation, enqueue-failure persistence, and the extracted recovery-candidate loader helper.

This PR depends on the separate migration PR #810 landing first.

## Tests

Passed locally before the split:
- docker compose run --rm codegen-check
- docker compose run --rm backend npm run lint
- docker compose run --rm backend npm run build
- docker compose run --rm client npm run lint
- docker compose run --rm client npm run build
- docker compose run --rm test

Additional regression coverage was added for enqueue failures and recovery-state persistence.

## Rollout Plan

Merge PR #810 first, then land this PR on top of it.


juanmrad:
@Perkeloni We have this script: https://github.com/roostorg/coop/blob/main/server/bin/recoverMrtQueueLib.ts which is to recover jobs from the queue. How much different is the approach from this pr to this script and is there any benefit over that 🤔.

Perkeloni:
> @Perkeloni We have this script: https://github.com/roostorg/coop/blob/main/server/bin/recoverMrtQueueLib.ts which is to recover jobs from the queue. How much different is the approach from this pr to this script and is there any benefit over that 🤔.

@juanmrad 
Hey Juan, I already reuse some of the components in that lib:
https://github.com/roostorg/coop/pull/807#issuecomment-4748304694

The thing I'm skipping is more this: https://github.com/roostorg/coop/blob/main/server/bin/recover-mrt-queue.ts
Which from what I've read is user operated, the underlying actions are shared between both.

I'm still tweaking my implementation but in essence:
The same way you have a orchestrator for a manual operator I created an orchestrator for the automated job.
One requires flags and arguments, the other works based off of configs, the CLI version is also 1 orgID 1 queueID whereas the job is any recoverable orgs and queues.
The original manual input has no persistent retry state hence the existance of the new table.
The automated orchestrator is also more scalable, built with concurrency and configs in mind.
There's a few more differences but that's the gist.

I understand your point though, there is something to say about every single contributor adding a new table and a new script, the reason I chose to implement it this way is because I think the two requirements differ enough to make them their separate structure.
Either that or I'd have to refactor the old CLI version to conform with the automated one.

One such requirement for the automation that doesn't exist for the CLI version is recording failures.
The CLI version is happy to inform the operator that it failed, an automated job must persist a record of what failed and why. ( and I was a chicken when it came to altering pre existing code 😁 )

Happy to discuss more but these are just my thoughts!
Also happy to refactor the old CLI to keep the same operator experience while supporting the necessary nits for an automated job.

reitblatt (changes requested):
@Perkeloni Thanks for the contribution! 

As @juanmrad pointed out, it would be easier to understand this PR if you added a detailed rationale for why we think we should automate this, as well as a high level explanation of the delta between the existing solution (the manual script) and this one.

To my understanding, we expect this to be a very rare failure scenario. In that case, requiring a manual clean-up is not only reasonable, but desirable: if you don't do it often, you want someone paying close attention to make sure it works as expected.

Perkeloni:
@juanmrad @reitblatt 

Alright finally time to revisit this.
I have found some things to fix in this PR as I was re-reading it to re contextualize myself.
To be fair the rationale for the automation is simply because I saw a ticket asking for this feature and given that the manual path was already there I assumed it meant to add an automated solution.

Perhaps it's a case of a stale ticket.
From a high level the main thing this PR adds that wasn't there prior is the automation itself with some guardrails and persistent retries.
It's optional as defined by a boolean flag and it defaults to false.

I can however understand the danger of operationalizing and automating these tasks without a human doing it.
I'm happy to close this as "Already done" if you think it best, otherwise I have a few more fixes to put in if we want to still use this.

It's no bother for me to chuck this away it served as a great exercise to familiarize myself with coop.

Let me know what you think!

Issue #738, opened 2026-06-09
Deleting a queue fails silently when other rows still reference it


julietshen:
## What happens

Clicking the trash icon on a queue row in the Manual Review Queues dashboard opens the confirmation modal. Confirming closes the modal but the queue stays. No error, no toast, no feedback.

## Root cause

Two combined bugs.

### Server: incomplete cascade

`QueueOperations.deleteManualReviewQueue` only removes rows from `manual_review_tool.manual_review_queues` and `users_and_accessible_queues`. Five other tables have FK constraints back to the queue with no ON DELETE rule, so any row in any of them rolls back the whole transaction.

### Client: errors are swallowed

`useGQLDeleteManualReviewQueueMutation` in `ManualReviewQueuesDashboard.tsx` is configured with `onError: () => {}`. The server's FK failure (or any other failure) produces no UI signal at all. Even when the underlying deletion is fixed, a future error in this path will still be invisible.

## Repro

1. Create a queue.
2. Favorite it (star icon).
3. Try to delete it from the dashboard.
4. Expected: queue is gone. Actual: modal closes, queue remains, no error.

## Open product questions: what should happen to dependent rows?

The five blocking tables represent different decisions, not just FK noise. We should decide each before writing the fix.

| Table | What it means | Proposed behavior |
| :--- | :--- | :--- |
| `users_and_favorite_mrt_queues` | Some reviewer favorited this queue. | Delete the favorite rows. The queue is gone, the favorite is meaningless. |
| `queues_and_hidden_actions` | Admin hid certain actions from this queue's review UI. | Delete the rows. Config tied to a deleted queue. |
| `moderator_skips` | Reviewers skipped jobs in this queue. The underlying jobs are also being obliterated. | Delete the rows. Skips reference jobs that no longer exist. |
| `routing_rules` (destination) | A routing rule routes incoming reports to this queue. | **Refuse deletion** with a clear error naming the rule(s). Silently deleting orphans the rule or breaks routing. Client already gates on this. |
| `appeals_routing_rules` (destination) | An appeals routing rule routes appeals here. | **Refuse deletion** with a clear error naming the rule(s). Same reason. Client does not gate on this today. |

Open question for in-flight jobs at deletion time. `queue.obliterate({ force: true })` wipes every BullMQ job in the queue. Should those jobs be released back to the org's default queue first so they still get reviewed, or is "this queue is gone, the work goes too" acceptable? Today it is the latter, silently.

## Proposed fix

[...]

PR #808, opened 2026-06-18
Refuse queue deletion when routing rules still reference it

reitblatt:

Part of #738 (design #1).

Previously, `routing_rules.destination_queue_id` and `appeals_routing_rules.destination_queue_id` had `ON DELETE CASCADE` FK constraints, so deleting a queue would silently drop any routing rules that pointed to it, breaking routing for the org without any warning.

This PR:

- Changes both FK constraints to `ON DELETE RESTRICT` via a new migration (`db/src/scripts/api-server-pg/2026.06.17T18.15.16.restrict_routing_rules_queue_fkeys.sql`).
- Catches the resulting FK violation in `QueueOperations.deleteManualReviewQueue`, queries the blocking rule names from both tables, and throws a `QueueHasDependentRoutingRulesError` (HTTP 409).
- Surfaces the error on the client as a modal listing the blocking rule names, each linked to the routing rules dashboard, with a prompt to update or delete those rules first.

**Note on inflight jobs:** The current implementation calls `queue.obliterate({ force: true })` before the DB delete, which removes all waiting, active, and delayed Bull jobs from Redis immediately. More nuanced handling of inflight jobs (e.g. draining rather than obliterating) is tracked in #738 and will be addressed in a follow-up PR.

## Tests

- Two new regression tests in `QueueOperations.test.ts` — one for `routing_rules` and one for `appeals_routing_rules` — verify that `deleteManualReviewQueue` rejects with `QueueHasDependentRoutingRulesError` when a rule references the target queue.
- Manual: verified the error modal renders correctly with linked rule names. Temporarily disabled client-side check, and confirmed server error was surfaced back to user: 
Screenshot 2026-06-18 at 09 42 21

Note: I did not test the logic for appeals rules because I couldn't figure out how to set one up. Pointers welcome!

## (Optional) Rollout Plan

The migration adds `ON DELETE RESTRICT` constraints in place of the existing `ON DELETE CASCADE` ones. No data migration is needed. The constraint change is safe to apply to a live database — it only affects future deletes, not existing rows.

## Summary by CodeRabbit

* **New Features**
  * Added clearer feedback when deleting a manual review queue is blocked by routing rules.
  * The deletion error dialog identifies linked routing rules and explains how to resolve the issue.

* **Bug Fixes**
  * Prevented queues from being deleted while active routing rules still reference them.
  * Improved handling of queue deletion failures with informative fallback messages.
  * Ensured failed queue cleanup does not affect successful database deletion.


client/src/webpages/dashboard/mrt/ManualReviewQueuesDashboard.tsx
```diff
@@ -621,7 +678,7 @@ export default function ManualReviewQueuesDashboard() {
                         userHasPermissions(data.me?.permissions, [
                           GQLUserPermission.EditMrtQueues,
                         ]) &&
-                        rulesForQueue.length === 0 &&
+                        true &&
```
reitblatt:
```suggestion
                        rulesForQueue.length === 0 &&
```

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -465,8 +470,56 @@ export default class QueueOperations {
...
-    );
+      });
+    } catch (e) {
+      if (
+        isForeignKeyViolationError(e) &&
+        ((e as { constraint?: string }).constraint ===
+          'routing_rules_destination_queue_id_fkey' ||
+          (e as { constraint?: string }).constraint ===
+            'appeals_routing_rules_destination_queue_id_fkey')
+      ) {
```
juanmrad: Nit: The cast `(e as { constraint?: string }).constraint` is repeated twice. Consider a single narrow:
```
const constraint = (e as { constraint?: string }).constraint;
if (
  isForeignKeyViolationError(e) &&
  (constraint === 'routing_rules_destination_queue_id_fkey' ||
   constraint === 'appeals_routing_rules_destination_queue_id_fkey')
)
```

Reads better and avoids the duplication.

juanmrad (approved):
small nit. please fix merge conflicts. Otherwise LGTM

julietshen:
Hi @reitblatt! Setting up an Appeals queue requires going into the settings (new settings page released in 1.0 a month ago) and say "enable appeals". When you create a queue, you should then see a checkbox that says "This is an appeals queue" and you can populate it with jobs that correspond to those. I'll double check how to do that programmatically...

PR #814, opened 2026-06-20
Fix AGENTS.md first-run setup instructions

reitblatt:

AGENTS.md was missing instructions contained in other docs (e.g. docs/development/local.md).

1. Missing .env copy step for /client — only /server and /db were documented; client/.env.example exists and should be copied too.
1. db:create before db:update for Scylla and ClickHouse — both require the database/keyspace to exist before migrations can run. Omitting this
produces a "keyspace does not exist" error. Postgres auto-creates so it's unaffected.
1. Wrong ports — client is :3000, server is :8080; the file had :3001/:3000.
1. --website example — added https:// scheme, which create-org requires.
## Tests

Had Claude Code follow instructions, build and run ran fine.

## (Optional) Rollout Plan

n/a: dev doc only change

## Summary by CodeRabbit

* **Documentation**
  * Updated development setup instructions with explicit database initialization steps
  * Corrected local development server URLs for client and server
  * Enhanced environment configuration guidelines

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,10 +57,11 @@ cp db/.env.example db/.env
 cp client/.env.example client/.env
 
 # Create databases, then run migrations.
-# db:create is required before db:update for Scylla and ClickHouse;
-# Postgres creates its database automatically so db:create can be skipped there.
+
+npm run db:create -- --env staging --db api-server-pg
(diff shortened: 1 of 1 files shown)


juanmrad (approved):
LGTM

PR #817, opened 2026-06-20
Add per-queue "clear other reports for a user" action sweep

### LINKED ISSUE #650: Allow an action to remove all other reports for a given user
When an action is taken on a user i.e. Ban/Delete It should be possible to also clear out the queue of all other reports for this given user. 

This allows moderators to effectively take action on a user + delete any other further report from this user as they may have already been completely kicked out of the platform and no further action is required. 

The only caveat would be any reports that are marked as CSAM for this user given the sensitivity.

juanmrad:

Fixes #650. 

Adds an optional, per-queue configuration so that when a moderator takes a configured action on a user, Coop can also dispose of that user's *other* pending reports. Saving moderators from working through reports for a user who's already been actioned.

Each queue can now configure three things:
- **Trigger actions**" which custom action(s) set off the sweep.
- **Disposition**: what happens to the other reports:
  - `AUTOMATIC_CLOSE`. close them silently with a new `USER_ACTIONED` reason (no webhooks), the same way `autoCloseJobs` closes already-deleted items.
  - `IGNORE`. mark them ignored (fires the org's ignore callback if configured).
  - `SAME_ACTION`. re-apply the triggering custom action(s) to each report's item.
- **Scope**:`CURRENT_QUEUE` (only the queue the action was taken in) or `ALL_QUEUES` (every non-appeal queue in the org).

The feature is **off by default**: a queue does nothing unless it has a non-null disposition *and* at least one trigger action, so existing queues are unaffected.

Screenshot 2026-06-20 at 11 23 57 AM
Screenshot 2026-06-20 at 11 24 09 AM
Screenshot 2026-06-20 at 11 24 22 AM
Screenshot 2026-06-20 at 11 24 29 AM

## Release Notes

* **New Features**
  * Administrators can configure “Clear Other Reports for a User” on non-appeals manual review queues, including disposition, scope (current queue or all queues), and trigger actions.

* **UI Enhancements**
  * Added conditional “Trigger Actions” and “Scope” controls when the setting is enabled.
  * Save/Create is disabled with guidance when enabled but no trigger actions are selected.

* **Backend Improvements**
  * After moderator decisions, the system can automatically clear other matching pending jobs per the configured trigger logic.

* **Bug Fixes**
  * Improved handling of legacy pending job snapshots with missing IDs.

* **Tests**
  * Added Jest coverage for enabled/disabled and trigger-action behavior.

### DIFF
--- a/server/iocContainer/index.ts
+++ b/server/iocContainer/index.ts
@@ -113,6 +113,7 @@ import {
   type NormalizedItemData,
 } from '../services/itemProcessingService/index.js';
 import {
+  isReportJob,
   ManualReviewToolService,
   type ManualReviewAppealJobInput,
   type ManualReviewJobInput,
@@ -924,9 +925,11 @@ export default async function getBottle() {
         decisionComponents,
         relatedActions,
         job,
+        queueId,
         reviewerId,
         reviewerEmail,
         decisionReason,
+        suppressUserReportSweep,
       }) {
         const { orgId } = job;
         const { itemId, itemTypeIdentifier, data } = job.payload.item;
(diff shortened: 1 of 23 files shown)


juanmrad:
@coderabbitai review

server/iocContainer/index.ts
```diff
@@ -1397,6 +1400,39 @@ export default async function getBottle() {
             }
           }),
         );
+
+        // Clear all other reports for this user. Skipped for
+        // sweep-applied dispositions (so SAME_ACTION can't recurse) and appeals.
+        if (!suppressUserReportSweep && isReportJob(job)) {
+          const reportJob = job;
```
calebmcquaid: Is there a reason we're reassigning `job` to `reportJob`?

server/services/manualReviewToolService/modules/UserReportSweep.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        actionedQueueId: queue.id,
+        customActions: [triggerAction(actionedJob.payload.item)],
+        reviewerId: user.id,
+        reviewerEmail,
+      });
+
+      expect(result).toBeUndefined();
+    },
+  );
+});
```
calebmcquaid, line 274: Could we add a test to verify the case of a user being cleared of all reports except NCMEC reports?

calebmcquaid (approved):
2 minor questions/nits. Approving

server/iocContainer/index.ts
juanmrad: The reassignment `const reportJob = job` exists because `isReportJob(job)` is a type guard, and after the if check, TypeScript knows job is a report job (not an appeal job), but the narrowed type doesn't survive into  closures like the .catch() callback. 

Assigning to a new const captures the narrowed type so it's usable inside the closure and makes the intent clear at the call site (`actionedJob: reportJob`). it is mostly for code clarity TBH.

server/services/manualReviewToolService/modules/UserReportSweep.test.ts
juanmrad: Done.

PR #819, opened 2026-06-20
Make clickhouse configs for spillover on queries

juanmrad:

This helps with graceful degradation on groupby queries when they are too big so they spillover to the disk and not OOM the machine.

### DIFF
--- a/server/graphql/datasources/RuleApi.ts
+++ b/server/graphql/datasources/RuleApi.ts
@@ -6,6 +6,7 @@ import { type Kysely } from 'kysely';
 import { uid } from 'uid';
 
 import { inject, type Dependencies } from '../../iocContainer/index.js';
+import { safeGetEnvInt } from '../../iocContainer/utils.js';
 import { type ActionCountsInput } from '../../services/actionStatisticsService/index.js';
 import { type AggregationClause } from '../../services/aggregationsService/index.js';
 import { type ConditionSetWithResultAsLogged } from '../../services/analyticsLoggers/index.js';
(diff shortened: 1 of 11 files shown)


server/graphql/datasources/RuleApi.ts
```diff
@@ -603,28 +605,53 @@ class RuleAPI {
...
+    );
+    const startAt = new Date(Date.now() - lookbackDays * DAY_MS);
+
+    const runSafely = async <T>(
+      fn: () => Promise<readonly T[]>,
+    ): Promise<readonly T[]> => {
+      try {
+        return await fn();
+      } catch {
+        return [];
```
calebmcquaid, line 628: Is there a way to pass through the logs so the errors don't get swallowed?

calebmcquaid:
```suggestion
     } catch (err) {
  logErrorJson('rule insights query failed', err);
  return [];
```

server/graphql/datasources/RuleApi.ts
```diff
@@ -6,6 +6,7 @@ import { type Kysely } from 'kysely';
 import { uid } from 'uid';
 
 import { inject, type Dependencies } from '../../iocContainer/index.js';
+import { safeGetEnvInt } from '../../iocContainer/utils.js';
```
calebmcquaid, line 9:
```suggestion
import { safeGetEnvInt } from '../../iocContainer/utils.js';
import { logErrorJson } from '../../utils/logging.js';
```

calebmcquaid (approved):
Small suggestions

PR #821, opened 2026-06-22
test: migrate DB-backed tests to transactional harness

taobojlen:

This migrates the remaining tests to the new DB-transaction test harness that I added in https://github.com/roostorg/coop/pull/732.

This new helper (called `makeTransactionalTestWithFixture`) runs tests inside a database transaction that's rolled back at the end of the test. This means that test authors no longer need to worry about cleaning up the database state afterwards (which was often forgotten already!).

This leads to a) tests that are easier to write and b) better test hygiene, i.e. less chance of introducing flakiness or cross-test dependencies.

**Important for reviewers**: the diff looks pretty massive but it's mostly whitespace! Make sure to check "Hide whitespace" in GitHub's review UI.

In two test files, the moderationService + the MRT one, the diff is bigger because the diffing algorithm doesn't render the changes super cleanly. But the tests continue to assert the same things as before, just written in a cleaner way (and with some existing cross-test dependencies removed).

## Tests

The entire PR 😉

### DIFF
--- a/server/api.ts
+++ b/server/api.ts
@@ -128,12 +128,6 @@ export default async function makeApiServer(deps: Dependencies) {
   /**
    * Passport & User Session Configuration
    */
-  // Tracked so `shutdown()` can stop connect-pg-simple's recurring
-  // pruneSessions timer. Without this, every test that calls
-  // makeApiServer (via makeMockedServer) leaks a setInterval that keeps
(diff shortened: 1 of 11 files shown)


taobojlen:
@coderabbit review

calebmcquaid (approved):
This is great, thank you!

julietshen (approved):
LGTM

PR #824, opened 2026-06-22
ci: make image pulls resilient to Docker Hub flakiness

taobojlen:

I was hitting occassional errors from Docker Hub when running CI. This PR tries to avoid this kind of flakiness in two ways:

- by using `mirror.gcr.io`, Google's pull-through mirror of Docker Hub
- and also by retrying pulls a few times.

Hopefully this makes CI less flaky.

I made one other change to test this: CI now runs when we modify files like `apply_pr_changes.yaml` so we don't accidentally merge anything that breaks CI.

## Tests

You can see CI passing on this PR!

## Summary by CodeRabbit

* **Chores**
  * Improved CI/CD infrastructure for Docker image handling and dependency management in automated checks.

### DIFF
--- a/.github/actions/prepare-docker-images/action.yml
+++ b/.github/actions/prepare-docker-images/action.yml
@@ -1,0 +1,51 @@
+name: Prepare Docker images
+description: >
+  Pull (and optionally build) images from a docker-compose
+  file, with special handling to avoid flakiness (since Docker Hub
+  sometimes returns errors for no good reason).
+
+inputs:
+  pull:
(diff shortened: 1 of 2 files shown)


juanmrad (approved):
I am not opposed, but are the error 429 over just problems with infra? if so I think we will keep having issues and instead we need to add a docker login we can use to increase the rate limits. we can create a roost account for this.

taobojlen:
@juanmrad yesterday i saw some 500 errors, not 429s! so hopefully this'll handle both those (and also eventual 429s).

PR #826, opened 2026-06-22
test: handle concurrent queries in the transactional test harness

### LINKED ISSUE #221: Alleviate effort to process Dependabot updates by running automated integration tests in CI
CI Pipeline runs various static checks as well as internal server and client tests. It lacks component integration tests which means that PRs created automatically by Dependabot should still be verified that do not introduce regressions by the Reviewer.

Extending CI Pipeline to cover integration tests as well could improve Reviews' confidence in proposed changes - having a green light from the CI should allow hitting merge button right away (if there are no other comments on review). This would directly lead to higher throughput of security-related PRs which currently have been significantly reduced (great job Team!) but with many external dependencies it can change rapidly.

taobojlen:

I found a bug in the new transaction-test harness I added: it broke when handling concurrent DB queries, e.g. via `Promise.all(...)`.

This PR fixes that by serializing such queries (in tests only).

## Tests

I added an integration test to verify that this works. Note that this test doesn't actually run in CI until we fix #221 but I am on that. It passes locally, of course.

## Summary by CodeRabbit

* **Tests**
  * Added new integration coverage for transactional database behavior, including concurrent commits, concurrent rollback isolation, and nested application transactions.
  * Validated correct ordering and isolation under parallel execution to prevent race conditions and reduce deadlock risk.

* **Bug Fixes**
  * Improved reliability of transactional query execution on the pinned connection by preventing interleaving and ensuring transaction-control operations behave correctly with nested and top-level flows.

### DIFF
--- a/server/test/harness/transactionalPgPool.integ.test.ts
+++ b/server/test/harness/transactionalPgPool.integ.test.ts
@@ -164,10 +164,6 @@ describe('createTransactionalTestDb', () => {
       });
       const transactionWithRetry = makeKyselyTransactionWithRetry(db);
 
-      // `seq` captures real write order (auto-assigned at INSERT). Ordering the
-      // final read by `seq` — not by `id`, which is the dispatch index and so
-      // would make the assertion trivially circular — proves the harness ran
-      // the concurrent transactions in FIFO dispatch order.
       await sql`create table concurrency_probe (id int primary key, seq serial)`.execute(
         db,
       );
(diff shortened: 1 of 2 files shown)


juanmrad:
@coderabbitai review

juanmrad (approved):
Overall looks good to me, but will wait for coderabbit and copilot to talk.

taobojlen:
coderabbit had some good comments; addressed them.

PR #827, opened 2026-06-22
ci: run server integration tests in CI

### LINKED ISSUE #221: Alleviate effort to process Dependabot updates by running automated integration tests in CI

taobojlen:

Fixes #221 .

We have a bunch of integration tests that were never running in CI. This fixes that, so that we can merge with confidence!

Note that I had to add `--runInBand` to the integration tests. I don't love this, because it's slower (it runs the tests serially). But right now the tests are not written to be safely run in parallel.

## Tests

It's all tests!

### DIFF
--- a/.github/workflows/apply_pr_checks.yaml
+++ b/.github/workflows/apply_pr_checks.yaml
@@ -205,8 +205,7 @@ jobs:
           echo "ATTENTION: If this step fails, look in the next step for the migration logs."
 
       - # Only output the migration logs if the prior step failed, as they
-        # could be helpful for debugging in that case; otherwise, they're just
-        # noise that's costing us money in Datadog.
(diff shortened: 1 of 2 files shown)


juanmrad (approved):
lgtm

PR #834, opened 2026-06-23
Apply prettier globally and enforce it in CI

### LINKED ISSUE #799: Apply prettier globally
We use prettier for code formatting, but only in a pre-commit hook.

This means that there's quite a lot of code in the codebase that isn't formatted correctly (177 files at time of writing).

This is a downside because it means that a small change to a file might lead to lots of unrelated stylistic changes in your PR, making it harder to review. It also means that a developer can't run e.g. `npx prettier --write` without creating a massive diff.

I suggest fixing this in a single pass/PR, and then enforcing correct formatting via CI.

taobojlen:

Fixes #799 .

We had a lot of unformatted files. However, we run prettier as a git hook, which meant that our PRs often combined stylistic/formatting changes with feature work. That makes them harder to review!

So to avoid that, it's worth doing a single pass, applying Prettier's code style to our entire codebase and enforcing this via CI.

## Tests

Just formatting changes.


juanmrad (approved):
👍🏻

PR #835, opened 2026-06-23
[Fix] Ensure sweep also handles related items

juanmrad:

Sweep/fanout actions should also take into account related items they handle.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved action processing to ensure related actions are properly handled during decision cleanup operations, enhancing the completeness of action workflows.

### DIFF
--- a/server/iocContainer/index.ts
+++ b/server/iocContainer/index.ts
@@ -1426,7 +1426,7 @@ export default async function getBottle() {
                 .map((ra) => ({
                   type: 'CUSTOM_ACTION' as const,
                   actions: ra.actionIds.map((id) => ({ id })),
-                  policies: [] as { id: string }[],
+                  policies: ra.policyIds.map((id) => ({ id })),
                   itemIds: [...ra.itemIds],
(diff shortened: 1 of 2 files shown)


ThatKoffe (approved):
Tested & works

server/iocContainer/index.ts
```diff
@@ -1417,9 +1417,20 @@ export default async function getBottle() {
         } finally {
           if (!suppressUserReportSweep && isReportJob(job)) {
             const reportJob = job;
-            const customActions = decisionComponents.flatMap((decision) =>
-              decision.type === 'CUSTOM_ACTION' ? [decision] : [],
-            );
+            const customActions = [
```
taobojlen, line 1420: a test for this would be great! seems like important behavior that should be a part of our spec (i.e. test suite)?

taobojlen: (could be a fast follow)

juanmrad: yeah will push in a sec was working on them but wanted to push the fix for Kyoto to test first.

PR #840, opened 2026-06-25
ncmec: add `email` schema field role for reported user (closes #839)

julietshen:
## Summary

Closes #839. Adds an `email` schema field role for user item types and uses it as a fallback when the NCMEC additional-info webhook is not configured (or does not return email). Includes the database migration, persistence layer, GraphQL surface, and admin UI to make the role configurable end-to-end.

## What's in this PR

**Backend (`server/`):**
- `UserSchemaFieldRoles.email` + `FieldRoleToScalarType.email`
- `NCMECUserParams.email`
- `buildSubmitReportParamsFromDecision` reads the email role and surfaces it on `reportedUser.email`
- New `resolveReportedPersonEmail` helper in `ncmecReporting.ts` encoding the precedence rule (webhook wins, field-role email is the fallback)
- `ItemTypeOperations.createUserType` / `updateUserType` accept the new role; `dbResultToItemType` returns it
- Kysely `dbTypes.ts` updated for the new column on `public.item_types` + `public.item_type_versions`
- GraphQL: `email: String` on `UserSchemaFieldRoles` type + `UserSchemaFieldRolesInput`; codegen regenerated

**Migration (`db/src/scripts/api-server-pg/`):**
- Adds `email_field` column to `public.item_types` and the temporal mirror `public.item_types_history`
- `valid_email_field_field_type` CHECK constraint requires STRING type (mirrors the pattern for `display_name_field`)
- Drops and recreates `item_type_versions` materialized view to include the new column, with all four indexes and the dependent `item_type_latest_versions` view rebuilt
- Uses `OWNER TO CURRENT_USER` per AGENTS.md

**Admin UI (`client/`):**
- Adds `Email` to the field-role picker in the user item-type editor (`ItemTypeForm.tsx` + `itemTypeUtils.ts`), mirroring the existing `IP Address` picker

**Tests:**
- 3 new in `buildSubmitReportParamsFromDecision.test.ts` (role mapped + data present, role unmapped, role mapped but data absent)
- 4 new in `ncmecReporting.test.ts` for `resolveReportedPersonEmail`'s precedence rule

## AGENTS.md callouts

[...]

PR #842, opened 2026-06-26
ncmec: tighten email field role from STRING to EMAIL_ADDRESS scalar

julietshen:
## Summary

Follow-up to #840 and #841. Now that `@roostorg/coop-types@2.4.0` ships a dedicated `EMAIL_ADDRESS` scalar, switch the email field role plumbing to require it instead of any STRING-typed field. Adopters can no longer accidentally route arbitrary string fields (bio, displayName) into NCMEC reports.

Mirrors the two-PR pattern PR #559 → PR #583 used for `IP_ADDRESS`.

## What's in this PR

**Dependency bump:**
- `@roostorg/coop-types`: `^2.3.0` → `^2.4.0` in `server/package.json` + `client/package.json`; lockfiles refreshed

**GraphQL surface:**
- `EMAIL_ADDRESS` added to `ScalarType`, `SignalInputType`, `FieldType` SDL enums (`server/graphql/schema.ts`)
- Codegen regenerated; new `GQLScalarType.EmailAddress` available to consumers

**Server behavior:**
- `[ScalarTypes.EMAIL_ADDRESS]` handler in `fieldTypeHandlers.ts`: trims whitespace, treats empty as null, rejects non-string. Format validation (RFC 5321/5322 shape) deferred to a follow-up issue per the reviewer note on #840.
- `FieldRoleToScalarType.email`: `STRING` → `EMAIL_ADDRESS`

**Client:**
- `schemaFieldRolesFieldTypes[SchemaFieldRoles.EMAIL]`: `String` → `EmailAddress` — the field-role picker now requires an `EMAIL_ADDRESS`-typed field
- Exhaustive-switch updates in `ManualReviewJobFieldsComponent.tsx` and `RuleTestModal.tsx` for the new scalar

**Migration:** `2026.06.26T03.32.16.tighten_email_field_role_to_email_address.sql` drops the existing `valid_email_field_field_type` CHECK (which required STRING) and adds a new one requiring `EMAIL_ADDRESS`.

**Test fixtures:** `EmailAddressArbitrary` for fast-check, with culturally diverse example names on RFC 2606 reserved domains.

## AGENTS.md callouts

- **Database migration included.** Tightens an existing CHECK constraint. Forward-only; rollback would re-add the looser STRING constraint.
- **Dependency bump.** `@roostorg/coop-types` 2.3.0 → 2.4.0 (additive change introducing `EMAIL_ADDRESS` to `ScalarTypes`). Published via the auto-publish workflow when #841 merged.

## Migration safety note (deliberate, please confirm)

Any existing `email_field` mapping that points to a `STRING`-typed field will fail the new CHECK constraint. #840 (which introduced the role) merged hours before this; no adopter should have had time to configure it in production yet. If a deployment between #840 and this migration did configure an email role on a STRING field, the migration will throw and the operator will need to either change the schema field type to `EMAIL_ADDRESS` or null out the email role mapping before re-applying. I considered bundling a data migration but it would silently null out adopter config — preferred to fail loud.

## Test plan

- [x] `(cd server && npx jest services/ncmecService services/itemProcessingService)` — 116 passed, 9 suites, 5 snapshots
- [x] Server `tsc --noEmit` clean (pre-existing `e2e/` errors unrelated)
- [x] Client `tsc --noEmit` clean for changed files (pre-existing errors in `signal.ts`, `FrontendTracer.ts`, `itemTypeCodeSampleUtils.ts`, `RuleInsightsSamplesTable.tsx`, `extractItemDataValues.test.ts` unrelated)
- [x] Lint clean
- [ ] Reviewer: confirm the migration's "fail loud rather than silent backfill" approach is right
- [ ] CI: integration tests verify the `EMAIL_ADDRESS` handler round-trips through item submission

## Summary by CodeRabbit

* **New Features**
  * Added full support for dedicated **email** fields across the app, including dashboard views, rule testing, and backend processing.
  * Email values are now handled as a dedicated type and normalized automatically when saved.
* **Bug Fixes**
  * Fixed rule-test placeholders and improved email field rendering to display properly and behave consistently.
  * Tightened backend validation so email field mappings must reference email-capable fields only, reducing invalid configurations.


db/src/scripts/api-server-pg/2026.06.26T03.32.16.tighten_email_field_role_to_email_address.sql
```diff
@@ -0,0 +1,29 @@
+-- Follow-up to #840: now that `@roostorg/coop-types@2.4.0` ships a dedicated
+-- `EMAIL_ADDRESS` scalar (#841), tighten the `email_field` field-role CHECK
+-- constraint from STRING to EMAIL_ADDRESS so adopters can't map arbitrary
+-- string fields (e.g. bio, displayName) into NCMEC reports.
+--
+-- Safety: any existing `email_field` mappings that point to a STRING-typed
+-- field will fail validation after this swap. #840 merged hours before this
```
taobojlen: > #840 merged hours before this migration; no adopter has had time to configure the role in production

i'm afraid that's not true :)

we need a data migration to handle any deployments that may have configured an email role already.

taobojlen (changes requested):

db/src/scripts/api-server-pg/2026.06.26T03.32.16.tighten_email_field_role_to_email_address.sql
julietshen: ack! on it

db/src/scripts/api-server-pg/2026.06.26T03.32.16.tighten_email_field_role_to_email_address.sql
```diff
@@ -0,0 +1,53 @@
...
+        SELECT id, org_id, name, email_field
+        FROM public.item_types
+        WHERE email_field IS NOT NULL
+          AND NOT jsonb_path_exists(
+            (array_to_json(fields))::jsonb,
+            '$[*]?(@."name" == $"name" && @."type" == "EMAIL_ADDRESS")'::jsonpath,
+            jsonb_build_object('name', email_field)
+          )
+    LOOP
+        RAISE NOTICE 'Clearing email_field=% from item_type id=% (name=%, org=%): no EMAIL_ADDRESS field of that name exists. Reconfigure via admin UI if needed.',
```
taobojlen, line 39: i think this is a good approach!

i just tried it locally and it looks like `npm run db:update` doesn't actually print these notices, though. i think the best way to resolve this is to update the `db` code to print NOTICE events to `console.warn` or `console.error`.

once that's done, we should still make sure to tell users that when they deploy this, they should check their DB migration logs and see if any item types need to be reconfigured in the UI. otherwise it'll be easy to miss.

taobojlen (changes requested):
one more thing to fix, sorry!

db/src/scripts/api-server-pg/2026.06.26T03.32.16.tighten_email_field_role_to_email_address.sql
julietshen: ack! good idea. cc @cassidyjames for what we need to tell users

julietshen: 1. Added an afterConnect hook to the shared Sequelize driver options in pg-base.ts so pg NOTICE events are forwarded to console.warn. Applies to all migrations using this config
2. Updated the migration's header comment to direct operators to scan db:update output for NOTICE lines after applying, so cleared mappings are communicated

Verified locally. Turns out NOTICEs were being filtered server-side because Sequelize raises client_min_messages to WARNING. Lowered it back to NOTICE in the same afterConnect hook and confirmed the per-row NOTICE now surfaces in db:update output.

(draft assisted by claude)

server/services/ncmecService/ncmecReporting.test.ts
```diff
@@ -350,6 +350,9 @@ describe('NCMEC reporting', () => {
     });
   });
 
+  // toOriginalFileHashes tests moved to ./toOriginalFileHashes.test.ts
```
taobojlen, line 353: nit: we don't need a comment like this! comments should not describe what *was*, though LLMs love to do that, but they should describe what *is*

taobojlen (approved):
confirmed that it works!

though the logs are very verbose: [example](https://gist.github.com/taobojlen/84a89d531b63925caa5a5638668ba762)

users can grep for `postgres NOTICE` to find any relevant lines.

calebmcquaid (approved):

Issue #843, opened 2026-06-26
Audit NCMEC fields


julietshen:
#839 / #840 / #842 fixed a missing email on reported users: NCMEC rejected reports as "incomplete" because email had no schema-field-role path. Adopters who didn't run the `ncmec_additional_info_endpoint webhook` submitted reports without it.
                                                                                                                                                                                
Other NCMEC fields likely have the same gap.

## Audit steps

1. Cross-reference NCMEC's CyberTip XSD against what buildSubmitReportParamsFromDecision + submitReport populate today. Per field, capture:
- Sent today (yes / no / webhook-only)
- Whether a field-role path is feasible
- Priority (P1 / P2 / P3)                                         
2. For webhook-only fields with a feasible field-role path, patch                                                        
3. P1 anything NCMEC requires that coop doesn't send. Known suspects: phoneNumber, address. Probably others.
4. File a follow-up PR per field, or cluster where it makes sense.

julietshen:
## NCMEC field audit — gaps in `buildSubmitReportParamsFromDecision` + `submitReport`

Cross-referenced the [NCMEC CyberTipline ISPWS docs](https://report.cybertip.org/ispws/documentation/index.html) against `server/services/ncmecService/`. #27 (closed by #64) and the recent `email` (#840 / #842) and `ipAddress` (#559 / #583) work cover most of what was previously P0.

**Strictly required by NCMEC XSD:** `reportingPerson.email`, `incidentSummary.incidentType`, `incidentSummary.incidentDateTime`, plus transport `reportId` / `fileId` on file details. All four are populated today. So "incomplete" rejections from NCMEC are about *report-quality* heuristics beyond strict XSD validation.

### P1 (likely causes "incomplete" rejection or materially blunts report actionability)

| NCMEC field | Status today | Suggested path |
|---|---|---|
| `personOrUserReportedPerson.phone[]` | webhook-only, no field role | Add `phone` field role on `UserSchemaFieldRoles` (mirror `email` pattern) + new `PHONE_NUMBER` scalar in `coop-types` |
| `fileDetails.originalFileHash[]` (PDQ / MD5 / SHA) | not populated despite HMA running | We hash via HMA. Wire HMA's output into the report's `originalFileHash` entries with `hashType` attribute. No field role needed; we own the hashes |
| `victim` block (per #27 P0) | not populated | Verify whether the victim block ever gets populated; if not, add reviewer-decision capture of victim ID + map to a victim user item |

[...]

PR #844, opened 2026-06-26
ncmec: wire HMA hashes into outgoing report's fileDetails.originalFileHash

julietshen:
## Summary

First P1 from the audit in #843. NCMEC's CyberTipline accepts an `originalFileHash[]` element with a `hashType` attribute per entry. HMA already computes per-image hashes at item-submission time (`HMAHashBankService.hashContentFromUrl`, stored on the image object alongside `url`), but the NCMEC report builder discarded them. This wires them through.

**Stacked on #842** because the same node_modules already has `@roostorg/coop-types@2.4.0`. Base will auto-flip to main when #842 merges.

## What's in this PR

**Types (`server/services/ncmecService/ncmecReporting.ts`):**
- New `OriginalFileHash` type (`{ _text: hash; _attributes: { hashType } }`)
- `originalFileHash?: OriginalFileHash[]` added to `FileDetails` (positioned after `industryClassification`, before `ipCaptureEvent`, in line with NCMEC XSD's forensic-fields grouping)
- `hashes?: Record<string, string>` added to internal `Media` type so the report assembly can pass them through to `#upload`

**Assembly (`buildSubmitReportParamsFromDecision.ts`):**
- New `extractHashesForUrl` helper walks the item data (scalar, array, and map containers) looking for an image-shaped value (`{ url, hashes }`) whose `url` matches the reported media's URL, returns the `hashes` map
- Each `media` entry now carries through `hashes` when present

**Submit (`ncmecReporting.ts:#upload`):**
- New `toOriginalFileHashes` exported helper converts the HMA map into NCMEC's `originalFileHash[]` shape: trims, drops empties, uppercases the algorithm name into the `hashType` attribute
- `#upload` includes the resulting array in `fileDetails` when non-empty

**Tests:**
- 4 new in `buildSubmitReportParamsFromDecision.test.ts`: scalar IMAGE case, ARRAY-of-IMAGE container traversal, URL-mismatch omission, missing-hashes omission
- 4 new in `ncmecReporting.test.ts` for `toOriginalFileHashes`: uppercasing, trimming, empty-value drops, undefined-on-empty

## What's NOT in this PR (deliberate)

- The webhook's `additionalInfo.media[].fileDetails.hash` (singular) is still not wired into the outgoing report. That's a separate gap noted in the audit; this PR focuses on HMA-sourced hashes which are the higher-leverage source since HMA runs for every adopter with image fields, webhook-sourced hashes only fire when an adopter has stood up the additional-info webhook
- No filtering of which algorithms NCMEC accepts; whatever HMA produces (MD5, SHA1, SHA256, PDQ, etc.) is forwarded as-is with the algorithm name uppercased. If NCMEC rejects unknown `hashType` values, we'd add a whitelist as a follow-up

## AGENTS.md callouts

- No migration
- No `coop-types` bump
- Pure backend plumbing wiring data we already have

## Test plan

- [x] `(cd server && npx jest services/ncmecService)` — 75 passed, 5 suites
- [x] Server `tsc --noEmit` clean (pre-existing `e2e/` errors unrelated)
- [ ] Reviewer: confirm the XSD ordering (`originalFileHash` between `industryClassification` and `ipCaptureEvent`) matches the actual NCMEC sequence
- [ ] Reviewer: confirm the algorithm-name uppercasing (e.g. `md5` → `MD5`, `pdq` → `PDQ`) matches what NCMEC's `hashType` attribute expects
- [ ] CI: integration tests verify the hash round-trips from HMA through to NCMEC

### DIFF
--- a/server/services/ncmecService/buildSubmitReportParamsFromDecision.test.ts
+++ b/server/services/ncmecService/buildSubmitReportParamsFromDecision.test.ts
@@ -376,4 +376,88 @@ describe('buildSubmitReportParamsFromDecision', () => {
       expect(result.reportedUser).not.toHaveProperty('email');
     });
   });
+
+  describe('HMA hash extraction on media', () => {
+    it('attaches hashes from the matching image in item data', async () => {
+      const result = await buildSubmitReportParamsFromDecision(
+        makeInput({
+          reportedUserItemType: makeUserItemType({}),
+          reportedUserData: asNormalizedData({ display_name: 'Alice' }),
+          contentItemType: makeContentItemType({}),
+          contentData: asNormalizedData({
+            created_at: FIXED_NOW,
+            image: {
+              url: 'https://example.com/m1.png',
+              hashes: { md5: 'abc123', pdq: 'def456' },
+            },
+          }),
+        }),
+      );
+
+      expect(result.media[0]).toMatchObject({
+        url: 'https://example.com/m1.png',
+        hashes: { md5: 'abc123', pdq: 'def456' },
+      });
+    });
+
+    it('finds the matching image inside an ARRAY-of-IMAGE container', async () => {
+      const result = await buildSubmitReportParamsFromDecision(
+        makeInput({
+          reportedUserItemType: makeUserItemType({}),
+          reportedUserData: asNormalizedData({ display_name: 'Alice' }),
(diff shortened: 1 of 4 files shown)


taobojlen (approved):
LGTM!

PR #866, opened 2026-06-29
[COOP-661] remove no media ncmec gate

calebmcquaid:

`ENQUEUE_TO_NCMEC` silently refused any account with no media, and the downstream submission path failed on a payload with no media. This blocked text-only incident types that NCMEC accepts like Online Enticement, Misleading Words, grooming chat logs, and DM-based harassment.

This PR removes the media requirement from both the enqueue and submission paths so text-only reports flow through. (This is server only. The client reviewer UI and reported-message capture will follow in separate PRs.)

## Tests

- `ncmecReporting.test.ts`: 41/41 pass

There will be follow-on PRs to cover client side gating (`canSendReport`), capture reported message identifiers, and surface enqueue skip reasons

## Summary by CodeRabbit

* **New Features**
  * Incident evidence time now considers both media and message timestamps for more accurate report dating.
  * Manual human-review jobs may be queued when no media items are present.
* **Bug Fixes**
  * Improved invalid-timestamp handling and validation classification with clearer error messages.
  * Reports can now be saved with an empty reported-media list when evidence comes from messages.
* **Tests**
  * Expanded coverage for evidence timestamp selection and validation edge cases.


server/services/ncmecService/ncmecReporting.ts
```diff
@@ -512,9 +512,7 @@ export function clampIncidentDateTimeToPast(
 ): { value: string; wasClamped: boolean } {
   const maxCreatedAtMs = new Date(maxCreatedAt).getTime();
   if (Number.isNaN(maxCreatedAtMs)) {
-    throw new Error(
```
julietshen, line 515: Removing this leaves an orphan in server/services/ncmecService/ncmecReviewerErrors.ts:20 (the catalogue entry) and its test at ncmecReviewerErrors.test.ts:38-39. All good but might be good to clean up the reference there and in the test

server/services/ncmecService/ncmecReporting.ts
```diff
@@ -525,6 +523,31 @@ export function clampIncidentDateTimeToPast(
...
+    ...media.map((m) => m.createdAt),
+    ...threads.flatMap((t) => t.reportedContent.map((c) => c.sentAt)),
+  ];
+  if (rawTimestamps.length === 0) {
+    throw new Error('Report has neither media nor messages');
+  }
+  const evidenceTimestampsMs = rawTimestamps.map((raw) => {
+    const ms = raw instanceof Date ? raw.getTime() : Date.parse(raw);
+    if (Number.isNaN(ms)) {
+      throw new Error(`Invalid timestamp for incidentDateTime: ${String(raw)}`);
```
julietshen: if i understand this, this PR makes it so a single unparseable `sentAt` in any flagged thread stops the entire submission with the same error. If that's intended, that's fine but maybe update the PR description or add a comment to make it clear that's what happens. If we want to be more forgiving for messages specifically, the message branch would need to filter `NaN` instead of throwing an error.

julietshen (approved):
lgtm! a few nits

server/services/ncmecService/ncmecReporting.ts
calebmcquaid: Good call, fixed!

server/services/ncmecService/ncmecReporting.ts
```diff
@@ -525,6 +520,32 @@ export function clampIncidentDateTimeToPast(
...
+  }[],
+): string {
+  const rawTimestamps: (string | Date)[] = [
+    ...media.map((m) => m.createdAt),
+    ...threads.flatMap((t) => t.reportedContent.map((c) => c.sentAt)),
+  ];
+  if (rawTimestamps.length === 0) {
+    throw new Error('Report has neither media nor messages');
+  }
+  // updated to allow a lenient approach to timestamps. If none parse, throw an
```
taobojlen, line 538:
```suggestion
  // allow a lenient approach to timestamps. If none parse, throw an
```

code comments should explain the current state of the code, not what was! as a reader i see this and think "updated from what?" -- but more importantly that question is just not important for the reader's understanding of the codebase, so it's only a distraction

server/services/ncmecService/ncmecReporting.ts
```diff
@@ -525,6 +520,32 @@ export function clampIncidentDateTimeToPast(
...
+  // error and surface the bad data via the validation error path
+  const evidenceTimestampsMs = rawTimestamps
```
taobojlen, line 540: what's an example where an invalid date string might end up here? how might that happen?

(i am wondering if this handling is necessary or if it's unnecessarily defensive).

taobojlen (approved):
thank you!

some comments/questions. not blocking but would be good to address!

juanmrad (approved):

selena-lustig:
Hey @calebmcquaid, is this good to merge?

server/services/ncmecService/ncmecReporting.ts
calebmcquaid: That is a validation check for retries. Retires aren't validated at the GraphQL level, I don't believe.

calebmcquaid:
> Hey @calebmcquaid, is this good to merge?

@selena-lustig I would like to merge this and #881 together.

PR #865, opened 2026-06-29
CHANGELOG: Document unreleased changes in main

cassidyjames:
Helps us prep for a 1.0.2 or 1.1.0 release. All of this information was scattered throughout issues/PR descriptions, it's just been summarized in one place to prep the release notes.

## Summary by CodeRabbit

* **Documentation**
  * Added a new unreleased changelog section with categorized updates for recent work across security, review tools, investigations, and infrastructure.
  * Included a comparison link to help users review changes since the last release.

PR #868, opened 2026-06-29
PR template: add checkbox for updating CHANGELOG/docs

cassidyjames:
Just a little nudge I thought of  while I was working on #865


julietshen (approved):
thank you! lgtm.

julietshen:
should we update agents.md too?

cassidyjames:
@julietshen we recently explicitly mentioned using the PR template, so I think we're covered; if we find we need to add an explicit callout to update the CHANGELOG and docs, we can do that in a follow-up.

PR #871, opened 2026-06-30
[COOP-661] remove table constraint for NCMEC non-media submissions

### LINKED ISSUE #661: ENQUEUE_TO_NCMEC silently refuses text-only content, blocks enticement/grooming reports
## Summary

The `ENQUEUE_TO_NCMEC` action silently refuses to create a job when the
reported user has no media items, even though Coop supports text-based
NCMEC incident types like Online Enticement, Misleading Words,
Misleading Domain Name, and Unsolicited Obscene Material — all of which
NCMEC accepts with text-only evidence (chat logs, DMs, etc.).

Net effect: grooming, enticement, and text-based harassment cases
cannot be escalated to NCMEC review at all when the suspect's account
has no images.

## What's happening

`NcmecEnqueueToMrt.enqueueForHumanReviewIfApplicable` returns
`{ status: 'SKIPPED' }` whenever `allMediaItems.length === 0`
(`server/services/ncmecService/ncmecEnqueueToMrt.ts:180-181`). This
gate runs at enqueue time — before the NCMEC reviewer triages the
job and picks the `NCMECIncidentType`. So the path from "frontline
reviewer flags account for NCMEC" to "NCMEC reviewer sees chat logs
and submits an Online Enticement report" is broken end-to-end for
text-only content.

Compounding this, `ActionPublisher.ts:494` discards the SKIPPED
return value, the decision-submit GraphQL response returns success,
and the reviewer has no signal that their escalation did nothing.

## Repro

1. Org with `ncmec_org_settings.default_ncmec_queue_id` configured.
2. Pick a user with text-only posts/DMs (no images on the account).
3. In the Review Console, submit a decision with the built-in
   `Enqueue to NCMEC` action.
4. Observe: "Decision submitted ✓"; designated NCMEC queue stays at 0.

## Proposed fix

**P1 — functional:**
- Remove the `allMediaItems.length === 0` gate, or scope it to incident
  types that genuinely require media (CSAM possession/manufacture).
  The enqueue path doesn't know the incident type yet, so the cleanest
  cut is: always enqueue, let the downstream NCMEC reviewer decide
  whether there's a reportable case.
- Make sure the downstream submit path (`buildSubmitReportParamsFromDecision`,
  `ncmecReporting.submitReport`) doesn't independently require media
  for text-only incident types. Audit this.

**P2 — UX:**
- Plumb `SKIPPED` + reason back through `ActionPublisher` into the
  `submitDecision` response so any legitimate skip (e.g., already-reported
  user) surfaces as a non-blocking toast.

## Context

Found while testing escalation against Bluesky firehose content (TAP
ingestion, text-only posts). Account-level moderation often surfaces
patterns that don't have associated media — grooming DMs, predatory
profile bios, repeated harassment text — and these are the cases NCMEC
most wants reported under Online Enticement / Misleading Words.

---

🤖 Filed with [Claude Code](https://claude.com/claude-code)

## Related symptom: NCMEC Previous Messages panel always empty

The `NCMECPreviousMessages` component in the NCMEC review UI calls `getLatestUserSubmittedItemsWithThreads` with `reportedMessages: []` hardcoded, so the "X reported" badge on threads never appears and reviewers can't tell which messages triggered the report (`client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECPreviousMessages.tsx:78-83`).

This isn't a wiring oversight — the `NcmecManualReviewJobPayload` GraphQL type has no field that carries reported message identifiers (`server/graphql/modules/manualReviewTool.ts:199-204`). The payload was built around `allMediaItems` only. And on the server, `ncmecEnqueueToMrt.ts` only collects media via `#getAllMediaForUser`; no path captures message/thread identifiers when the reported item is itself a message or post.

So the same fix that removes the media-gate at enqueue time needs to also:

1. Add `allReportedThreads` (or similar) to `NcmecManualReviewJobPayload`.
2. Update `ncmecEnqueueToMrt.ts` to capture message/thread identifiers when the reported item is a message/post (not media).
3. Update the JSON-payload TS type for NCMEC jobs accordingly.
4. Thread the field through the GraphQL resolver and the `NCMECPreviousMessages` query.

Folds in #682.

calebmcquaid:

PR 2 of X to fix #661 

See #866. This PR needs to go out alongside #866. This is the migration change to change the DB constraint to allow non-media submissions to NCMEC.

## Tests

Validated against the local DB by running the migration body inside a transaction and rolling it back

## (Optional) Rollout Plan

This PR needs to be the first or second PR to go out alongside #866 

## Checklist

### DIFF
--- a/db/src/scripts/api-server-pg/2026.06.30T03.50.47.allow_text_only_ncmec_reports.sql
+++ b/db/src/scripts/api-server-pg/2026.06.30T03.50.47.allow_text_only_ncmec_reports.sql
@@ -1,0 +1,30 @@
+-- The existing `reported_media_check_non_empty` CHECK on ncmec_reporting.ncmec_reports
+-- requires >= 1 media item, which blocks storing legitimate text-only reports
+-- even after the application-side media gates were removed (#661).
+--
+-- We want to replace it with a constraint that matches the new logic. A
+-- report must carry media OR messages (but there must be one). `array_length` of an
+-- empty/NULL array is NULL, so coalesce to 0 before comparing. The column stays
+-- `jsonb[] NOT NULL`, so rows still cannot be NULL.
+--
+-- Safe for existing data: every current row satisfied the old "media non-empty"
+-- constraint, so it satisfies "media OR messages non-empty".
+--
+-- Rollback caveat: reverting to the media-only constraint will FAIL once any
+-- text-only (empty reported_media) row exists. Reverting is only clean
+-- before such rows are written; afterward it requires removing/backfilling
+-- those rows, and deleting NCMEC report records has compliance implications.
+-- Treat this as effectively forward-only.
+
+BEGIN;
+
+ALTER TABLE ncmec_reporting.ncmec_reports
+  DROP CONSTRAINT IF EXISTS reported_media_check_non_empty;
+
+ALTER TABLE ncmec_reporting.ncmec_reports
+  ADD CONSTRAINT reported_media_or_messages_non_empty CHECK (
+    coalesce(array_length(reported_media, 1), 0) > 0
+    OR coalesce(array_length(reported_messages, 1), 0) > 0
+  );
+
+COMMIT;


julietshen (changes requested):
Looks mostly good! The matching TypeScript type at `server/services/ncmecService/dbTypes.ts:47-53` still says `reported_media: NonEmptyArray<NcmecMediaReport>`. That was accurate under the old setup but won't be once text-only reports can be saved.

maybe do a discriminated union so the check for non-empty messages is captured at the type level?

pls also update CHANGELOG.md 😁  thank you!

calebmcquaid:
@julietshen the DB type changes are made in #866. I don't think I went with a discriminated union but I can revisit that!

I will update the CHANGELOG.md ✅

julietshen (approved):

taobojlen (approved):

juanmrad (approved):

PR #872, opened 2026-06-30
fix(mrt): scope queue mutations to the correct org

taobojlen:

Fixes GHSA-mf74-gf5j-hxr9.

## Tests

I added regression tests for the issue(s) here (and used them to guide the implementation).

## Checklist

- [x] **If you changed `server/models/**/{ContentTypeModel,ActionModel,RuleModel,PolicyModel}.ts`:**

- [x] **If you changed `db/src/scripts/**` and used `CREATE TABLE`, `ADD COLUMN`, or `ALTER COLUMN`:**

- [x] **If you added a new signal in `server/services/signalsService/signals/**`:**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved queue access controls so changes stay within the correct organization.
  * Prevented granting or removing queue access across organization boundaries.
  * Added stricter validation when creating or updating manual review queues to avoid invalid access changes.

### DIFF
--- a/server/graphql/modules/manualReviewTool.ts
+++ b/server/graphql/modules/manualReviewTool.ts
@@ -2460,10 +2460,11 @@ const Mutation: GQLMutationResolvers = {
       throw forbiddenError('User does not have permission to edit MRT queues');
     }
 
-    await context.services.ManualReviewToolService.addAccessibleQueuesForUser(
-      params.input.userId,
-      params.input.queueIds,
-    );
+    await context.services.ManualReviewToolService.addAccessibleQueuesForUser({
+      orgId: user.orgId,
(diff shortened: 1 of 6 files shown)


taobojlen:
@coderabbit review

taobojlen:
@coderabbit review

julietshen (approved):
LGTM! does the GHSA impact any other migrations or just these 2? I'm thinking of `createQueue (QueueOperations.ts:301)` and `updateQueue (QueueOperations.ts:380)` specifically.

taobojlen:
ah @julietshen i missed that there actually is an issue in those, too! updating this PR now with a fix.

PR #878, opened 2026-07-01
fix: Support non-standard scylla ports in DB Migrator

jess-upscrolled:
The iocContainer for Coop supports a "SCYLLA_PORT" env var, but the DB migrator does not.  This makes it harder for adopters to run migrations against Scylla

## Tests
I've run `npm run build` and `npm run typecheck`.  I've also used it to connect to our dev and prod instances and run migrations.

## (Optional) Rollout Plan
Changes should be backwards compatible.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database connection configuration to use a custom port when provided through the environment, helping Scylla-based setups connect correctly.
  * If no port is set, the app now continues using the default connection settings unchanged.

### DIFF
--- a/db/src/configs/scylla.ts
+++ b/db/src/configs/scylla.ts
@@ -24,6 +24,9 @@ const driverOpts = {
     consistency: scyllaTypes.consistencies.localQuorum,
   },
   keyspace: process.env.SCYLLA_KEYSPACE!,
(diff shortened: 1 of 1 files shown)


taobojlen (approved):
LGTM!

PR #881, opened 2026-07-03
[COOP-661] remove no media gate - client

calebmcquaid:

This is the client side of #661  (ENQUEUE_TO_NCMEC silently refused text-only content). It's the reviewer UI for reviewing and reporting no-media/text-only NCMEC cases: online enticement, grooming chat logs, DM harassment, plus a non-blocking toast when an escalation is silently skipped.

This PR is client-only and stacks on the server work: the no-media gate removal (#866), the table-constraint migration (#871), and the server side of the message panel + skip warning.

Screenshot 2026-07-02 at 12 23 18 AM
Screenshot 2026-07-01 at 11 04 44 PM

## Tests

- [ ] Verified in-browser: a no-media NCMEC job renders the Messages tab, reported message threads display, selecting messages enables send and the skip toast fires on an already-reported user.

## (Optional) Rollout Plan

This PR should land last out of the 3

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Manual reviewers can now see reported messages associated with a case.
  - Text-only and media-based cases receive more appropriate review tabs and messaging.
  - Confirmation media supports enhanced safety preferences, including grayscale and sepia display options.
  - Decision submissions may show reviewer-facing warnings, including when an NCMEC escalation may be skipped due to an existing report.

- **Bug Fixes**
  - Improved handling of missing media, loading errors, and invalid timestamps during review.


CHANGELOG.md
taobojlen, line 1: i haven't been updating the changelog in my PRs.

@cassidyjames is that something we should be doing? if so, should we enforce it or add a PR comment to remind PR authors to do it?

(i actually would prefer not to do it at PR time, just because it can lead to merge conflicts -- if we decide to do that then @calebmcquaid you can remove this file from the PR!)

client/src/graphql/generated.ts
```diff
@@ -3095,6 +3095,7 @@ export type GQLNcmecManualReviewJobPayload = {
   readonly allMediaItems: ReadonlyArray<GQLNcmecContentItem>;
   readonly enqueueSourceInfo?: Maybe<GQLManualReviewJobEnqueueSourceInfo>;
   readonly item: GQLUserItem;
```
taobojlen, line 3079: where's the server side that caused these changes?

let's include the server-side changes alongside changes to this generated file, i.e. in the same PR, so we don't get things out of sync on main.

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
```diff
@@ -232,6 +232,109 @@ const NCMEC_INCIDENT_TYPE_OPTIONS = Object.entries(
...
+  onAdditionalInfoChange: (additionalInfo: string) => void;
+}) {
+  return (
+    <>
+      <div className="!my-4 divider" />
+      <div className="flex flex-col gap-2">
+        <label htmlFor="ncmecIncidentType" className="text-base font-bold">
+          Incident Type Category
+        </label>
+        <select
```
taobojlen: don't we already have a component library in `coop-ui`? let's not deviate from our design system!

taobojlen: also, can you explain why this change is needed? i don't quite follow why it's necessary for removing the no-media gate!

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
```diff
@@ -264,32 +367,37 @@ export default function NCMECReviewUser(
   );
 
   const [erroredMedia, setErroredMedia] = useState<ItemIdentifier[]>([]);
-  const [selectedTab, setSelectedTab] = useState<'MEDIA' | 'MESSAGES'>('MEDIA');
+  // Flatten media items to their playable URLs, sorted so confirmed CSAM and
+  // the reported item come first. Computed once so the default tab and the
+  // "text-only" checks below agree: a media item that yields zero URLs must
+  // count as no media, not send the reviewer to an empty Media tab.
+  const initialMediaItemsWithUrls = uniqueMediaItems
```
taobojlen, line 271: no action needed just a comment:

this change is sensible but it took me a bit longer to review to understand if it was a refactoring or a behavioral change

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
```diff
@@ -768,87 +883,17 @@ export default function NCMECReviewUser(
...
-            onChange={(e) => setAdditionalInfo(e.target.value)}
-            placeholder="Any other context/notes to be added to the report."
-            className="min-h-[80px] w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50"
-            rows={3}
-          />
-          <span className="text-xs text-slate-500">
-            {additionalInfo.length}/3000 characters
-          </span>
-        </div>
+        <NcmecReportFormFields
```
taobojlen: oh this explains my earlier comment. this refactoring makes it harder to review -- are there any changes to the component or is it only moving code around?

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECThreadItemComponent.tsx
```diff
@@ -58,62 +58,54 @@ export default function NCMECThreadItemComponent(props: {
...
+              <ManualReviewJobMagnifyImageComponent
+                itemIdentifier={{ id: author.id, typeId: author.typeId }}
+                imageUrl={profileImage?.url}
+                magnifiedUrls={backgroundImage ? [backgroundImage.url] : []}
+                fallbackComponent={
+                  <UserAlt4 className="w-6 p-1 fill-slate-500" />
+                }
+              />
+            </span>
+            <div className="font-medium text-slate-500">
```
taobojlen: just wondering what these visual changes are? unclear if they're related to the no-media gate or not!

taobojlen (changes requested):
ty! my main request is keeping the server-side change + the change to generated graphQL types in the same PR.

client/src/graphql/generated.ts
calebmcquaid: Those changes are in #866 I believe. I will pull this out into that PR.

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
calebmcquaid: This is almost pure movement. 2 exceptions: I added the a11y fix recommended by CodeRabbit, and I added a `my-4` class on the escalation block wrapper

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECThreadItemComponent.tsx
calebmcquaid: This was just some visual tweaks such as compacting the rows and having the panel fill up the full possible width on the page. I'll revert these

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
calebmcquaid: `coop-ui`: good call. I'll swap that out

Why is it needed:  it's a refactor and I can inline it back if you prefer. What is needed is at the bottom of the file: on `main` the "send report" modal only renders inside the Media tab branch, and a text-only job lands on the Messages tab, so the modal would never appear. Moving the modal render to the top level is the functional change.

taobojlen: yeah i wouldn't do that in this PR! we should standardize on the component library at some point but it's unrelated to removing the media gate

(also fwiw i find LLM writing generally harder to read than just human-written!)

client/src/graphql/generated.ts
calebmcquaid: Correction on this. I implemented the `generated.ts` changes here after the warning work.

calebmcquaid:
@taobojlen requesting a rereview. One note going back through things: we weren't alerting users that NCMEC escalations were being skipped, so I wired up the `warnings` field to show a toast message if the user has already been reported.

Thanks!

Screenshot 2026-07-15 at 1 56 45 PM

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -462,6 +467,63 @@ export default class JobDecisioning {
...
+          operation: 'ncmecEscalationSkipWarnings',
+        },
+        (span) => {
+          span.setAttribute('org.id', opts.job.orgId);
+          span.setAttribute('item.id', opts.job.payload.item.itemId);
+          this.tracer.logSpanFailed(span, error);
+          return null;
+        },
+      );
+      return false;
```
taobojlen: this patterns feels a bit iffy. what's the failure mode for `getUserHasExistingNcmecReport`, i.e. why might it throw an error? and why is that common enough to justify this fallback?

taobojlen (changes requested):
one more comment about the new silent fallback!

server/services/manualReviewToolService/modules/JobDecisioning.ts
taobojlen: @calebmcquaid friendly ping on this :)

CHANGELOG.md
taobojlen: @calebmcquaid let's remove this from this PR!

taobojlen (commented):
still have an unaddressed comment!

selena-lustig:
Hey @calebmcquaid following up here - were you able to address the comments?

CHANGELOG.md
calebmcquaid: It should be gone

server/services/manualReviewToolService/modules/JobDecisioning.ts
calebmcquaid: Yeah, that's fair. It would only fail if the db was down or the pool was exhausted.

I changed it so it returns unknown so that there's a distinct message to be surfaced to the user.

calebmcquaid:
> Hey @calebmcquaid following up here - were you able to address the comments?

@selena-lustig they have been addressed. Once this is approved, I can merge this and #866

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
```diff
@@ -429,20 +436,23 @@ export default function NCMECReviewUser(
...
       itemId: allMediaItemsWithUrls[0].contentItem.id,
       urlInfo: allMediaItemsWithUrls[0].urlInfo,
       itemTypeId: allMediaItemsWithUrls[0].contentItem.type.id,
     });
   }
-  if (mediaInDetailView === undefined) {
+  // Media was reported but none of it could be loaded — nothing to review.
+  // (A genuinely media-less job falls through to the tabbed UI, defaulting to
+  // the Messages tab.)
+  if (allMediaItemsWithUrls.length === 0 && erroredMedia.length > 0) {
```
julietshen: does this mean that jobs whose media all fail to load will go to a dead end, even when reportable messages exist?

When every media item errors, `onMediaError` drains `allMediaItemsWithUrls` and fills `erroredMedia`, so the component returns the "Could not find any valid media or messages" screen. I think we might have a situation where the copy was updated to say "or messages" but the condition wasn't, since this screen never checks messages

server/services/ncmecService/ncmecService.ts
```diff
@@ -105,6 +105,14 @@ export class NcmecService {
     return this.ncmecReporting.getNcmecMessages(orgId, userId, reportedMedia);
   }
 
+  async getUserHasExistingNcmeReport(params: {
```
julietshen: small typo, "c" missing from NcmecReport

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
```diff
@@ -429,20 +436,23 @@ export default function NCMECReviewUser(
...
-  }
+  const reportedMessageCount = selectedThreadsWithMessages.reduce(
+    (sum, thread) => sum + thread.reportedContent.length,
+    0,
+  );
+  const isTextOnlyJob = allMediaItemsWithUrls.length === 0;
 
-  if (mediaInDetailView === undefined) {
+  if (allMediaItemsWithUrls.length > 0 && mediaInDetailView === undefined) {
     setMediaInDetailView({
```
julietshen: i think this might be dead code as written: it's seeded from `allMediaItemsWithUrls[0]` in its useState initializer, and onMediaError re-points it to the next surviving item when the list shrinks. I don't think it can ever be undefined while the list is non-empty, so `length > 0 && mediaInDetailView === undefined` never fires. Maybe delete the block instead of preserving a setState-during-render pattern?


this predates your work but this might be a good time to remove it!

julietshen (dismissed):
I used Claude to go through and do some light review, but defer to actual engineers!

Biggest flag for me is that I don't think anything ever writes `reportedMessages` into the NCMEC job payload: `NcmecEnqueueToMrt.enqueueForHumanReviewIfApplicable ` builds `{ kind: 'NCMEC', item, allMediaItems, reportHistory }`, and #866 doesn't add it either. So this resolver always hits the `?? []` default, and `NCMECPreviousMessages` sends `reportedMessages: []`, the same behavior as the TODO hardcode this PR removes. The reviewer-facing "which message was reported" wiring would never trigger.


The enqueue path already distinguishes "reported item is content, not the user" (the ternary passed to `#getAllMediaForUser`), so it could populate `reportedMessages: [{ id: input.item.itemId, typeId: input.item.itemTypeIdentifier.id }]` under that same condition. Should that go here or in #866, which already touches `ncmecEnqueueToMrt.ts`?

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
calebmcquaid: Removed 👍

client/src/webpages/dashboard/mrt/manual_review_job/v2/ncmec/NCMECReviewUser.tsx
calebmcquaid: Yes, the copy got updated but the condition didn't. I removed the dead-end screen.

Errored jobs go to the tabbed UI and flip to Messages

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -484,6 +497,63 @@ export default class JobDecisioning {
...
+    );
+    if (!escalatesToNcmec) {
+      return [];
+    }
+    const hasExistingReport = await this.getUserHasExistingNcmecReport({
+      orgId: opts.job.orgId,
+      userId: opts.job.payload.item.itemId,
+      userItemTypeId: opts.job.payload.item.itemTypeIdentifier.id,
+    }).catch((error) => {
+      this.tracer.addSpan(
```
taobojlen, line 538: let's not have this catch block at all.

we need to assume that the DB is always available. without it, coop fundamentally cannot work, users cannot get sessions, etc.

if we wanted to silently handle the case where the DB is unavailable, then our codebase would be mostly fallbacks! LLMs love to introduce this patterns of silent fallbacks but it's almost always a bad idea.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -484,6 +497,63 @@ export default class JobDecisioning {
...
+          span.setAttribute('item.id', opts.job.payload.item.itemId);
+          this.tracer.logSpanFailed(span, error);
+          return null;
+        },
+      );
+      return 'unknown' as const;
+    });
+
+    if (hasExistingReport === 'unknown') {
+      return [NCMEC_ESCALATION_SKIP_UNKNOWN_WARNING];
```
taobojlen, line 554: and then we don't need `NCMEC_ESCALATION_SKIP_UNKNOWN_WARNING` at all.

taobojlen (changes requested):

server/services/manualReviewToolService/modules/JobDecisioning.ts
calebmcquaid: This has been removed.

server/services/manualReviewToolService/modules/JobDecisioning.ts
calebmcquaid: This has been removed.

server/services/ncmecService/ncmecService.ts
calebmcquaid: Fixed

calebmcquaid:
> I used Claude to go through and do some light review, but defer to actual engineers!
> 
> Biggest flag for me is that I don't think anything ever writes `reportedMessages` into the NCMEC job payload: `NcmecEnqueueToMrt.enqueueForHumanReviewIfApplicable ` builds `{ kind: 'NCMEC', item, allMediaItems, reportHistory }`, and #866 doesn't add it either. So this resolver always hits the `?? []` default, and `NCMECPreviousMessages` sends `reportedMessages: []`, the same behavior as the TODO hardcode this PR removes. The reviewer-facing "which message was reported" wiring would never trigger.
> 
> The enqueue path already distinguishes "reported item is content, not the user" (the ternary passed to `#getAllMediaForUser`), so it could populate `reportedMessages: [{ id: input.item.itemId, typeId: input.item.itemTypeIdentifier.id }]` under that same condition. Should that go here or in #866, which already touches `ncmecEnqueueToMrt.ts`?

These have been fixed. We now send `reportedMessages: [reportedItemIdentifier]` in PreviousMessages and we run the check that the reported item isn't the user themselves. Added tests for this in `ncmecEnqueueToMrt.test.ts`

taobojlen (approved):
thank you!!

Issue #851, opened 2026-06-28
Add local NCMEC test-report harness


julietshen:
we need a local test harness to verify information being sent to NCMEC is expected.

This test should include a server bin/ script that takes a decision (or a synthetic one) and calls the existing `buildSubmitReportParamsFromDecision` to the XML serializer path inside `ncmecReporting.ts`, then writes the XML to disk instead of POSTing. This gives a per-field artifact we can diff against NCMEC's XSD for the audit, with no NCMEC-side state and no test creds needed. 

This specifically tests what coop constructs, not what NCMEC accepts

julietshen:
Findings:

### Without NCMEC additional info webhook configured:

- personOrUserReported.displayName
- personOrUserReported.ipCaptureEvent
- personOrUserReported.personOrUserReportedPerson.email[]._text
- fileDetails.ipCaptureEvent.ipAddress

### With NCMEC additional info webhook configured:

- NCMEC email attributes (verified/type)
- screenName
- publiclyAvailable
- fileAnnotations
- ipCaptureEvent.{port,possibleProxy}
- fileDetails.additionalInfo

### only through the manual review tool:

- escalateToHighPriority
- top-level additionalInfo (free text)

taobojlen:
this is resolved by #887 .

PR #887, opened 2026-07-06
add NCMEC field + integration tests

### LINKED ISSUE #858: add full-pipeline integration test (mocked NCMEC HTTP)
Add an integration test under services/ncmecService/ that uses the existing test:integ docker-compose harness. 

Test should: create a synthetic org with NCMEC settings + item types + field roles, submit synthetic items, drive submitReport end-to-end with the NCMEC HTTP layer mocked, assert on the generated request XML and the persisted ncmec_reports row. Covers wiring bugs that unit tests + XSD validation miss.

taobojlen:
This adds some more NCMEC tests: an integration test for the full flow, and unit tests to ensure that our generated XML matches what NCMEC expects. As part of writing this I uncovered a potential bug -- we didn't always generate elements in the right order in `ipCaptureEvent` elements. That's fixed here too.

Fixes https://github.com/roostorg/coop/issues/858 by adding an integration test that submits to a mocked NCMEC endpoint.

Also addresses https://github.com/roostorg/coop/issues/851, not by actually writing the XML anywhere, but by writing unit tests that are equivalent to doing that and validating the XSD. We don't want to commit NCMEC's XSD to this repo since it's technically not public (even though all the info contained in it *is* public via their docs).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NCMEC report XML generation to strictly follow required field ordering, reducing submission rejections.
  * Fixed IP capture event assembly so optional fields and values are included consistently and preserve expected data in the final submission.
* **Tests**
  * Added regression and integration tests to lock down XML output structure and end-to-end submission flow.

### DIFF
--- a/server/services/ncmecService/ncmecReporting.builders.test.ts
+++ b/server/services/ncmecService/ncmecReporting.builders.test.ts
@@ -137,9 +137,9 @@ describe('buildFileDetailsObject', () => {
       'publiclyAvailable',
       'fileRelevance',
       'fileAnnotations',
-      'ipCaptureEvent',
       'industryClassification',
       'originalFileHash',
+      'ipCaptureEvent',
       'additionalInfo',
     ]);
   });
--- a/server/services/ncmecService/fieldCoverage.test.ts
+++ b/server/services/ncmecService/fieldCoverage.test.ts
@@ -257,9 +257,67 @@ describe('NCMEC field coverage', () => {
     }
   });
 
-  it('renders all three scenarios to non-empty XML', () => {
+  it('renders all three scenarios to well-formed XML with scenario-specific markers', () => {
+    // The presence table checks the object tree; this is the only test that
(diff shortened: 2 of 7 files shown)


taobojlen:
@coderabbit review

server/services/ncmecService/ncmecReporting.builders.test.ts
```diff
@@ -137,9 +137,9 @@ describe('buildFileDetailsObject', () => {
       'publiclyAvailable',
       'fileRelevance',
       'fileAnnotations',
-      'ipCaptureEvent',
       'industryClassification',
       'originalFileHash',
+      'ipCaptureEvent',
```
taobojlen, line 142: `ipCaptureEvent` comes after `originalFileHash` in the NCMEC XSD.

julietshen (approved):
LGTM. thank you!

PR #888, opened 2026-07-07
ci: stop logging every DB query in test runs

taobojlen:
I find our CI test logs really hard to read, because we log every single DB query that occurs. For example here: https://github.com/roostorg/coop/actions/runs/28807521505/job/85426791456?pr=875

Let's not do that in CI, so we can more clearly see which tests are running.

Look at how neat these logs are (from this branch's E2E test CI run):

```
Run cd server && npm run test:e2e
> server@1.0.0 test:e2e
> playwright test --config e2e/playwright.config.ts
Running 1 test using 1 worker
  ✓  1 [chromium] › e2e/tests/login.spec.ts:3:1 › a user can log in (1.9s)
Notice:   1 passed (7.5s)
  1 passed (7.5s)
```

Easy to read!

## Summary by CodeRabbit

* **Chores**
  * Reduced database log output during CI runs.
  * Disabled database logging by default in the example environment configuration.


cassidyjames (approved):
🙏

PR #890, opened 2026-07-07
Add MRT video playback E2E coverage

taobojlen:

This adds an integration test for video rendering in the manual review tool. I'm adding this in a separate PR because there was a pre-existing bug to fix to make our video player work in local dev.

The Playwright video recording is not that useful because it's mostly a blank screen, and the video rendering is off-screen, but hey here it is regardless:

[video.webm](https://github.com/user-attachments/assets/6045fe03-312c-4820-8f89-cecc336a6c42)

Here's what it looks like just in my own browser:

CleanShot 2026-07-07 at 17 10 15

(Yes, it looks a little weird, UI-wise, but that's just the pre-existing UI).

## Tests

It's all tests!

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added video media support to manual review job content.
  - Reviewers can now view and play video attachments directly within the review interface.

- **Bug Fixes**
  - Improved video playback handling to ensure videos load and play reliably during content review.

- **Tests**
  - Expanded end-to-end coverage to validate video display, playback, and media fixture loading across supported environments.


client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobContentBlurableVideo.tsx
```diff
@@ -1,6 +1,6 @@
 import { PlayCircleFilled } from '@ant-design/icons';
 import { useEffect, useRef, useState } from 'react';
-import ReactPlayer from 'react-player/lazy';
+import ReactPlayer from 'react-player';
```
taobojlen, line 3: `react-player/lazy` is not compatible with vite: https://github.com/cookpete/react-player/issues/1443

so, for now, let's just import it normally. this will increase bundle size but at least videos will work in local dev.

calebmcquaid (approved):
👍

PR #892, opened 2026-07-07
feat: 670 weighted job priority

calebmcquaid:

Part 2 of 3 for #670, stacked on #718. Adds a third sort mode, **WEIGHTED**: jobs are scored from org-configurable weights instead of report count alone.

- Score = Σ weight × signal, scaled ×1000 into a BullMQ priority. Signals: `numReports` (raw count, weight = points per report) and `userScore` (mapped onto [0,1], worst offender = 1, so weight = points at the user's worst). Signals live in an allowlist in `JobPriority.ts`.
- New `job_priority_weights` table (PK `(org_id, property)`, `weight >= 0` CHECK) + migration widening the `job_sort_type` CHECK to allow `WEIGHTED`.
- GraphQL: `setJobPriorityWeights` mutation (rejects negative/non-finite/duplicate weights) and a weights query on the org type.
- Changing weights re-sorts every WEIGHTED queue's pending jobs via the same background sweep introduced in #718.
- UI: new Job Priorities settings  are on the Review Console tab with a plain-language summary of what the current weights mean; queue form gains a "Custom (weighted)" option.

Review focus: the weights-change re-sort path (`setJobPriorityWeights` → `#scheduleQueuePriorityRecompute`) inherits the swallowed-error design from #718 — if the sweep fails, weights save but pending jobs keep their old order with no user-visible signal. Suggestions on observability welcome.

## Tests

- `JobPriorityWeights.test.ts` — weights CRUD, validation (negative, duplicate, non-finite).
- `JobPriority.test.ts` (extended) — weighted scoring: per-signal contributions, zero-weight = no contribution, all-zero = FIFO ties.
- `jobPriorityWeightSummary.test.ts` — settings-page summary text.
- Manual: with weights (numReports=9, userScore=4), seeded jobs at 0–40 reports dequeued strictly by report count with priorities matching the formula; with (numReports=0, userScore=10) and a 1.0–4.5 user-score spread, order fully inverted to worst-user-first.

## (Optional) Rollout Plan

Ships behind the sort-mode selector: no queue behaves differently until an org sets a queue to WEIGHTED. Orgs with no weights rows get zero contributions (jobs tie → FIFO behavior), so enabling WEIGHTED before configuring weights is safe.

## Checklist

- [X] **If you changed `db/src/scripts/**` and used `CREATE TABLE`, `ADD COLUMN`, or `ALTER COLUMN`:**


db/src/scripts/api-server-pg/2026.07.06T15.43.10.add_job_priority_weights_table.sql
```diff
@@ -0,0 +1,16 @@
+-- Per-org weights for the WEIGHTED job sort mode. One row per (org,
+-- property); properties are an allowlist maintained in application code
+-- (server/services/manualReviewToolService/modules/JobPriority.ts).
+--
+-- No FK to public.orgs: tables in the manual_review_tool schema
```
taobojlen: can you explain this a bit more? why don't we want a FK to the org table?

server/graphql/modules/org.ts
```diff
@@ -806,6 +807,28 @@ const Org: GQLOrgResolvers = {
 
     return partialItemsEndpoint != null;
   },
+  async jobPriorityWeights(org, _, context) {
+    // Moderation-priority config: gate reads behind the same MANAGE_ORG
```
taobojlen: i think this comment is redundant (and quite hard to read!)

taobojlen (commented):
Just started reviewing this but then I realized that I don't actually see this weighted job priority in the #670 ticket.

Has anyone asked for this? How come we're adding it now?

db/src/scripts/api-server-pg/2026.07.06T15.43.10.add_job_priority_weights_table.sql
calebmcquaid: This wasn't a pattern used in other areas. I went ahead and added the FK.

server/graphql/modules/org.ts
calebmcquaid: Removed

julietshen:
@taobojlen for context, I think we had talked about this in the Discord channel for coop or during a working group. The way I thought about custom job ranking in queues and the ordering I discussed with @calebmcquaid and @juanmrad was:

MVP: first-in-first-out (existing) and then by # of reports
fast follow: allow orgs to create a custom way of ranking jobs in their queues (ie. # of people who have viewed the post, classifiers, etc)

taobojlen:
@julietshen thanks! if no-one has asked for it, i'd personally lean towards not doing it, mostly because sorting by FIFO + # of reports may be Good Enough.

(my instinct is always to ask if we really need to build \<thing X\>, i love simplicity!)

calebmcquaid:
> @julietshen thanks! if no-one has asked for it, i'd personally lean towards not doing it, mostly because sorting by FIFO + # of reports may be Good Enough.
> 
> (my instinct is always to ask if we really need to build <thing X>, i love simplicity!)

@taobojlen @julietshen I'm going to close this PR and rebase #893 onto #718 and we'll be much closer to merging those 2

julietshen:
Reopening because we actually do have a few adopters who would like to implement custom job ranking, particularly using signals/classifiers as a way to do so in the future.

client/src/webpages/settings/tabs/JobPriorityWeightsSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+                ) : null}
+              </div>
+            );
+          },
+        )}
+      </div>
+
+      <div className="rounded-md border border-slate-200 bg-slate-50 p-4">
+        <Text className="text-sm text-slate-600">
+          {summarizeWeighting(
```
taobojlen, line 69: is this text really needed?

IMO the `summarizeWeighting` function adds a bunch of complexity but limited benefit to the user. they can already see the weights directly so do they really need the summary?

client/src/webpages/settings/jobPriorityWeights.ts
```diff
@@ -0,0 +1,65 @@
...
+    help: 'Items with more user reports are reviewed sooner. Each report adds the weight to the item’s score. Set to 0 to ignore.',
+    example: (w) =>
+      w === 0
+        ? "Currently disabled: report counts won't affect queue order."
+        : '',
+  },
+  {
+    property: GQLJobPriorityProperty.UserScore,
+    label: 'User Score',
+    help: "Items from users with a history of policy violations are reviewed sooner. Coop assigns each user a moderation score from 1 to 5 based on the ratio of penalties they've received to total submissions. 1 means many penalties (likely a repeat offender), 5 is the default for new or clean users. Set to 0 to ignore user history.",
```
taobojlen, line 24: is this moderation score visible in the UI anywhere? if not, then i think explaining it might be overcomplicating it.

db/src/scripts/api-server-pg/2026.08.21T16.02.00.add_job_priority_weights_table.sql
```diff
@@ -0,0 +1,18 @@
+-- Per-org weights for the WEIGHTED job sort mode. One row per (org,
+-- property); properties are an allowlist maintained in application code
+-- (server/services/manualReviewToolService/modules/JobPriority.ts).
+--
+-- org_id is a real FK with ON DELETE CASCADE, matching how public tables
+-- reference orgs (actions_org_id_fkey, api_keys_org_id_fkey, hash_banks).
+-- Deleting an org drops its weights; without the constraint they would be
+-- orphaned rows that still match a recycled org id.
+CREATE TABLE manual_review_tool.job_priority_weights (
```
taobojlen, line 9: this means that weights are global for an entire org, right? i wonder if this should really be queue-scoped instead so different queues can have different weights.

what do you think? (cc @julietshen you may have opinions too based on what adopters are saying)

server/graphql/modules/org.ts
```diff
@@ -806,6 +807,25 @@ const Org: GQLOrgResolvers = {
 
     return partialItemsEndpoint != null;
   },
+  async jobPriorityWeights(org, _, context) {
+    const user = context.getUser();
+    if (user == null || user.orgId !== org.id) {
```
taobojlen, line 812: how does `org` get passed in to the resolver?

just wondering if it's even possible for `user.orgId` to not equal `org.id`.

server/services/manualReviewToolService/modules/JobPriority.ts
```diff
@@ -58,6 +78,51 @@ export function toBullPriority(score: number): number {
   return MAX_BULL_PRIORITY - Math.round(clamped);
 }
 
+// Lowest possible user score. 1 = repeat offender; 5 (`initialUserScore`) is
+// a clean user.
+const MIN_USER_SCORE = 1;
+
+// A weighted score is scaled up by this factor before becoming an (integer)
+// BullMQ priority, so fractional contributions — a user score maps onto
+// [0, 1] — survive rounding instead of collapsing into ties.
```
taobojlen, line 87: I don't really understand this logic. Why do we need to scale scores by 1000? And why exactly 1000, and not 100 or 10,000 for example?

server/services/manualReviewToolService/modules/JobPriority.ts
```diff
@@ -58,6 +78,51 @@ export function toBullPriority(score: number): number {
...
+  item: ItemSubmissionWithTypeIdentifier,
+): UserIdentifier {
+  if (item.creator) return item.creator;
+  return { id: item.itemId, typeId: item.itemTypeIdentifier.id };
+}
+
+// A user score mapped onto [0, 1]: worst offender (MIN_USER_SCORE) = 1,
+// clean user (initialUserScore) = 0.
+function normalizedUserScore(score: number): number {
+  return (initialUserScore - score) / (initialUserScore - MIN_USER_SCORE);
```
taobojlen, line 102: let's work through this...

user score = 1: (5 - 1) / (5 - 1) = 1.0
user score = 2: (5 - 2) / (5 - 1) = 0.75
user score = 3: (5 - 3) / (5 - 1) = 0.5
user score = 4: (5 - 4) / (5 - 1) = 0.25
user score = 5: (5 - 5) / (5 - 1) = 0.0

would be slightly easier to read if `initialUserScore` was also formatted as a constant, i.e. `INITIAL_USER_SCORE`, but i realize that this code was pre-existing.

server/services/manualReviewToolService/modules/JobPriority.ts
```diff
@@ -58,6 +78,51 @@ export function toBullPriority(score: number): number {
...
+// signal is unweighted all jobs tie, i.e. arrival order.
+function weightedScore(opts: {
+  weights: ReadonlyMap<JobPropertyKey, number>;
+  numReports: number;
+  userScore: number;
+}): number {
+  const reportsWeight = opts.weights.get('numReports') ?? 0;
+  const userScoreWeight = opts.weights.get('userScore') ?? 0;
+  return (
+    reportsWeight * opts.numReports +
```
taobojlen, line 121: so we normalize the user score into `[0, 1]` -- won't this mean that the number of reports will always have much more of an effect on the ranking than the user score?

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -976,6 +983,78 @@ export class ManualReviewToolService {
...
+    // WEIGHTED queue, so re-sort those queues in the background. Bump each
+    // queue's version first so a sweep already running on another instance
+    // picks the new weights up on its next pass.
+    const queues =
+      await this.queueOps.getAllQueuesForOrgAndDangerouslyBypassPermissioning(
+        opts.orgId,
+      );
+    for (const queue of queues) {
+      if (
+        queue.isAppealsQueue ||
```
taobojlen, line 1023: I vaguely remember a similar comment elsewhere but I can't find it now: why don't we want this to take effect in appeals queues?

server/services/userStatisticsService/userStatisticsService.ts
```diff
@@ -78,6 +78,46 @@ export function internalMakeUserStatisticsService(
       return score as UserScore;
     },
 
+    /**
+     * Scores for many users in one query, keyed by `"${typeId}\x00${id}"`.
+     * Users with no cached score are absent from the map (callers treat them
+     * as `initialUserScore`). Used by the queue priority sweep, where a
+     * per-user lookup would be an N+1 against the replica.
```
taobojlen, line 85: IMO no need to mention where a function is used. there isn't much benefit and such comments are likely to get stale as the codebase evolves -- and it's easy to look up anyway.

server/services/userStatisticsService/userStatisticsService.ts
```diff
@@ -78,6 +78,46 @@ export function internalMakeUserStatisticsService(
...
+      const CHUNK = 500;
+      for (let i = 0; i < opts.users.length; i += CHUNK) {
+        const chunk = opts.users.slice(i, i + CHUNK);
+        if (chunk.length === 0) continue;
+        const rows = await pgQueryReplica
+          .selectFrom('user_statistics_service.user_scores')
+          .select(['user_type_id', 'user_id', 'score'])
+          .where('org_id', '=', opts.orgId)
+          .where((eb) =>
+            eb.or(
```
taobojlen, line 101: postgres `OR` queries can be pretty slow; see e.g. https://www.cybertec-postgresql.com/en/rewrite-or-to-union-in-postgresql-queries/.

so a potential performance improvement here would be to rewrite this to a `UNION ALL` query instead.

taobojlen (changes requested):
My main question is whether this should live on the org, or if it really should be a per-queue setting! (I think probably the latter, personally).

PR #893, opened 2026-07-07
feat: 670 per reviewer skip

calebmcquaid:

Part 3 of 3 for #670. Previously, skipping a job just released the lock. The job came right back to the same reviewer, and skip was two separate client calls (`logSkip` + `releaseJobLock`) that could partially fail.

- **Skip is now one server-side operation**: `logSkip` atomically logs the skip, records the job in a per-reviewer Redis skip set (sorted set scored by expiry, with a TTL backstop), and releases the lock. The client's separate `ReleaseJobLock` call is removed.
- **Dequeue steps past skipped jobs for that reviewer only**: the scan holds skipped jobs aside under the reviewer's lock token and releases them in a `finally`, so they return to the shared pool immediately for everyone else.
- `getNextJob` now uses `block: false`, so a drained queue returns null immediately instead of long-polling and hanging the reviewer's request.
- UI: skip failures show a retry modal instead of failing silently; when no reviewable job remains (drained, or everything left is skipped by this reviewer), the reviewer is redirected to the queue list instead of a perpetual spinner.

Review focus: the held-aside-then-release pattern in `dequeueNextJobWithLock`, particularly lock-token semantics (the token is the reviewer's userId) and what happens if the release in `finally` partially fails.

## Tests

- `QueueOperations.reviewerSkips.test.ts` — skip set recording/expiry, dequeue filtering per reviewer, held-aside jobs released for other reviewers.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 **Full Changelog**: https://github.com/roostorg/coop/compare/1.0.2...main
 
+## Review Console
+
+- Added per-queue job sort modes for manual review: FIFO (default, unchanged), most-reported-first, and custom weighted — ordering is computed at enqueue via BullMQ job priority, so the locked dequeue path is unchanged (#718)
+- Added org-configurable job priority weights (Settings → Job Priorities) for weighted queues; changing weights or a queue's sort mode re-sorts already-queued jobs in a background sweep (#892)
+- Skip is now per-reviewer: a skipped job is hidden from that reviewer for 30 minutes while returning to the shared pool immediately; entering a drained or fully-skipped queue redirects to the queue list (#893)
+
 # Coop 1.0.2
 
 This release addresses reported security advisories, improves NCMEC CyberTipline reporting, and includes front-end quality-of-life improvements.
--- a/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
@@ -39,7 +39,6 @@ import {
   useGQLDequeueManualReviewJobMutation,
   useGQLLogSkipMutation,
   useGQLManualReviewJobInfoQuery,
-  useGQLReleaseJobLockMutation,
   useGQLSubmitManualReviewDecisionMutation,
   type GQLActionParameter,
   type GQLThreadManualReviewJobPayload,
@@ -221,10 +220,6 @@ gql`
   mutation LogSkip($input: LogSkipInput!) {
     logSkip(input: $input)
   }
-
-  mutation ReleaseJobLock($input: ReleaseJobLockInput!) {
-    releaseJobLock(input: $input)
-  }
 `;
 
 enum BuiltInActionType {
@@ -342,17 +337,18 @@ function ManualReviewJobReviewImpl(props: {
 
   const actionStore = useContext(ManualReviewActionStore);
 
-  const setSelectedRelatedActions = (
-    actions: ManualReviewJobEnqueuedActionData[],
-  ) => {
-    actionStore?.setActions(
-      actions.map((it) => ({
-        itemId: it.target.identifier.itemId,
-        action: it.action,
-      })),
-    );
-    selectedRelatedActionsSetter(actions);
-  };
+  const setSelectedRelatedActions = useCallback(
+    (actions: ManualReviewJobEnqueuedActionData[]) => {
+      actionStore?.setActions(
+        actions.map((it) => ({
+          itemId: it.target.identifier.itemId,
+          action: it.action,
+        })),
+      );
+      selectedRelatedActionsSetter(actions);
+    },
+    [actionStore],
+  );
 
   const { queueId, jobId, lockToken } = useParams<{
     queueId?: string;
@@ -364,12 +360,12 @@ function ManualReviewJobReviewImpl(props: {
   const mrtParentComponentRef = useRef<HTMLDivElement>(null);
   const reportedUserRef = useRef<HTMLDivElement>(null);
 
-  const resetState = () => {
+  const resetState = useCallback(() => {
     setSelectedPrimaryActions([]);
     setSelectedPrimaryPolicies([]);
     setSelectedRelatedActions([]);
     setDecisionReason(undefined);
-  };
+  }, [setSelectedRelatedActions]);
 
   const {
     data,
@@ -389,9 +385,12 @@ function ManualReviewJobReviewImpl(props: {
     onCompleted: (data) => {
       // Here, we update the URL to include the queue ID, job ID, and lock
       // token. That way, users are able to send around the URL to others.
-      // In case we can't find the required job, we can just fail silently.
       const { dequeueManualReviewJob } = data;
       if (dequeueManualReviewJob == null) {
+        // No reviewable job: the queue is drained, or everything left is
+        // skipped by this reviewer. Send them back to the queue list instead
+        // of leaving them on a perpetual loading spinner.
+        navigate('/dashboard/manual_review/queues', { replace: true });
         return;
       }
 
@@ -739,10 +738,6 @@ function ManualReviewJobReviewImpl(props: {
     fetchPolicy: 'no-cache',
   });
 
-  const [releaseJobLock] = useGQLReleaseJobLockMutation({
-    fetchPolicy: 'no-cache',
-  });
-
   const advanceToNextJobAfterInvalidation = useCallback(async () => {
     setIsAdvancingToNextJob(true);
     try {
@@ -754,8 +749,7 @@ function ManualReviewJobReviewImpl(props: {
     } finally {
       setIsAdvancingToNextJob(false);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [getNextJob, navigate]);
+  }, [getNextJob, navigate, resetState]);
 
   // Runs after the invalidate mutation resolves. Refreshes the job view,
   // and if invalidation deleted the current job, advances to the next one.
@@ -783,20 +777,21 @@ function ManualReviewJobReviewImpl(props: {
   }, [jobId, queueId, refetchJobInfo, advanceToNextJobAfterInvalidation]);
 
   const skipToNextJob = async () => {
-    // First, release the lock on the current job and log the skip
+    // Skipping is one server-side operation: it logs the skip, hides the job
+    // from this reviewer for the skip window, and releases the lock so the
+    // job returns to the shared pool for everyone else.
     if (queueId && job?.id && lockToken) {
-      await Promise.all([
-        logSkip(),
-        releaseJobLock({
-          variables: {
-            input: {
-              queueId,
-              jobId: job.id,
-              lockToken,
-            },
-          },
-        }),
-      ]);
+      const result = await logSkip();
+      if (result.data?.logSkip !== true) {
+        // Nothing was released or hidden; stay on the current job so the
+        // reviewer can retry (or decide) instead of advancing past it.
+        setModalInfo({
+          visible: true,
+          modalBody: 'Failed to skip this job. Please try again.',
+          footer: [{ title: 'Ok', type: 'primary', onClick: hideModal }],
+        });
+        return;
+      }
     }
 
     // Reset state and try to get the next job
@@ -809,7 +804,11 @@ function ManualReviewJobReviewImpl(props: {
     }
   };
 
-  if (loading || jobDataLoading || (!closedJob && !lockToken)) {
+  // `loading && !data`: the job-info query reloads in place (e.g. when its
+  // jobIds variable changes while advancing, or on refetch after an
+  // invalidation). Once we have data, keep the current view up during those
+  // reloads instead of flashing the full-screen spinner.
+  if ((loading && !data) || jobDataLoading || (!closedJob && !lockToken)) {
     return (
       <div className="flex items-center justify-center w-full h-screen">
         <ComponentLoading />
--- a/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
@@ -39,7 +39,6 @@ import {
   useGQLDequeueManualReviewJobMutation,
   useGQLLogSkipMutation,
   useGQLManualReviewJobInfoQuery,
-  useGQLReleaseJobLockMutation,
   useGQLSubmitManualReviewDecisionMutation,
   type GQLActionParameter,
   type GQLThreadManualReviewJobPayload,
@@ -221,10 +220,6 @@ gql`
   mutation LogSkip($input: LogSkipInput!) {
     logSkip(input: $input)
   }
-
-  mutation ReleaseJobLock($input: ReleaseJobLockInput!) {
-    releaseJobLock(input: $input)
-  }
 `;
 
 enum BuiltInActionType {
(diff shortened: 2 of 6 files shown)


client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -342,17 +337,18 @@ function ManualReviewJobReviewImpl(props: {
...
-  ) => {
-    actionStore?.setActions(
-      actions.map((it) => ({
-        itemId: it.target.identifier.itemId,
-        action: it.action,
-      })),
-    );
-    selectedRelatedActionsSetter(actions);
-  };
+  const setSelectedRelatedActions = useCallback(
```
taobojlen, line 340: are you adding useCallback in response to an observed performance issue, or is it hypothetical?

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -342,17 +337,18 @@ function ManualReviewJobReviewImpl(props: {
...
+    (actions: ManualReviewJobEnqueuedActionData[]) => {
+      actionStore?.setActions(
+        actions.map((it) => ({
+          itemId: it.target.identifier.itemId,
+          action: it.action,
+        })),
+      );
+      selectedRelatedActionsSetter(actions);
+    },
+    [actionStore],
```
taobojlen, line 350: should we not have `selectedRelatedActionsSetter` in the dependency array, too?

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -364,12 +360,12 @@ function ManualReviewJobReviewImpl(props: {
...
   const reportedUserRef = useRef<HTMLDivElement>(null);
 
-  const resetState = () => {
+  const resetState = useCallback(() => {
     setSelectedPrimaryActions([]);
     setSelectedPrimaryPolicies([]);
     setSelectedRelatedActions([]);
     setDecisionReason(undefined);
-  };
+  }, [setSelectedRelatedActions]);
```
taobojlen, line 368: why does this dependency array only have one of the functions?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 **Full Changelog**: https://github.com/roostorg/coop/compare/1.0.2...main
 
+## Review Console
```
taobojlen, line 5: (as in another PR) we should probably decide how we want to handle the changelog, i.e. writing to it per-PR or manually writing it all once in a while

cc @cassidyjames

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -1667,6 +1667,22 @@ export class ManualReviewToolService {
...
+    // reads this back and steps past it.
+    await this.queueOps.recordReviewerSkip({
+      orgId: opts.orgId,
+      queueId: opts.queueId,
+      reviewerId: opts.userId,
+      jobId: opts.jobId,
+    });
+    // Release the reviewer's lock (the lock token is the reviewer's userId)
+    // so the job returns to the shared pool immediately for everyone else.
+    await this.releaseJobLock({
```
taobojlen: are there any cases where we'd want to record a skip and *not* release the job lock? i can't think of any!

so if that's the case, can we call `releaseJobLock` from inside `recordReviewerSkip`? that way, callers won't have to remember to also release the lock manually each time they call it

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1304,47 +1306,104 @@ export default class QueueOperations {
...
+      // Release the held-aside jobs so other reviewers can pick them up
+      // immediately. This reviewer stays excluded via the skip set.
+      for (const held of heldAside) {
+        await this.releaseJobLock({
           orgId,
           queueId,
+          jobId: held.data.id,
           lockToken,
-          jobId: convertedJob.data.id,
         }).catch(() => {});
```
taobojlen: the silent error catching is a little worrying! seems like we should at the very least log that, but i think it's fine to raise an exception here.

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1304,47 +1306,104 @@ export default class QueueOperations {
...
-      } else {
-        // this is the most likely case, where there is a job
-        // and it has never been decided before
-        return { job: convertedJob.data, lockToken };
       }
     }
-    return null;
+  }
+
+  static readonly REVIEWER_SKIP_TTL_MS = 30 * 60 * 1000;
```
taobojlen, line 1492: hmm, how come we only skip for 30 minutes?

i guess in general -- what's the user story we imagine when someone skips a job? is it "i am not qualified to review this", "this is a hard job and i don't wanna deal with it / lower my average handling time by taking it", or something else? cc @julietshen

server/services/manualReviewToolService/modules/QueueOperations.reviewerSkips.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        mrtService: container.ManualReviewToolService,
+        userId: user.id,
+      });
+
+      return {
+        org,
+        queue,
+        user,
+        mrtService: container.ManualReviewToolService,
+        cleanup: async () => {
```
taobojlen: can you use the new transactional-backed test setup so we don't need to do the cleanup manually?

taobojlen (changes requested):

server/services/manualReviewToolService/modules/QueueOperations.ts
calebmcquaid: Changed it to log with the job context

server/services/manualReviewToolService/modules/QueueOperations.reviewerSkips.test.ts
calebmcquaid: Done

server/services/manualReviewToolService/manualReviewToolService.ts
calebmcquaid: This has been moved into `recordReviewerSkip`

taobojlen (changes requested):
Image

i think this needs a rebase 😅 this is too big to review and includes lots of unrelated changes!

github's native stacks feature might be useful!

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: `advanceToNextJobAfterInvalidation` calls `resetState` which calls `setSelectedRelatedActions`. `advanceToNextJobAfterInvalidation` is a `useCallback` that didn't have `resetState` in the deps array. I added it, removed the eslint disables.

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: We can, but I don't think it's needed. `selectedRelatedActionsSetter` is the setter in a `useState` and isn't caught by `exhaustive-deps`

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
calebmcquaid: Same deps answer here as above. `setSelectedRelatedActions` is the only non-useState setter used by `resetState`

Probably worth a refactor ticket to clean up the language standards at least in this file.

server/services/manualReviewToolService/modules/QueueOperations.ts
calebmcquaid: The user story is more "As a reviewer, I don't want to see this job right now", but it's not "never show me this job again." I was trying to prevent "stranded" jobs by everyone on a small team skipping the job and it never reappearing.

calebmcquaid:
> Image
> i think this needs a rebase 😅 this is too big to review and includes lots of unrelated changes!
> 
> github's native stacks feature might be useful!

rebase is completed. I will look into setting up a stack for these PRs

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -783,20 +777,21 @@ function ManualReviewJobReviewImpl(props: {
...
+        // Nothing was released or hidden; stay on the current job so the
+        // reviewer can retry (or decide) instead of advancing past it.
+        setModalInfo({
+          visible: true,
+          modalBody: 'Failed to skip this job. Please try again.',
+          footer: [{ title: 'Ok', type: 'primary', onClick: hideModal }],
+        });
+        return;
+      }
     }
```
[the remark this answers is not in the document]
calebmcquaid: fixed, stale

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1304,47 +1306,131 @@ export default class QueueOperations {
...
+        if (decision !== undefined) {
+          await this.removeJob({
+            orgId,
+            queueId,
+            lockToken,
+            jobId: convertedJob.data.id,
+          }).catch(() => {});
+          // then continue while loop
+        } else {
+          // this is the most likely case, where there is a job
```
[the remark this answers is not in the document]
calebmcquaid: `getNextJob` already pulled the job out of wait, so a failed `removeJob` can't hand it back to the scan

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1304,47 +1306,131 @@ export default class QueueOperations {
...
+        }
+      }
+    } finally {
+      // Release the held-aside jobs so other reviewers can pick them up
+      // immediately. This reviewer stays excluded via the skip set.
+      for (const held of heldAside) {
+        await this.releaseJobLock({
           orgId,
           queueId,
+          jobId: held.data.id,
```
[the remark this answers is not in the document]
calebmcquaid: fixed, stale

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -1661,6 +1661,14 @@ export class ManualReviewToolService {
     userId: string;
   }) {
     await this.skipOps.logSkip(opts);
+    // Hides the job from THIS reviewer for the skip window and releases their
+    // lock so it returns to the shared pool immediately for everyone else.
+    await this.queueOps.recordReviewerSkip({
```
taobojlen, line 1666: hmm, so this is a separate operation from `skipOps.logSkip` -- why? are there any circumstances where you'd want to call `skipOps` but *not* `queueOps.recordReviewerSkip`?

if not, then i suggest we instead modify `skipOps.logSkip` so that it also calls `queueOps.recordReviewerSkip` itself. that keeps the API smaller and cleaner. otherwise, you always have to remember to call both these methods together -- that's easy to forget!

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1373,7 +1375,12 @@ export default class QueueOperations {
           queueId,
           lockToken,
           jobId: job.data.id,
-        }).catch(() => {});
+        }).catch((error: unknown) => {
+          // Non-fatal: the scan should still hand this reviewer a job. But it
+          // must not be silent — see the note on the same call in
```
taobojlen, line 1380: best not to reference other code comments in code comments. very easy to get stale / out of sync.

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1397,47 +1404,139 @@ export default class QueueOperations {
...
-      if (hasDecision) {
-        await this.removeJob({
+        if (decision !== undefined) {
+          await this.removeJob({
+            orgId,
+            queueId,
+            lockToken,
+            jobId: convertedJob.data.id,
+          }).catch((error: unknown) => {
+            // Non-fatal, but not silent either. The common cause is another
```
taobojlen, line 1453: there's a bit of an LLM antipattern here in this comment, IMO.

it word-dumps a *lot* of detail into the comment, and most of it is sort of an extreme edge case -- if redis is returning failures, coop is pretty much broken anyway and we don't need special handling for each individual operation. as a reader of the code, it means i have to put all this complexity into my head, but with quite little return.

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1397,47 +1404,139 @@ export default class QueueOperations {
...
+          // this is the most likely case, where there is a job
+          // and it has never been decided before
+          return { job: convertedJob.data, lockToken };
+        }
+      }
+    } finally {
+      // Release the held-aside jobs so other reviewers can pick them up
+      // immediately. This reviewer stays excluded via the skip set.
+      for (const held of heldAside) {
+        // No `.catch` here on purpose: `releaseJobLock` never rejects — it
```
taobojlen, line 1476: i think this is probably over-eager error handling. why should `releaseJobLock` not reject/throw under some circumstances?

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1714,9 +1813,14 @@ export default class QueueOperations {
...
       await job.moveToDelayed(Date.now(), lockToken);
     } catch (error: unknown) {
-      // If the lock has already expired or the job is in a different state,
-      // we can safely ignore the error as the job is already released
-      // or will be handled by the stalled job checker
+      // Non-fatal: if the lock has already expired or the job moved state, the
+      // job is effectively released already (or the stalled checker will get
+      // it), and callers — including the `releaseJobLock` mutation and the
+      // held-aside loop in `dequeueNextJobWithLock` — treat releasing as
+      // best-effort cleanup that must not fail the operation around it.
```
taobojlen, line 1820: when an LLM says something is "best-effort" it should usually raise alarm bells a bit. they tend to add over-eager fallbacks for every single case, and all that does in the long run is lead to a system that's overly complex and unpredictable. again, i realize that here the code was pre-existing, just sharing this as a note. i don't know if the code comment is needed here -- future LLMs will see it and think that we *want* these kinds of best-effort and fallback code paths. but we don't!

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -783,33 +777,62 @@ function ManualReviewJobReviewImpl(props: {
   }, [jobId, queueId, refetchJobInfo, advanceToNextJobAfterInvalidation]);
 
   const skipToNextJob = async () => {
-    // First, release the lock on the current job and log the skip
+    // This is wired straight to a button's `onClick`, so nothing downstream
+    // catches a rejection: any error escaping here is an unhandled promise
+    // rejection and the reviewer gets no feedback at all. Both the skip and
+    // the follow-up dequeue therefore handle rejection explicitly.
+    const showSkipFailed = () =>
```
taobojlen, line 784: why would a skip fail? if it does fail, is retrying the best approach for the user? do we really need special handling for this case? (my instinct is no!)

client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobReview.tsx
```diff
@@ -783,33 +777,62 @@ function ManualReviewJobReviewImpl(props: {
...
-            },
-          },
-        }),
-      ]);
+      let skipped: boolean;
+      try {
+        const result = await logSkip();
+        skipped = result.data?.logSkip === true;
+      } catch {
+        // Network/GraphQL failure. Same reviewer-facing outcome as a falsy
```
taobojlen, line 800: i think this is overly complex. "Network/GraphQL failure" is vague. there are always a million things that can go wrong in a computer system -- but building in handling for every eventuality is an antipattern that does not lead to good products.

taobojlen (changes requested):

PR #901, opened 2026-07-13
fix: Replace bcrypt with Argon2id, upgrading legacy hashes on login

### LINKED ISSUE #900: Passwords are hashed with bcrypt at cost factor 5, below the OWASP minimum
### What happened

`hashPassword` in `server/services/userManagementService/utils.ts` hashes passwords with bcrypt at cost factor **5**.

Why this is a problem: bcrypt's cost factor is the only thing that makes offline brute-forcing of a leaked `users.password` column expensive, and it scales exponentially — each +1 doubles the work per guess. [OWASP's minimum recommendation is 10](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#bcrypt); at cost 5 an attacker who obtains the password column (SQL injection, stolen backup, exposed replica) can run a dictionary attack roughly **32× faster** than against the OWASP floor, putting every weak-to-moderate password within practical cracking range.

### Agreed resolution

Rather than raising the bcrypt cost factor — which would churn the hash format now and again later — move straight to **Argon2id**, per [OWASP's current recommendation](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id): a minimum of 19 MiB of memory, 2 iterations, and 1 degree of parallelism. (Discussed in #901.)

Note that no migration is possible: re-hashing requires the plaintext password, which is only ever in memory for the instant of a login. Existing bcrypt hashes must therefore keep verifying indefinitely, and are upgraded opportunistically after a successful password check.

### Version/commit

_No response_

### Affected browser(s)

_No response_

### Relevant logs/errors

```shell

```

### Anything else

References:

- https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id
- https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#bcrypt
- https://soatok.blog/2022/12/29/what-we-do-in-the-etc-shadow-cryptography-with-passwords/
- https://soatok.blog/2024/11/27/beyond-bcrypt/

### Screenshots

_No response_

### Checklist

serendipty01:
## Context

Closes #900.

`hashPassword` was minting bcrypt at cost factor **5**. This PR originally raised it to 12; per review discussion below, it now moves straight to **Argon2id** so the hash format only churns once.

Argon2id at [OWASP's recommended minimum](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id): 19 MiB of memory (`m=19456` KiB), `t=2`, `p=1`. The parameters are encoded in every hash, so a future bump needs no flag day.

Re-hashing requires the plaintext password, which we only hold for the instant of a login — so there is no migration, only an opportunistic upgrade:

- **New hashes** (signup, password change/reset, `create-org`) are Argon2id.
- **Existing bcrypt hashes keep verifying.** `bcryptjs` therefore stays a permanent dependency: an account that never logs in again is never re-hashed.
- **Rehash-on-login:** after a *successful* password check, a stale hash is transparently re-minted and persisted. Best-effort — a failed write is logged, never fails the login, and the row is retried on the next login.

## For reviewers

1. **The rehash write is compare-and-swap** on the verified hash, not an update by user id. If a concurrent password change lands between verification and the write, zero rows match and the stale rehash is dropped instead of resurrecting the old password.

2. **It is not the transactional path.** `changePassword` / `resetPasswordForToken` also purge the user's other sessions (#778). A rehash is not a password change; reusing that path would log users out of their other sessions on their first login after deploy.

3. **`passwordNeedsRehash` compares `v`/`m`/`t`/`p` for strict equality**, so hashes converge on exactly one configuration rather than leaving a tail of stronger-but-different ones. This mirrors [`needsRehash` in the reference `argon2` package](https://github.com/ranisalt/node-argon2/blob/60e11ef/argon2.cjs#L134-L161) — except that one doesn't check the algorithm variant, so an `argon2i` hash with matching parameters slips past it. Ours is anchored on `$argon2id$`. Any bcrypt hash is flagged regardless of cost: cost is no longer the criterion, bcrypt is.

4. **Corrupt input: one path resolves `false`, the other surfaces as an error — both logged.** `argon2Verify` is inconsistent on corrupt input — some malformed digests resolve `false` (an ordinary wrong-password result), others throw. A throw is not special-cased into "incorrect password": it propagates to `verifyEmailPasswordCredentials`'s existing catch-all, which logs it via the tracer and returns a generic internal-server error instead (per review discussion — no handling was added for a case nobody could explain how to reach). Without that log, a total Argon2 failure (the 19 MiB allocation failing under memory pressure affects *every* login) would be indistinguishable from a flood of users mistyping their passwords.

## Dependency: `@node-rs/argon2` (MIT) — approval requested

[`argon2`](https://npmx.dev/package/argon2) is the more obvious pick, but it is node-gyp native and our server image (`node:24.14.1-bullseye-slim`) installs only `git` — no `python3`/`make`/`g++` — so it would depend entirely on a prebuilt binary resolving, with no fallback. [`@node-rs/argon2`](https://npmx.dev/package/@node-rs/argon2) ships per-platform binaries as optional deps and has no install script. Same Argon2id, same parameters, interchangeable hashes.

Verified rather than assumed: `docker compose build backend` → `npm ci` succeeds in the image and the binding loads and hashes inside it; the lockfile carries all 15 platform binaries, including `linux-x64-gnu` for CI.

Trade-off: it exports no `needsRehash`, so the ~12 lines that parse the encoded parameters are ours.

Node 24 also ships an experimental [`crypto.argon2`](https://nodejs.org/docs/latest-v24.x/api/crypto.html#cryptoargon2algorithm-parameters-callback) — worth dropping the dependency for once it stabilises.

## Tests

15 unit tests across `server/services/userManagementService/utils.test.ts` (parameters, round-trip, legacy bcrypt verification, corrupt-input handling, `passwordNeedsRehash` across bcrypt / parameter drift / wrong variant) and `server/graphql/datasources/userApiCredentials.test.ts` (rehash persisted and CAS-scoped, no write when current, no write on wrong password, corrupt hash surfaces a generic error and logs, failed rehash write still succeeds the login).

Manual, on a local instance against real Postgres:

| Scenario | Result |
| --- | --- |
| `create-org` admin user | minted `$argon2id$v=19$m=19456,t=2,p=1$` |
| Login against a legacy `$2a$05$` hash | succeeds; row upgraded to Argon2id |
| Login against a `$2a$12$` hash | succeeds; also upgraded |
| Second login on a current Argon2id hash | succeeds; no rewrite (hash and `updated_at` unchanged) |
| Wrong password | `LoginIncorrectPasswordError`, no write |
| Corrupt hash | generic internal-server error, logged, no write |

## Rollout

Nothing required: no migration, no new env vars, no API/UI change. `public.users.password` is `varchar(255)` and an encoded Argon2id hash is 97 characters. Legacy hashes upgrade lazily as users log in.

Expected side effect: the password hashing paths (login, signup, password change) get deliberately slower — that is the point, and it affects only those endpoints.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * New passwords are now hashed with Argon2id.
  * Existing bcrypt passwords continue to work during sign-in.
  * After a successful login, outdated password hashes are opportunistically upgraded.
* **Bug Fixes**
  * Password upgrades only run after verification succeeds and the stored password is still the expected legacy value.
  * Wrong passwords no longer trigger any upgrade attempts.
  * If an upgrade write fails, sign-in still succeeds while the failure is recorded.
  * Unverifiable stored hashes are handled safely (“fails closed”).
* **Tests**
  * Added Jest coverage for hashing, verification (including legacy support), rehash decisioning, and login upgrade scenarios.

### DIFF
--- a/server/graphql/datasources/userApiCredentials.test.ts
+++ b/server/graphql/datasources/userApiCredentials.test.ts
@@ -82,7 +82,7 @@ function makeDeps(kyselyPg: Kysely<unknown>) {
 describe('verifyEmailPasswordCredentials', () => {
   const password = 'correct horse battery staple';
 
-  it('rehashes and persists a legacy work-factor-5 hash on successful login', async () => {
+  it('rehashes a legacy bcrypt hash to Argon2id on successful login', async () => {
     const legacyBcryptHash = await bcrypt.hash(password, 5);
     const userRow = makeUserRow({ password: legacyBcryptHash });
     const { kyselyPg, updateTable, updateBuilder } = makeMockKyselyPg({
@@ -99,7 +99,9 @@ describe('verifyEmailPasswordCredentials', () => {
     expect(result.id).toBe('user-123');
     expect(updateTable).toHaveBeenCalledWith('public.users');
     const [[persistedPatch]] = updateBuilder.set.mock.calls;
-    expect(persistedPatch.password).toMatch(/^\$2[aby]\$12\$/);
+    expect(persistedPatch.password).toMatch(
+      /^\$argon2id\$v=19\$m=19456,t=2,p=1\$/,
+    );
     // Compare-and-swap guard (PR #901 review): the write must be scoped to
     // the exact hash that was just verified, not the user id alone, so a
     // concurrent password change can never be clobbered by a rehash of the
@@ -117,7 +119,7 @@ describe('verifyEmailPasswordCredentials', () => {
     ).resolves.toBe(true);
   });
 
-  it('does not rehash when the stored hash is already at the target work factor', async () => {
+  it('does not rehash when the stored hash is already a current Argon2id hash', async () => {
     const currentHash = await hashPassword(password);
     const userRow = makeUserRow({ password: currentHash });
     const { kyselyPg, updateTable } = makeMockKyselyPg({ userRow });
@@ -138,6 +140,28 @@ describe('verifyEmailPasswordCredentials', () => {
       verifyEmailPasswordCredentials(deps, 'test@example.com', 'wrong'),
     ).rejects.toThrow();
 
+    expect(updateTable).not.toHaveBeenCalled();
+  });
+
+  it('fails closed and logs when the stored hash cannot be evaluated', async () => {
+    // A corrupt row, or Argon2 failing operationally (the 19 MiB allocation
+    // can fail under memory pressure, taking down *every* login). Both throw
+    // out of `passwordMatchesHash`. The user-facing answer is still "wrong
+    // password" — but it must be logged, or a total verification outage is
+    // indistinguishable from a flood of users mistyping their passwords.
+    // A bad base64 salt is one of the shapes `argon2Verify` throws on rather
+    // than resolving false — see the corrupt-input cases in `utils.test.ts`.
+    const userRow = makeUserRow({
+      password: '$argon2id$v=19$m=19456,t=2,p=1$!!!!$!!!!',
+    });
+    const { kyselyPg, updateTable } = makeMockKyselyPg({ userRow });
+    const { deps, logActiveSpanFailedIfAny } = makeDeps(kyselyPg);
+
+    await expect(
+      verifyEmailPasswordCredentials(deps, 'test@example.com', password),
+    ).rejects.toThrow();
+
+    expect(logActiveSpanFailedIfAny).toHaveBeenCalled();
     expect(updateTable).not.toHaveBeenCalled();
   });
 
--- a/server/graphql/datasources/userApiCredentials.test.ts
+++ b/server/graphql/datasources/userApiCredentials.test.ts
@@ -1,0 +1,156 @@
+import bcrypt from 'bcryptjs';
+import { type Kysely } from 'kysely';
+
+import {
+  hashPassword,
+  passwordMatchesHash,
+} from '../../services/userManagementService/index.js';
+import { verifyEmailPasswordCredentials } from './userApiCredentials.js';
+
+// `verifyEmailPasswordCredentials` only touches the injected Kysely instance
+// (user lookup + the rehash-on-login update); `orgSettingsService` and
+// `tracer` are separate mocks below.
+function makeMockKyselyPg(opts: {
+  userRow: Record<string, unknown> | undefined;
+  updateShouldThrow?: boolean;
+}) {
+  const selectExecuteTakeFirst = jest.fn().mockResolvedValue(opts.userRow);
+  const selectBuilder = {
+    select: jest.fn().mockReturnThis(),
+    where: jest.fn().mockReturnThis(),
+    executeTakeFirst: selectExecuteTakeFirst,
+  };
+  const selectFrom = jest.fn().mockReturnValue(selectBuilder);
+
+  const updateExecuteTakeFirst = jest.fn();
+  if (opts.updateShouldThrow) {
+    updateExecuteTakeFirst.mockRejectedValue(new Error('update failed'));
+  } else {
+    updateExecuteTakeFirst.mockResolvedValue(opts.userRow);
+  }
+  const updateBuilder = {
+    set: jest.fn().mockReturnThis(),
+    where: jest.fn().mockReturnThis(),
+    returning: jest.fn().mockReturnThis(),
+    executeTakeFirst: updateExecuteTakeFirst,
+  };
+  const updateTable = jest.fn().mockReturnValue(updateBuilder);
+
+  const kyselyPg = {
+    selectFrom,
+    updateTable,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- test stub
+  } as unknown as Kysely<any>;
+
+  return { kyselyPg, selectFrom, updateTable, updateBuilder };
+}
+
+function makeUserRow(overrides: Partial<Record<string, unknown>> = {}) {
+  return {
+    id: 'user-123',
+    email: 'test@example.com',
+    password: 'placeholder',
+    first_name: 'Test',
+    last_name: 'User',
+    role: 'ADMIN',
+    approved_by_admin: true,
+    rejected_by_admin: false,
+    login_methods: ['password'],
+    permissions: [],
+    created_at: new Date(),
+    updated_at: new Date(),
+    org_id: 'org-456',
+    ...overrides,
+  };
+}
+
+function makeDeps(kyselyPg: Kysely<unknown>) {
+  const getSamlSettings = jest.fn().mockResolvedValue(null);
+  const logActiveSpanFailedIfAny = jest.fn();
+  return {
+    deps: {
+      kyselyPg,
+      orgSettingsService: { getSamlSettings } as unknown as never,
+      tracer: { logActiveSpanFailedIfAny } as unknown as never,
+    },
+    getSamlSettings,
+    logActiveSpanFailedIfAny,
+  };
+}
+
+describe('verifyEmailPasswordCredentials', () => {
+  const password = 'correct horse battery staple';
+
+  it('rehashes and persists a legacy work-factor-5 hash on successful login', async () => {
+    const legacyBcryptHash = await bcrypt.hash(password, 5);
+    const userRow = makeUserRow({ password: legacyBcryptHash });
+    const { kyselyPg, updateTable, updateBuilder } = makeMockKyselyPg({
+      userRow,
+    });
+    const { deps } = makeDeps(kyselyPg);
+
+    const result = await verifyEmailPasswordCredentials(
+      deps,
+      'test@example.com',
+      password,
+    );
+
+    expect(result.id).toBe('user-123');
+    expect(updateTable).toHaveBeenCalledWith('public.users');
+    const [[persistedPatch]] = updateBuilder.set.mock.calls;
+    expect(persistedPatch.password).toMatch(/^\$2[aby]\$12\$/);
+    // Verified through `passwordMatchesHash` — the same path a real login
+    // takes — rather than reimplementing the comparison in the test.
+    await expect(
+      passwordMatchesHash(password, persistedPatch.password),
+    ).resolves.toBe(true);
+  });
+
+  it('does not rehash when the stored hash is already at the target work factor', async () => {
+    const currentHash = await hashPassword(password);
+    const userRow = makeUserRow({ password: currentHash });
+    const { kyselyPg, updateTable } = makeMockKyselyPg({ userRow });
+    const { deps } = makeDeps(kyselyPg);
+
+    await verifyEmailPasswordCredentials(deps, 'test@example.com', password);
+
+    expect(updateTable).not.toHaveBeenCalled();
+  });
+
+  it('rejects a wrong password and does not attempt a rehash write', async () => {
+    const legacyBcryptHash = await bcrypt.hash(password, 5);
+    const userRow = makeUserRow({ password: legacyBcryptHash });
+    const { kyselyPg, updateTable } = makeMockKyselyPg({ userRow });
+    const { deps } = makeDeps(kyselyPg);
+
+    await expect(
+      verifyEmailPasswordCredentials(deps, 'test@example.com', 'wrong'),
+    ).rejects.toThrow();
+
+    expect(updateTable).not.toHaveBeenCalled();
+  });
+
+  it('still succeeds the login when the rehash write fails (#778 regression guard)', async () => {
+    const legacyBcryptHash = await bcrypt.hash(password, 5);
+    const userRow = makeUserRow({ password: legacyBcryptHash });
+    const { kyselyPg } = makeMockKyselyPg({
+      userRow,
+      updateShouldThrow: true,
+    });
+    const { deps, logActiveSpanFailedIfAny } = makeDeps(kyselyPg);
+
+    // The login itself must succeed even though the opportunistic rehash
+    // write throws — the session must never be sacrificed for a background
+    // hash upgrade. This is exactly the failure mode the transactional
+    // session-purging update paths (`changePassword`, `resetPasswordForToken`,
+    // #778) would trigger if reused here.
+    const result = await verifyEmailPasswordCredentials(
+      deps,
+      'test@example.com',
+      password,
+    );
+
+    expect(result.id).toBe('user-123');
+    expect(logActiveSpanFailedIfAny).toHaveBeenCalled();
+  });
+});
--- a/server/graphql/datasources/userApiCredentials.test.ts
+++ b/server/graphql/datasources/userApiCredentials.test.ts
@@ -22,17 +22,18 @@ function makeMockKyselyPg(opts: {
   };
   const selectFrom = jest.fn().mockReturnValue(selectBuilder);
 
-  const updateExecuteTakeFirst = jest.fn();
+  const updateExecute = jest.fn();
   if (opts.updateShouldThrow) {
-    updateExecuteTakeFirst.mockRejectedValue(new Error('update failed'));
+    updateExecute.mockRejectedValue(new Error('update failed'));
   } else {
-    updateExecuteTakeFirst.mockResolvedValue(opts.userRow);
+    // Kysely's `execute()` on an update resolves to UpdateResult[]; the
+    // rehash path ignores it (zero matched rows = lost the CAS, no-op).
+    updateExecute.mockResolvedValue([]);
   }
   const updateBuilder = {
     set: jest.fn().mockReturnThis(),
     where: jest.fn().mockReturnThis(),
-    returning: jest.fn().mockReturnThis(),
-    executeTakeFirst: updateExecuteTakeFirst,
+    execute: updateExecute,
   };
   const updateTable = jest.fn().mockReturnValue(updateBuilder);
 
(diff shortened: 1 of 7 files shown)


server/graphql/datasources/userApiCredentials.ts
```diff
@@ -8,9 +12,39 @@ import {
...
+  user: GraphQLUserParent,
+  plaintextPassword: string,
+): Promise<void> {
+  try {
+    const rehashed = await hashPassword(plaintextPassword);
+    await kyselyUserUpdate(deps.kyselyPg, user.id, { password: rehashed });
+  } catch (e) {
+    deps.tracer.logActiveSpanFailedIfAny(e);
+  }
+}
```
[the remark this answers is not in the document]
serendipty01: fixed

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -8,9 +12,39 @@ import {
...
+  },
+  user: GraphQLUserParent,
+  plaintextPassword: string,
+): Promise<void> {
+  try {
+    const rehashed = await hashPassword(plaintextPassword);
+    await kyselyUserUpdate(deps.kyselyPg, user.id, { password: rehashed });
+  } catch (e) {
+    deps.tracer.logActiveSpanFailedIfAny(e);
+  }
```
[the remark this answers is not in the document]
serendipty01: Good catch — fixed in acbd76d2. The rehash write is now a compare-and-swap
  scoped to the exact hash that was just verified
  (`WHERE id = ? AND password = <verified hash>`), so if a concurrent
  changePassword / resetPasswordForToken commits in between, zero rows match
  and the stale rehash is dropped instead of clobbering the newer hash.
  Best-effort semantics are preserved: a lost CAS is a silent no-op (nothing
  to redo — the competing write already left a cost-12 hash), and the test
  now pins both WHERE clauses so the guard can't be silently removed later.

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -8,9 +12,39 @@ import {
 } from './userApiErrors.js';
 import {
   kyselyUserFindByEmail,
+  kyselyUserUpdate,
   type GraphQLUserParent,
 } from './userKyselyPersistence.js';
```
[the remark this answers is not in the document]
serendipty01: done

taobojlen (changes requested):
thanks for this!

i think the rehash-on-login approach is sensible. but it's a good question about argon2. i think that if we're going to change our password hashes, it makes most sense to move to the OWASP recommendation of

> Use [Argon2id](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) with a minimum configuration of 19 MiB of memory, an iteration count of 2, and 1 degree of parallelism.

otherwise we'll have to maintain several different hash formats all at once; seems we might as well do it in one go! what do you think?

serendipty01:
hey — agreed, let's do it in one go. i'll add the argon2id changes to this PR rather than opening a new one.

I was thinking to use [@node-rs/argon2](https://npmx.dev/package/@node-rs/argon2) (MIT)

Reasoning from Claude regarding this: 
> the more obvious pick is [argon2](https://npmx.dev/package/argon2), but it's a node-gyp native module and our server image (node:24.14.1-bullseye-slim) only installs git — no python3/make/g++ — so it'd be relying on a prebuilt binary resolving for linux/glibc, with no fallback if one doesn't. @node-rs/argon2 ships per-platform prebuilts as optional deps and skips node-gyp entirely, so it avoids that failure mode. same argon2id under the hood, same OWASP params. 

let me know if you'd rather use something else

node also has [crypto.argon2](https://nodejs.org/docs/latest-v24.x/api/crypto.html#cryptoargon2algorithm-parameters-callback) now, but it's still experimental, probably worth dropping the dependency for it once it stabilises.

i was also thinking about adding a pepper, and will raise a separate issue to discuss that.

marking this as draft for now; i'll ping you once it's done and tested on my end.

serendipty01:
@coderabbitai review

server/services/userManagementService/utils.ts
```diff
@@ -1,18 +1,84 @@
...
+  }
+  try {
+    return await argon2Verify(hash, givenPassword);
+  } catch {
+    // A malformed `$argon2`-prefixed hash can't authenticate anyone. Fail it
+    // closed as a non-match rather than letting the throw escape: the caller
+    // turns an exception into a 500, but the user-facing truth is simply that
+    // this password does not match.
+    return false;
+  }
```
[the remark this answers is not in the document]
serendipty01: fixed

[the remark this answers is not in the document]
serendipty01: fixed

server/graphql/datasources/userApiCredentials.test.ts
```diff
@@ -0,0 +1,191 @@
...
+      password,
+    );
+
+    expect(result.id).toBe('user-123');
+    expect(updateTable).toHaveBeenCalledWith('public.users');
+    const [[persistedPatch]] = updateBuilder.set.mock.calls;
+    expect(persistedPatch.password).toMatch(
+      /^\$argon2id\$v=19\$m=19456,t=2,p=1\$/,
+    );
+    // Compare-and-swap guard (PR #901 review): the write must be scoped to
```
taobojlen:
```suggestion
    // Compare-and-swap guard: the write must be scoped to
```

server/graphql/datasources/userApiCredentials.test.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    await expect(
+      verifyEmailPasswordCredentials(deps, 'test@example.com', password),
+    ).rejects.toThrow();
+
+    expect(logActiveSpanFailedIfAny).toHaveBeenCalled();
+    expect(updateTable).not.toHaveBeenCalled();
+  });
+
+  it('still succeeds the login when the rehash write fails (#778 regression guard)', async () => {
```
taobojlen: i don't really understand this test!

under what circumstances would the rewrite throw an exception, and how is that related to a user changing their password?

server/services/userManagementService/utils.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    // a silent `false` would make an operational Argon2 failure (the 19 MiB
+    // allocation failing under memory pressure takes down *every* login) look
+    // identical to one user mistyping their password. The login path catches
+    // it, logs it, and fails closed — see `userApiCredentials.test.ts`.
+    await expect(
+      passwordMatchesHash(PASSWORD, '$argon2id$v=19$m=19456,t=2,p=1$!!!!$!!!!'),
+    ).rejects.toThrow();
+
+    await expect(
+      passwordMatchesHash(PASSWORD, '$argon2id$v=19$corrupt'),
```
taobojlen: i think this should probably throw, too? i can't see why we'd want to handle this case elegantly; it should not come up in practice.

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -11,6 +15,52 @@ import {
...
 
+/**
+ * Best-effort upgrade of a stale password hash — a legacy bcrypt one, or an
```
taobojlen: I think we can simplify this comment; it's somewhat verbose.

We won't have any other Argon2 hashes so we don't need to mention any special handling there. I also don't think we need to explain why this is a plain column writ and not the `changePassword` path -- readers won't be wondering about that.

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -11,6 +15,52 @@ import {
...
+  try {
+    const rehashed = await hashPassword(plaintextPassword);
+    await deps.kyselyPg
+      .updateTable('public.users')
+      .set({ password: rehashed, updated_at: new Date() })
+      .where('id', '=', userId)
+      .where('password', '=', verifiedHash)
+      .execute();
+  } catch (e) {
+    deps.tracer.logActiveSpanFailedIfAny(e);
```
taobojlen: Under what circumstances would we hit this?

I am generally against swallowing errors unless they're actually expected.

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -60,13 +110,32 @@ export async function verifyEmailPasswordCredentials(
 
     // `loginMethods` includes 'password', so the DB CHECK constraint
     // guarantees `user.password` is non-null here.
-    if (
-      user.password == null ||
-      !(await passwordMatchesHash(password, user.password))
-    ) {
+    //
+    // `passwordMatchesHash` throws if the stored hash can't be evaluated at
```
taobojlen: let's not add special handling for this. i don't see how that would happen.

taobojlen (changes requested):
Thank you, this looks great!

I'd like to see us simplify this a little bit -- LLMs love to write overly-defensive code, but IMO it just complicates things and often isn't necessary. But this is very close!

server/graphql/datasources/userApiCredentials.test.ts
serendipty01: done

server/graphql/datasources/userApiCredentials.ts
serendipty01: updated the comment

server/graphql/datasources/userApiCredentials.test.ts
serendipty01: fixed the comment. The test covers: the rehash-on-login write can fail for any transient reason (DB connection drop, timeout, etc.), and the login must still succeed when that happens.

server/graphql/datasources/userApiCredentials.ts
serendipty01: yes, removed the try/catch. now the outer try/catch will handle that in `verifyEmailPasswordCredentials`

server/services/userManagementService/utils.test.ts
serendipty01: This test exists only to pin the library's current behavior, so a future `@node-rs/argon2` upgrade that changes it trips a test instead of silently drifting. Happy to delete it if you'd rather not carry that insurance — but there's no handling on our side to remove.

server/services/userManagementService/utils.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+});
+
+describe('passwordNeedsRehash', () => {
+  it('flags any bcrypt hash, regardless of cost factor', async () => {
+    // Cost is no longer the question — bcrypt itself is. Both a weak legacy
+    // hash and one at a strong cost factor are upgraded to Argon2id on login.
+    expect(passwordNeedsRehash(LEGACY_BCRYPT_HASH)).toBe(true);
+    expect(passwordNeedsRehash(await bcrypt.hash(PASSWORD, 12))).toBe(true);
+  });
```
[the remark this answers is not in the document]
serendipty01: fixed

server/graphql/datasources/userApiCredentials.ts
serendipty01: Some scenarios: the UPDATE hitting a connection-pool limit, a timeout, or a deadlock, or hashPassword failing under memory pressure (Argon2 needs a 19 MiB allocation per call; if the process is memory-starved that can throw). 

They should not fail the login and will only log them.

server/graphql/datasources/userApiCredentials.ts
```diff
@@ -60,13 +107,19 @@ export async function verifyEmailPasswordCredentials(
...
-      !(await passwordMatchesHash(password, user.password))
-    ) {
+    if (user.password == null) {
       throw makeLoginIncorrectPasswordError({ shouldErrorSpan: true });
     }
 
+    const passwordMatches = await passwordMatchesHash(password, user.password);
+    if (!passwordMatches) {
+      throw makeLoginIncorrectPasswordError({ shouldErrorSpan: true });
+    }
```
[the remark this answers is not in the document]
serendipty01: This is intentional, not an oversight — see [taobojlen's comment above](https://github.com/roostorg/coop/pull/901#discussion_r3580207266), which asked to drop the special-casing that disguised this as "incorrect password." Going with that: a corrupt/unevaluable hash now surfaces as a generic internal-server error via the existing outer catch-all in verifyEmailPasswordCredentials, and is still logged via the tracer either way. I've updated the PR description (point 4) and the manual-test table to reflect this instead of the old "fail closed as non-match" language.

server/graphql/datasources/userApiCredentials.test.ts
taobojlen: > the rehash-on-login write can fail for any transient reason (DB connection drop, timeout, etc.), and the login must still succeed when that happens.

i think i fundamentally disagree with this! in all of these above cases, the login should fail.

server/services/userManagementService/utils.test.ts
taobojlen: let's remove it. this behaviour actually would cause the silent drift, we want to fail loudly if our fundamental assumptions get violated!

server/graphql/datasources/userApiCredentials.ts
taobojlen: all of those things should make the login fail! the user can just retry if it's truly transient but let's not have all sorts of silent behaviors in our code.

server/graphql/datasources/userApiCredentials.test.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    await expect(
+      verifyEmailPasswordCredentials(deps, 'test@example.com', 'wrong'),
+    ).rejects.toThrow();
+
+    expect(updateTable).not.toHaveBeenCalled();
+  });
+
+  it('surfaces a generic internal-server error and logs when the stored hash cannot be evaluated', async () => {
+    // A corrupt row, or Argon2 failing operationally (the 19 MiB allocation
```
taobojlen: if we don't have 19MiB of memory available then we have bigger problems than users not being able to log in!

server/graphql/datasources/userApiCredentials.test.ts
```diff
@@ -0,0 +1,194 @@
...
+  it('still succeeds the login when the rehash write throws', async () => {
+    const legacyBcryptHash = await bcrypt.hash(password, 5);
+    const userRow = makeUserRow({ password: legacyBcryptHash });
+    const { kyselyPg } = makeMockKyselyPg({
+      userRow,
+      updateShouldThrow: true,
+    });
+    const { deps, logActiveSpanFailedIfAny } = makeDeps(kyselyPg);
+
+    // `rehashPasswordOnLogin` swallows any error from the write — e.g. a
```
taobojlen: let's simplify this. we can just fail loudly in this case. no need for the added complexity!

server/graphql/datasources/userApiCredentials.ts
taobojlen: i.e. let's remove the try/catch here.

server/graphql/datasources/userApiCredentials.ts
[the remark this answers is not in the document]
taobojlen: yeah this is a bad suggestion from copilot!

taobojlen (changes requested):

server/services/userManagementService/utils.test.ts
serendipty01: removed

server/graphql/datasources/userApiCredentials.test.ts
serendipty01: Removed along with the try/catch — folded into the single rewritten "fails the login when the rehash write throws" test.

server/graphql/datasources/userApiCredentials.test.ts
serendipty01: Simplified to one test asserting the throw, per your suggestion.

AGENTS.md
```diff
@@ -165,6 +165,8 @@ Note: `check_migration_order` runs only in GitHub Actions — it's GitHub-specif
 - **Naming:** Use camelCase for variables/functions; PascalCase for components/classes; SCREAMING_SNAKE_CASE for constants.
 - **GraphQL:** Type-safe resolvers and queries via codegen; never hand-edit `generated.ts`.
 - **Imports:** Absolute imports configured via `tsconfig.json` paths; prefer `@/` prefix over relative paths where configured.
+- **Errors: fail loudly.** Don't catch an error unless you can name the condition that produces it _and_ say why continuing is correct. A `try`/`catch` that logs and carries on isn't resilience — it turns an unknown fault into a silent one. "It might be transient" (connection-pool limits, timeouts, deadlocks, resource exhaustion) is not a justification: let the request fail and let the caller retry. If an assumption is violated, the code should say so, not paper over it. See the [PR #901 review](https://github.com/roostorg/coop/pull/901) for a worked example.
```
taobojlen:
```suggestion
- **Errors: fail loudly.** Don't catch an error unless you can name the condition that produces it _and_ say why continuing is correct. A `try`/`catch` that logs and carries on isn't resilience — it turns an unknown fault into a silent one. "It might be transient" (connection-pool limits, timeouts, deadlocks, resource exhaustion) is not a justification: let the request fail and let the caller retry. If an assumption is violated, the code should say so, not paper over it.
```

let's not reference individual github conversations in AGENTS.md IMO

AGENTS.md
```diff
@@ -165,6 +165,8 @@ Note: `check_migration_order` runs only in GitHub Actions — it's GitHub-specif
...
+- **Don't test the papering-over.** No tests that assert an error was swallowed, and no tests that pin a third-party library's behaviour on input that should never reach it — those lock in the silence they document. Assert the loud failure instead.
```
taobojlen: can you explain why this is a good principle to put in AGENTS.md? is it always true?

taobojlen (changes requested):
thank you! almost there, just some comments/questions on the AGENTS.md now :)

AGENTS.md
taobojlen: the reason is that we don't want an agent going off and reading a totally unrelated PR discussion, polluting its context window, when the general principle is already stated here.

AGENTS.md
serendipty01: sorry, got accidentally committed. 
i will remove the file

AGENTS.md
serendipty01: will remove the file

taobojlen (approved):
thanks so much for this! tested locally, looks great 👍

PR #903, opened 2026-07-13
fix(mrt): collapse long text fields with Read more (#870)

### LINKED ISSUE #870: Long text from users creates an overly long horzontal scroll
### Problem being solved

Coop is currently loading too much data on users, like for example text (thousands of lines) or thousands of images in the case of a NCMEC report, this may cause the tab to lag or to become unresponsive making the user unable to be actioned.

### Proposed solution

Adding a "Read more..." would help for text fields, and a pagination system for NCMEC (or items in general) would help

### Alternatives considered

N/A

### Anything else

_No response_

### Screenshots

_No response_

### Checklist

taobojlen:
Fix #870.

This is a suggested UI, but feedback is very welcome!

I am slightly unsure about this -- if any users are regularly reviewing long blocks of text, then this actually introduces an extra click which could influence average handling time. I thought that 2000 graphemes seemed like a reasonable cutoff before we show the "Read more" button but I'm not sure about this!

# Long text (collapsed)
Screenshot 2026-07-13 at 14-20-53 Review Job

# Long text (expanded)
Screenshot 2026-07-13 at 14-21-07 Review Job

# Shorter text (below the limit)
Screenshot 2026-07-13 at 14-08-22 Review Job

## Summary by CodeRabbit

* **New Features**
  * Long text values in manual review fields can now be expanded and collapsed with “Read more” and “Read less” controls.
  * Text limits correctly account for emojis and other multi-character symbols.
  * Media fields retain horizontal scrolling, while other content uses improved responsive layouts.

* **Bug Fixes**
  * Updated text state resets correctly when displayed content changes.
  * Improved handling of narrow layouts to prevent content overflow.

### DIFF
--- a/client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
@@ -1,0 +1,73 @@
+import { memo, useMemo, useState } from 'react';
+
+type CollapsibleTextProps = {
+  text: string;
+  maxLines?: number;
+  maxGraphemes?: number;
+};
+
+/** Count graphemes using Intl.Segmenter (handles all scripts correctly). */
+function countGraphemes(text: string): number {
+  const segmenter = new Intl.Segmenter(undefined, { granularity: 'grapheme' });
+  let count = 0;
+  for (const _ of segmenter.segment(text)) {
+    count += 1;
+  }
+  return count;
+}
+
+/**
+ * Renders text with wrapping. When the text exceeds `maxGraphemes`, it is
+ * collapsed to `maxLines` visible lines with a "Read more" / "Read less"
+ * toggle. Short text renders in full with no toggle.
+ *
+ * Memoized: props are primitives (string, number), so React.memo's shallow
+ * comparison prevents re-renders (and re-segmentation) when a parent re-renders
+ * but the text hasn't changed — important for thread histories with many copies
+ * of the same message.
+ */
+function CollapsibleTextImpl({
+  text,
+  maxLines = 12,
+  maxGraphemes = 2000,
+}: CollapsibleTextProps) {
+  const isCollapsible = useMemo(
+    () => countGraphemes(text) > maxGraphemes,
+    [text, maxGraphemes],
+  );
+  const [expanded, setExpanded] = useState(false);
+
+  if (!isCollapsible) {
+    return (
+      <div className="whitespace-normal break-words text-start">{text}</div>
+    );
+  }
+
+  return (
+    <div className="flex flex-col text-start">
+      <div
+        className="whitespace-normal break-words overflow-hidden"
+        // Inline style rather than Tailwind's line-clamp-N so the line count
+        // is fully dynamic (Tailwind's JIT won't generate line-clamp-${maxLines}
+        // from a template literal). This is exactly what line-clamp compiles to.
+        style={{
+          display: '-webkit-box',
+          WebkitBoxOrient: 'vertical',
+          WebkitLineClamp: expanded ? 'none' : maxLines,
+        }}
+      >
+        {text}
+      </div>
+      <button
+        type="button"
+        className="self-start pt-1 font-semibold text-blue-500 cursor-pointer"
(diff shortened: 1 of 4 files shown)


client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  if (!isCollapsible) {
+    return (
+      <div className="whitespace-normal break-words text-start">{text}</div>
+    );
+  }
+
+  return (
+    <div className="flex flex-col text-start">
+      <div
```
juanmrad, line 64: When "Read more" is clicked, `WebkitLineClamp` is set to 'none' and the text expands fully with no height constraint. For very long text (e.g. a 50KB paste), this pushes the rest of the review UI actions panel, submit button far off screen.

Consider wrapping the expanded state in a scrollable container with a max-height, e.g.:
```
style={{
  display: '-webkit-box',
  WebkitBoxOrient: 'vertical',
  WebkitLineClamp: expanded ? 'none' : maxLines,
  ...(expanded ? { maxHeight: '24rem', overflowY: 'auto' } : {}),
}}
```

That way expanding shows much more text in-place without pushing the page layout out of reach. The reviewer can scroll within the text block to read the rest.

client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Memoized: props are primitives (string, number), so React.memo's shallow
+ * comparison prevents re-renders (and re-segmentation) when a parent re-renders
+ * but the text hasn't changed — important for thread histories with many copies
+ * of the same message.
+ */
+function CollapsibleTextImpl({
+  text,
+  maxLines = 12,
+  maxGraphemes = 2000,
```
juanmrad, line 40: I'm unsure about the 2000 graphemes as the cutoff.

For context, 2000 graphemes is roughly 300–400 words or about a page of text. That seems borderline to me. May be sorta reasonable default a reviewer when it needs to scan content quickly, but also may jump already the page too much.

But the real question is whether the grapheme count is the right proxy at all. The user's pain is vertical overflow, which depends on line wrapping, container width, and font size, not character count. 

A short string with many newlines could be tall. And a long dense paragraph could be short. 

I do wonder if we should just have both height or length ( char count ), and we could even count also line breaks on the text.

client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
taobojlen: Yeah I was unsure about this! I don't personally love scrollable sub-containers, at least not if the main page itself is also scrollable like it is in Coop. Happy to change this if you prefer? Or we can ask for design feedback!

client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
taobojlen: Yeah that's very true! Do you have a preference for this? Maybe just line height is the most sensible, like 10+ lines and we collapse by default?

cassidyjames: Is this a case where it depends on the adopter, and should be yet-another-setting for admins? If so, should it be per-content-type, or is that just ballooning the complexity for little impact?

cassidyjames (approved):
Giving this a 👍🏻 design-wise, with the caveat of trying to figure out the exact length and amount of configurability. But UX-wise I agree this makes sense to pursue.

client/src/webpages/dashboard/mrt/manual_review_job/v2/components/CollapsibleText.tsx
taobojlen: I'd start with leaving it as-is and iterating based on feedback. If we get lots of conflicting feedback then we can always make it configurable!

PR #909, opened 2026-07-13
Fix Oldest Task Age showing the newest job's age instead of the oldest

dom-notion:
The "oldest task age" column in the MRT dashboard was actually showing the newest one. This fixes it to show the oldest one instead.

## Tests
Added a regression test for this.

## Checklist


julietshen (approved):
LGTM, thanks for the fix!

cc @calebmcquaid for vis just because he's been working on some queue ranking stuff as well

calebmcquaid (approved):
Thanks!

juanmrad (approved):
LGTM

PR #913, opened 2026-07-14
Fix for Manual Review Tool: If createdAt datetime is in bad format, store it as null in decision log

### LINKED ISSUE #912: Reviewers unable to skip Jobs
### What happened

We received a bug report from an adopter at 5pm ET on July 14th. Their team is unable to skip jobs and keeps getting this error:
 "Job submission failed. Please try again."

### Impact
- Affected jobs cannot be actioned at all. Every decision type (Ignore, custom action, NCMEC report, appeal accept/reject) routes through the same decision-logging code, so all of them fail.
- The affected job is stuck: the decision is not recorded and the job is not removed from the queue, so it keeps reappearing.
- No data corruption or partial writes. The insert fails before the job is removed, so Postgres and the queue stay consistent. The task is blocked, not lost.
- Scoped to items whose `createdAt` field holds a truthy but unparseable value. Items submitted through the normal API do not reach this state.

### Version/commit

1.0.2

### Affected browser(s)

Unsure

### Relevant logs/errors
The adopter reported seeing logs dealing with internal server errors and the bug seems to only be affecting reports reported by system.

### Screenshots
Image

### Anything else

An initial investigation found that this error is triggered when skipping Jobs when the job's item has a `createdAt` field value that is present but not a parseable date. 

### Root Cause (theorized)
In `server/services/manualReviewToolService/modules/JobDecisioning.ts`, `#logDecision` builds the value for the nullable `manual_review_decisions.item_created_at` column directly from the item's createdAt field:

```
const itemCreatedAt = itemCreatedAtField ? new Date(itemCreatedAtField) : null;
```

The guard only filters falsy values ("", null, 0). A non-empty malformed value (for example whitespace, an epoch integer as a string, or a non-ISO timestamp) produces an **Invalid Date**. The Postgres driver throws Invalid time value when serializing that for a timestamptz column, which rejects the whole manual_review_decisions insert. That maps to `RecordingJobDecisionFailedError`, which the client renders as "Job submission failed."

The DATETIME field handler validates dates at ingestion (fieldTypeHandlers.ts:191): a non-empty string that doesn't parse is rejected outright with "must contain a valid date string," and empty string becomes null. So a genuinely malformed date on a proper DATETIME field can never get stored in the first place, which is likely why nobody hit this bug this for months.

### Why did it show up now?
- The decision-insert code has been latent since the initial open-source release in February, but it only fires when a job carries an unparseable createdAt. That data condition is the trigger, and it is new.

- The normal item-submission path validates dates at ingestion: the DATETIME field handler (`services/itemProcessingService/fieldTypeHandlers.ts:191`) rejects any non-empty string that does not parse and coerces empty string to null. A malformed date on a proper DATETIME field therefore cannot be stored, which is why no adopter hit this until now.

- A bad createdAt can only reach #logDecision through a path that skips that validation. The two known ways:
  - 1. Items whose payload is assembled by a path that does not run the DATETIME coerce step (for example system-created reports), so whatever it writes into `createdAt passes` straight through and doesn't get caught by the guard.
  - 2. A `createdAt` field role mapped to a non-DATETIME field (for example STRING), where ingestion only checks that the value is a string and new Date(...) later fails on it.

### Repro Steps
1. Create an item type with a createdAt field role.
2. Get a job into a queue whose item createdAt is truthy but unparseable, via a path that bypasses DATETIME validation (for example a system-generated report, or a createdAt role mapped to a STRING field).
3. Open the job in the review tool and submit any decision.
4. Observe the 500 and the "Job submission failed." modal. The job remains in the queue.

### Proposed Fix
- **normalize the value at the write boundary in `#logDecision`**. If it does not parse to a valid date, store null instead of handing an Invalid Date to Postgres. The column is already nullable, so the decision records and the task clears. This makes decision-logging robust to a bad createdAt from any source.
- Follow-up (pending confirmation of the trigger source): run the report-generation path that skips validation through the same createdAt coercion the normal submission API uses, so the bad value is never stored in the first place.

### Checklist

Co-Authored-By: Claude Opus 4.8

julietshen:

Might fix #912  - this is also a latent bug that we should fix regardless. Details on the likely root cause are in the issue.

The items endpoint validates `Datetime` fields at intake (`makeDateString`, which parses via date-fns and stores the normalized `toISOString()` result), so a value ingested that way is always clean and would never fail here. The bad `createdAt` reaches the job through a path that skips that validation: in the enqueue action, `getFullItem`returns the rule's target item as-is when it already carries its data, rather than a validated stored copy. So an item whose raw `createdAt` is in a format `new Date()`cannot parse flows straight through. Only those items fail, which is why it is intermittent and has no obvious pattern. The adopter's schema is configured correctly (the field is `Datetime` with the `Created At` role), so this is not a misconfiguration on their side.

### Details about this change
- Normalizes at the write boundary: parse the value and store `null` when it is not a valid date. The column is already nullable, so the decision records and the task clears.
- Pulls the parsing into a small exported helper, `parseItemCreatedAt`, with a unit regression test (valid inputs pass through; empty and unparseable inputs return `null`).

This fix protects the save no matter where the bad date came from, so it does not depend on tracking down the exact source. 

**For reviewers:** this only changes the one place where decisions are saved. I want some feedback on this implementation though, is null the right approach if the date is unreadable? The alternative would be to fall back to the current time. I went with empty because the column allows it and it avoids making up a date. 

## Tests
Manually tested with garbage datestamp/gibberish, and then with proper formats.

## (Optional) Rollout Plan

N/A

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed “Job submission failed” when an item’s **Created At** value was blank or unparseable.
  * Added safer timestamp parsing so invalid inputs are recorded without creating serialization errors.
  * Continued to support valid ISO strings, epoch-millis values, and existing date values.
  * Added tracing/visibility when an unparseable **Created At** value is encountered.


julietshen:
I was able to reproduce this locally. 

I confirmed empirically:
  1. A job whose item carries an unparseable `createdAt` fails the decision insert with exactly the same error the adopter's seeing.
  2. Storing null (the fix) makes it succeed. The fix is validated end to end against a real database, not just in a unit test.
  3.  The items API validator normalizes accepted dates to clean ISO and rejects everything else at intake, so an item submitted through the items API cannot carry a decision-breaking createdAt. The partial-items and reporting API paths reject too. They all run the same date validation, which either normalizes a good date or rejects a bad one. 

So we DO have validation at ingestion point, but this PR provides an escape hatch for if they get into a weird state.

It worries me that we don't know what caused this, but it's possibly due to a backfill that messed with the createdAt format.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -75,6 +76,24 @@ type MRTJobAutoCloseReason =
...
+ */
+export function parseItemCreatedAt(
+  value: string | number | Date | null | undefined,
+): Date | null {
+  // Guard only null/undefined/empty-string; a numeric 0 is a valid epoch.
+  if (value == null || value === '') {
+    return null;
+  }
+  const parsed = new Date(value);
+  return isValidDate(parsed) ? parsed : null;
```
calebmcquaid, line 95: Can we put in some `try/catch` logic to log any invalid dates here?

calebmcquaid (approved):
Approved with 1 comment to not hold up merge. It would also probably be worth doing a quick search to find consumers of this function that sort/filter based on `createdAt`

PR #916, opened 2026-07-15
Fix "Something Went Wrong" when opening MRT jobs with an unparseable Created At

julietshen:

Client-side counterpart to #913. That PR fixed the server decision-log path so a bad `createdAt` records as `null` instead of failing submission. The React client still rendered the same value through unguarded `date-fns` calls, which throw `RangeError: Invalid time value` on an invalid date and unmount the view into the "Something Went Wrong" error boundary.

This adds `safeFormat` / `safeFormatDistanceToNow` helpers in `client/src/utils/time.ts` that fall back to `Unknown` for invalid dates (epoch 0 stays valid), and uses them on the three MRT render sites:

- `ReportInfoComponent.tsx` (report/appeal received time on the job detail view)
- `ManualReviewJobCommentSection.tsx` (comment timestamps)
- `ManualReviewQueueJobsPreview.tsx` (Created At column in the queue job list)

Note: per `client/AGENTS.md` this touches the MRT area; no GraphQL schema or `generated.ts` changes are involved.

## Tests

- New `client/src/utils/time.test.ts` covering valid dates, epoch, unparseable strings, null/undefined, and custom fallback (regression test for the throw).
- `npx vitest run src/utils/time.test.ts` passes; lint and `tsc --noEmit` clean on changed files.

## (Optional) Rollout Plan

None. Client-only rendering change.

## Checklist

- [x] **If you changed anything user-facing** (i.e. user interface or APIs): Did you update the CHANGELOG.md and related docs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ## Review Console
 
 - Fixed a "Job submission failed" error that prevented reviewers from clearing jobs whose item had an unparseable `Created At` value; the decision now records instead of failing (#913)
-- Fixed a "Something Went Wrong" page when opening a job or queue whose `Created At` value was unparseable; the affected date now shows `Unknown` instead of blanking the view (#913)
+- Fixed a "Something Went Wrong" page when opening a job or queue whose `Created At` value was unparseable; the affected date now shows `Unknown` instead of blanking the view (#916)
 
 # Coop 1.0.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Review Console
 
 - Fixed a "Job submission failed" error that prevented reviewers from clearing jobs whose item had an unparseable `Created At` value; the decision now records instead of failing (#913)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## Review Console
 
 - Fixed a "Job submission failed" error that prevented reviewers from clearing jobs whose item had an unparseable `Created At` value; the decision now records instead of failing (#913)
+- Fixed a "Something Went Wrong" page when opening a job or queue whose `Created At` value was unparseable; the affected date now shows `Unknown` instead of blanking the view (#913)
```
juanmrad: This changelog entry cites (#913), but that's the server decision-log fix. This client change is #916. Please update the citation so release notes stay accurate.

juanmrad (commented):
Per #912/#913, the unparseable value is the item's Created At DATETIME field, which is rendered via `parseDatetimeToReadableStringInCurrentTimeZone` in ManualReviewJobFieldsComponent not the three call sites updated here (job.createdAt / comment.createdAt).

Those sites are worth hardening, but this PR may not fix the stated "Something Went Wrong when opening a job" failure. Worth either also guarding the `parseDatetime*` helpers (or the FieldsComponent DATETIME path).

juanmrad (approved):
LGTM

PR #918, opened 2026-07-16
Make Scylla-backed features (item investigation & user strikes) optional

### LINKED ISSUE #917: Make Scylla optional so the server can run without item investigation & user strikes
### Problem being solved

The server currently **requires a Scylla cluster to boot**. Starting without the
`SCYLLA_*` env vars crashes at startup:
Missing env var SCYLLA_HOSTS

However, Scylla only backs two features : **item investigation** (item/user
history) and **user strikes** (repeat-offender counts/thresholds). Operators who
don't need those features, or who deploy on a platform with no managed Scylla
offering, still have to stand up and maintain a Scylla cluster just to start the
app. This is a significant operational burden for a dependency that may be
entirely unused.

### Proposed solution

Add an opt-out flag: `ITEM_INVESTIGATION_AND_STRIKES_ENABLED` (defaults to
`true`, so existing deployments are unaffected). When set to `false`, the server
runs without Scylla: the two dependent features degrade gracefully to no-ops
(history/strike reads return empty, writes are dropped) and no `SCYLLA_*` env
vars are required.

Implementation-wise, gate at the single dependency-injection chokepoint (the
`Scylla` factory in `iocContainer`) and return a `NoOpScylla` when disabled, so
the ~100+ existing call sites keep compiling and running unchanged. This mirrors
the existing `NoOpAnalyticsAdapter` pattern already used for the
analytics/warehouse feature.

I've opened a PR implementing this: sunilatlas/coop#<PR-number>
(typecheck + lint pass, 8 new unit tests, and I verified the server boots and the
full item-submission → rule → review-queue flow works end-to-end with Scylla off).
Attached SS as a proof that reviewing of video also plays fine.

### Alternatives considered

- **Keep Scylla mandatory**- rejected: forces operators to run infrastructure
  for features they may not use.
- **Two separate flags** (one per feature) - deferred: both features share the
  same single dependency (Scylla), so one flag matches the actual infra
  boundary and keeps configuration simple. Could be split later if needed.
- **Gate each of the ~100+ Scylla call sites individually** -  rejected: large,
  error-prone diff. Gating the single DI factory is far cleaner and lower-risk.
- **Naming as `SCYLLA_ENABLED`** - chose a feature-oriented name instead so the
  flag reflects the capability, not the datastore. Open to a different name if
  maintainers prefer a convention.

### Anything else

Backwards-compatible: default is `true`, so nothing changes for existing
deployments unless the flag is explicitly set.

One honest trade-off to flag: with the flag off, item-investigation and
user-strike data silently becomes empty (reads) / dropped (writes). Happy to add
a startup warning if strikes/investigation are configured while the flag is off,
if that's preferred.

### Screenshots

[image.png](https://github.com/user-attachments/assets/1eebeca2-bf37-49f7-b057-3241faed36d3)

### Checklist

sunilatlas:
## Summary

Closes #917

Currently the server **requires a Scylla cluster to boot** — starting without
`SCYLLA_*` env vars crashes at startup:
Missing env var SCYLLA_HOSTS

Scylla only backs two features — **item investigation** (item/user history) and
**user strikes** (repeat-offender counts/thresholds). Operators who don't need
those, or who deploy on a platform with no managed Scylla offering, currently
still have to stand up and maintain a Scylla cluster just to start the app.

This PR adds an opt-out flag, **`ITEM_INVESTIGATION_AND_STRIKES_ENABLED`**
(defaults to `true`), that lets the server run without Scylla.

## Approach

Rather than gate the ~100+ call sites that touch Scylla, the change gates at the
single DI chokepoint — the `Scylla` factory in `iocContainer` — and returns a
`NoOpScylla` when the flag is off. `NoOpScylla` extends `Scylla`, so every
consumer keeps compiling and running unchanged; it simply drops writes and
returns empty reads. This mirrors the existing `NoOpAnalyticsAdapter` pattern
already used for the analytics/warehouse feature.

When disabled:
- `connect()` / `close()` resolve (the item-processing worker's eager
  `await scylla.connect()` still succeeds)
- `insert()` resolves and drops the write
- `select()` returns `[]`, `selectStream()` yields nothing

## Behavioural impact (explicit)

With `ITEM_INVESTIGATION_AND_STRIKES_ENABLED=false`, the two dependent features
**degrade gracefully to no-ops**: item/user history views show nothing, and user
strike counts read as 0 (writes are dropped). This is documented in
`.env.example`. **Default is `true`, so existing deployments are unaffected.**

## Changes
- `server/scylla/noOpScylla.ts` (new) — the no-op implementation
- `server/scylla/noOpScylla.test.ts` (new) — unit tests for the no-op + the
  flag-parsing predicate (guards the default-enabled behaviour)
- `server/iocContainer/index.ts` — gate the `Scylla` factory behind the flag
- `server/.env.example` — document the flag

## Testing
- `npm run typecheck` — passes
- `npm run lint` (server) — passes
- New unit tests — 8/8 pass
- Manually ran the server with `ITEM_INVESTIGATION_AND_STRIKES_ENABLED=false`
  and **no** `SCYLLA_*` vars: boots cleanly, `/api/v1/ready` returns 200, and the
  full item-submission → rule → review-queue flow works end to end.

## Notes / open questions for maintainers
- Naming: I chose a feature-oriented name over `SCYLLA_ENABLED` so it reflects
  the capability, not the datastore. Happy to rename if you prefer a convention.
- Optionally, a startup warning could be emitted if strikes/investigation are
  configured while the flag is off — happy to add if desired.

## Summary by CodeRabbit

* **New Features**
  * Added an environment-driven feature toggle for Item Investigation and User Strikes (enabled by default).
  * When disabled, related Scylla-backed operations are safely short-circuited: reads return empty results, strike writes are dropped, and strike processing is skipped.
  * The toggle supports explicit false values (including whitespace/case variations).

* **Tests**
  * Added Jest coverage for flag parsing rules and the Scylla-disabled/no-op behavior (connect/close/select/selectStream).


sunilatlas:
Updated PR as per the comments by CodeRabbit

julietshen:
Thank you for your PR @sunilatlas! I think this makes sense and takes us a step towards making Scylla entirely optional. There's a Request for Comments on that here: https://github.com/roostorg/coop/discussions/889

This PR looks good, but there are just a few formatting issues (missing final newlines, import ordering) that would trip up our CI.

You can `npm run prettier:fix` from root to get those resolved.

Note for ROOSTers: we should add a feature-gate for the user strike feature. Even with this gate, the strike logic itself will keep running when the flag is off. `applyUserStrikeFromPublishedActions` fires on every action and will keep counts as a constant 0, so threshold enforcement gets unreliable vs cleanly disabled. A policy whose single application already meets the threshold fires every time while a policy applying 1 strike against a threshold of 3 will never fire.

julietshen (approved):
LGTM, just some mild formatting issues

sunilatlas:
Thanks for the review and approval @julietshen 
I have updated the PR with the formatting changes.

sunilatlas:
I looked into the E2E failure and I don't think it's caused by this PR. The evidence points to a pre-existing race in the ingest path, not the Scylla change. Details:
What actually failed
Both failing tests (investigation.spec.ts, item-signals.spec.ts) fail at the item submission, not at the investigation assertion:
Error: submitContentItem expected 202, got 500:
From the uploaded Playwright trace, POST /api/v1/items/async returns a 500 with an empty body. That empty-body 500 comes from server/routes/items/submitItems.ts#L345 (res.status(500).end()), which only fires when the Redis/BullMQ enqueue (itemSubmissionQueueBulkWrite) throws. The enqueue client is intentionally built with enableOfflineQueue: false (iocContainer L646–651), so a submit that races the client's initial connection rejects immediately. The two failing tests are the first two alphabetically i.e. the first to submit, which is exactly where that startup race surfaces (trace timestamps confirm they ran first).

Why I don't think this PR is the cause:
•  The diff is 4 files: the Scylla DI factory + a no-op class/test + one .env.example line. Nothing in the Redis/enqueue/submit/worker-startup path.
•  The flag defaults to enabled, so in E2E the factory builds the same real Scylla as main;  the no-op is never instantiated. The only added code on that path is a one-line boolean check.


Please correct me if there is any gap in my understanding. Pushed an empty commit for retry.

PR #930, opened 2026-07-21
Remove hard-coded NCMEC test-org allowlist

taobojlen:

This removes some hard-coded test orgs that were a leftover from Coop's pre-ROOST days. These presumably no longer exist (and it's messy to hard-code stuff like this).

## Tests

None.

## (Optional) Rollout Plan

I can't think of anything here, unless a managed service provider happens to be using these org IDs but that seems deeply unlikely. These hard-coded IDs were there from the very first commit in this repo.

## Checklist


juanmrad (approved):
lgtm

PR #933, opened 2026-07-23
[Easy] client: make Vite dev server allowedHosts opt-in via env var

reitblatt:
## Summary
- Vite 5+'s dev server rejects requests whose `Host` header isn't on an allowlist (DNS-rebinding protection). This breaks the dev server behind any reverse proxy that forwards a hostname other than `localhost` (Coder workspace subdomains, ngrok, tailscale funnel, etc).
- Gates `server.allowedHosts` behind a new `VITE_DEV_ALLOWED_HOSTS` env var, defaulting to unset/undefined so plain `npm start` / local dev behavior is unchanged.

Note: 'VITE_DEV_ALLOWED_HOSTS' needs to be set to 'true', not the actual hostname.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npm start` (no env var) still rejects a foreign `Host` header, unchanged from current behavior
- [x] `VITE_DEV_ALLOWED_HOSTS=true npm start` accepts a foreign `Host` header

## Summary by CodeRabbit

* **Enhancements**
  * Added configurable development-server host access through environment settings.
  * Specify permitted hostnames as a comma-separated list, including dot-prefixed subdomain wildcards.
  * Empty or unset configuration preserves the default server behavior.
  * Host access can no longer be opened to all hosts through a broad “true” setting.
  * Documented the setting in the environment configuration example.


client/vite.config.ts
```diff
@@ -23,6 +23,14 @@ export default defineConfig({
...
   },
   server: {
+    // Vite rejects requests whose Host header isn't on its allowlist (a
+    // DNS-rebinding protection, on by default since Vite 5) -- this breaks
+    // the dev server behind any reverse proxy that forwards a different
+    // hostname than localhost (Coder workspace subdomains, ngrok, tailscale
+    // funnel, etc). Opt-in via env var so plain `npm start`/local dev is
+    // unaffected and this stays a no-op unless a proxying setup opts in.
+    allowedHosts:
+      process.env.VITE_DEV_ALLOWED_HOSTS === 'true' ? true : undefined,
```
ThisIsMissEm: This should really be a fixed list, not any host, as any host can enable various attack vectors: https://vite.dev/config/server-options#server-allowedhosts

From the documentation:

> [!CAUTION]
> Setting server.allowedHosts to true allows any website to send requests to your dev server through DNS rebinding attacks, allowing them to download your source code and content. We recommend always using an explicit list of allowed hosts. See [GHSA-vg6x-rcgg-rjx6](https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6) for more details.

ThisIsMissEm:
The other issue with this is that vite dev server really shouldn't be used in production at all, you should be using a built version, which your regular webserver like nginx or apache can serve as static files

reitblatt:
> The other issue with this is that vite dev server really shouldn't be used in production at all, you should be using a built version, which your regular webserver like nginx or apache can serve as static files

Yes, this is explicitly about dev environments, not production. I use Coder to run multiple workspaces on a dedicated dev server, which in turn requires adding allowedHosts. Because we have no way of configuring it without changing the code today, that means I need to carry around a patch and always be careful not to commit it by accident.

client/vite.config.ts
reitblatt: I can tighten it up, though I don't know if this is really needed for a dev environment.

ThisIsMissEm:
> > The other issue with this is that vite dev server really shouldn't be used in production at all, you should be using a built version, which your regular webserver like nginx or apache can serve as static files
> 
> Yes, this is explicitly about dev environments, not production. I use Coder to run multiple workspaces on a dedicated dev server, which in turn requires adding allowedHosts. Because we have no way of configuring it without changing the code today, that means I need to carry around a patch and always be careful not to commit it by accident.

Then explicitly listing the hostname allowed is the way to move forwards, rather than allowing any host name, per the review by the agents above and the documentation from Vite themselves.

client/vite.config.ts
```diff
@@ -51,6 +51,22 @@ export default defineConfig(({ mode }) => {
...
+      // Set to `true` to allow all hosts, or a comma-separated list of
+      // hostnames to allow only those.
+      allowedHosts:
+        process.env.VITE_DEV_ALLOWED_HOSTS === 'true'
+          ? true
+          : process.env.VITE_DEV_ALLOWED_HOSTS
+            ? process.env.VITE_DEV_ALLOWED_HOSTS.split(',').map((host) =>
+                host.trim(),
+              )
+            : undefined,
```
ThisIsMissEm: So like add this outside of the config object:
```ts
const allowedHosts = (process.env.VITE_DEV_ALLOWED_HOSTS ?? '').split(',').map((host) => host.trim()).filter((v) => !!v)
```

Then use it:
```suggestion
      allowedHosts: allowedHosts.length > 0 ? allowedHosts : undefined,
```

The reason we don't need `true` is because you can specify a host starting with a leading dot which means all subdomains under that hostname, e.g., `.princess.works` would accept all subdomains under that domain.

ThisIsMissEm (changes requested):
We shouldn't support the unsafe use-case ever, and prefer safety.

reitblatt:
@ThisIsMissEm Look good now?

ThisIsMissEm:
@reitblatt so I'm contemplating as to if this is even needed given other work (which is to make the frontend hosted by the backend container)

reitblatt:
> @reitblatt so I'm contemplating as to if this is even needed given other work (which is to make the frontend hosted by the backend container)

I'm not sure I follow: why would that make a difference? The issue is that my dev set-up has a reverse proxy in front of Coop, so any webserver needs to be configured with an allowed-hosts, regardless of which container it's running on.

At any rate, it turns out that Vite already exposes this configuration through an env var, so this change was redundant with their built-in functionality: [__VITE_ADDITIONAL_SERVER_ALLOWED_HOSTS](https://vite.dev/config/server-options)

PR #951, opened 2026-07-28
Eliminate manual review history item refetches

taobojlen:
Related to #870.

I discovered that in the MRT, when loading a user's other items, our frontend kicks off one network request for each related item. This is in spite of the fact that all the data is *already* loaded by another request!

This PR fixes that. We now pass in the data we already have to the component, rather than re-fetching it unnecesarily.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved the display of related content in manual review submissions by using preloaded item data.
  * Reduced unnecessary loading while rendering related items.
  * Preserved existing media visibility controls and item titles.

PR #952, opened 2026-07-28
Move related item query to its consumer

taobojlen:
Small refactoring, just moving a gql query to the component that actually uses it.

Stacked on top of #951, merge that first!


julietshen (approved):
also LGTM

Issue #950, opened 2026-07-28
Enable React compiler


taobojlen:
I think it'd be great to enable [React Compiler](https://react.dev/learn/react-compiler) in Coop.

In brief, it automatically optimizes a React app by memoizing wherever it can. It's not a panacea, but it handles a lot of low-hanging fruit.

Given that Coop sometimes renders large amounts of data / many components in one page, I think this'd raise the floor for its performance.

PR #955, opened 2026-07-28
Pilot React Compiler on Label

taobojlen:
First part of #950 .

This enables the [React Compiler](https://react.dev/learn/react-compiler). For now, it's opt-in and only enabled for the simple Label component to show that it works. From here on out, we can gradually enable the compiler via `"use memo"` in more components, and eventually have the compiler enabled globally.

This'll help speed up our frontend. It's not a silver bullet but it catches low-hanging fruit.

I can run this locally and the React devtools do indeed show that the Label component was auto-memoized by the compiler!

## Summary by CodeRabbit

* **Performance**
  * Enabled React Compiler support to optimize eligible React components automatically.
  * Added improved development diagnostics for compiler issues.
* **Build & Configuration**
  * Added production-specific handling for compiler errors while preserving existing build and testing settings.
* **Bug Fixes**
  * Updated label rendering to work correctly with the new compiler optimization.


reitblatt (approved):
Very cool!

PR #970, opened 2026-08-04
Disable max lines lint on tests files

juanmrad:

As we are more tests to cover regressions and new features our tests files may become bigger than 500 lines which is acceptable. We can certainly split test files, but some are logical part of specific features. as such this is not really as needed for tests compared to normal feature files.


julietshen (approved):
LGTM

PR #979, opened 2026-08-08
[818] Fix strike analytics link to navigate to the struck user

### LINKED ISSUE #818: Clicking affected strike object does not bring correct investigation UI
### What happened

Clicking on an item actioned by the strike system brings you to: /dashboard/manual_review/investigation
instead of the actual data page.

### Version/commit

_No response_

### Affected browser(s)

_No response_

### Relevant logs/errors

```shell

```

### Anything else

_No response_

### Screenshots

_No response_

### Checklist

juanmrad:

Fixes #818

The "User" column in the strike analytics table was linking to the content item that triggered the strike chain, not the user who was actually struck. Now uses `item_creator_id/item_creator_type_id` from the action execution log so the link opens the correct user in Investigation.

Screenshot 2026-08-08 at 11 05 47 AM
Screenshot 2026-08-08 at 11 05 52 AM
Screenshot 2026-08-08 at 11 05 58 AM

## Summary by CodeRabbit

* **New Features**
  * Strike analytics now displays creator identifiers for recent strike actions when available.
  * Investigation links use creator identifiers, with item identifiers used as a fallback.
  * Recent strike action data now supports creator ID and creator type ID fields.

* **Bug Fixes**
  * Improved accuracy when identifying the source of strike actions.

### DIFF
--- a/client/src/webpages/dashboard/userStrikes/StrikeAnalyticsTab.tsx
+++ b/client/src/webpages/dashboard/userStrikes/StrikeAnalyticsTab.tsx
@@ -219,17 +219,17 @@ function RecentUserStrikeActionsTable() {
           values.creatorId && values.creatorTypeId
             ? { id: values.creatorId, typeId: values.creatorTypeId }
             : null;
-        const userId = creatorIdentity?.id ?? values.itemId;
-        const userTypeId = creatorIdentity?.typeId ?? values.itemTypeId;
         return {
-          user: (
+          user: creatorIdentity ? (
             <Link
               className="cursor-pointer shrink-0"
-              to={`/dashboard/manual_review/investigation?id=${userId}&typeId=${userTypeId}`}
+              to={`/dashboard/manual_review/investigation?id=${creatorIdentity.id}&typeId=${creatorIdentity.typeId}`}
               target="_blank"
             >
-              {userId}
+              {creatorIdentity.id}
             </Link>
+          ) : (
+            <span className="text-gray-500">{values.itemId}</span>
           ),
           action: actionsById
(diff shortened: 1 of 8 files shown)


client/src/webpages/dashboard/userStrikes/StrikeAnalyticsTab.tsx
```diff
@@ -213,14 +215,20 @@ function RecentUserStrikeActionsTable() {
       ?.slice()
       ?.sort((a, b) => (a.time > b.time ? -1 : a.time < b.time ? 1 : 0))
       .map((values) => {
+        const creatorIdentity =
+          values.creatorId && values.creatorTypeId
+            ? { id: values.creatorId, typeId: values.creatorTypeId }
+            : null;
+        const userId = creatorIdentity?.id ?? values.itemId;
+        const userTypeId = creatorIdentity?.typeId ?? values.itemTypeId;
```
reitblatt: Do we really want to fall back to the wrong data here? It seems like it would be less confusing to instead explicitly fail rather than linking to the item.

reitblatt (changes requested):
Other than the Q about fall back, LGTM!

client/src/webpages/dashboard/userStrikes/StrikeAnalyticsTab.tsx
juanmrad: Sure. I was trying to still keep a way to have a link, but given this link fall back to the item was a bit misleading. Updated so we only link when both creatorId and creatorTypeId are present. 
otherwise we show the item ID as plain text with no link.

reitblatt (approved):

PR #988, opened 2026-08-11
Fix investigation result table filters

taobojlen:

The table filters on the investigation page were broken. This PR fixes them.

https://github.com/user-attachments/assets/bdbf2db0-e364-4a03-960f-0bc4120864e0

## Tests

Manual testing: see video.

I thought of adding regression tests for this but honestly it felt like too much. If we had to test every single table filter, wow, we'd have a lot of tests.

## (Optional) Rollout Plan

N/A

## Checklist

### DIFF
--- a/client/src/webpages/dashboard/investigation/ItemInvestigationRuleResults.tsx
+++ b/client/src/webpages/dashboard/investigation/ItemInvestigationRuleResults.tsx
@@ -15,11 +15,7 @@ import {
   DefaultColumnFilter,
   SelectColumnFilter,
 } from '../components/table/filters';
-import {
-  conditionOutcomeSort,
-  ruleStatusSort,
-  stringSort,
-} from '../components/table/sort';
+import { stringSort } from '../components/table/sort';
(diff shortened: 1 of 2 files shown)


client/src/webpages/dashboard/investigation/ItemInvestigationRuleResults.tsx
```diff
@@ -215,11 +217,9 @@ export default function ItemInvestigationRuleResults(props: {
           ),
           actions: (
             <div className="w-48 mr-10 grid gap-y-1">
-              {rules
-                ?.find((it) => ruleResult.ruleId === it.id)
-                ?.actions?.map((action, i) => (
-                  <InvestigationTag title={action.name} key={i} />
-                ))}
+              {actionNames.map((actionName, i) => (
```
taobojlen, line 220: Small refactoring to make this a bit easier to read.

juanmrad (approved):
Agree. not everything needs regression testing this is good.

PR #992, opened 2026-08-11
Fix child policy score display in User Strikes

taobojlen:

This fixes a bug I found while clicking around the UI.

At `/dashboard/rules/user_strikes`, say that you have a policy with sub-policies.

- Uncheck "Apply to sub-policies"
- Click "edit policy scores"
- Change only a child policy's score
- Save, then expand "show all" again
- The child incorrectly displays the parent policy's score.

This PR fixes that.

## Tests

Tested manually, and see the test I added.

https://github.com/user-attachments/assets/89a1538c-86c0-4b68-b752-6161df2124a8

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Child policy score fields now correctly display either inherited parent scores or their own saved scores, based on the configured policy behavior.

* **Tests**
  * Added coverage verifying that saved scores are displayed correctly for parent and child policies when all policies are expanded.


juanmrad (approved):
LGTM

PR #993, opened 2026-08-11
Fix enum column sorting

taobojlen:

Found this bug while upgrading to React 19. We were modifying the case of enum values before calling `indexOf`, which meant that `indexOf` always returned -1 -- and this meant that sorting by enum values had no effect. This is now fixed.

## Tests

See the test.

## (Optional) Rollout Plan

N/A

## Checklist


juanmrad (approved):
LGTM

PR #994, opened 2026-08-11
Fix report rule insights page

taobojlen:

Coop has two rule systems:

- Proactive rules run whenever an item is submitted to Coop.
- Report rules run when a user reports an item.

We have a Report Rule details page, and we already had components for a chart and table to give data about how the report rule works. But we weren't actually rendering these components on the Report Rule details page! So, this PR adds them. They're pretty useful.

However, these existing components incorrectly used the analytics path for *proactive rules*, not report rules! So, I fixed that too, here. To get the full flow working, there were a few other bugs that came up, too:

- Date handling in the Clickhouse adapter (we were passing JS Date objects directly)
- We weren't handling timezones quite right
- Report samples use fields like `itemId` and `itemData`, but the old code expected fields from proactive rules.

If I had to guess, these report-rules-insights components were a WIP from Cove? Anyway now they're fixed and working!

https://github.com/user-attachments/assets/c5e088f0-15d2-4266-a8bc-b881bc632f66

## Tests

I tested this manually, and there's one integration test.

## (Optional) Rollout Plan

N/A

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed reporting rule insights so they appear correctly on rule detail pages.
  - Fixed inspection of reporting rule sample details, including selection of the correct execution by date and item type.
  - Improved reporting rule analytics and sample retrieval for more accurate results.
  - Preserved execution timestamps in sample details for more reliable troubleshooting.
  - Fixed sample detail views so they reset correctly when switching between lookback versions.

### DIFF
--- a/client/src/webpages/dashboard/rules/info/insights/ReportingRuleInsightsActionsChart.tsx
+++ b/client/src/webpages/dashboard/rules/info/insights/ReportingRuleInsightsActionsChart.tsx
@@ -42,7 +42,7 @@ import RuleInsightsEmptyCard from './RuleInsightsEmptyCard';
 
 gql`
   query ReportingRulePassRateAnalytics($id: ID!) {
-    rule(id: $id) {
+    reportingRule(id: $id) {
       insights {
         passRateData {
           date
@@ -69,12 +69,14 @@ export default function ReportingRuleInsightsActionsChart(props: {
 
   const [chartType, setChartType] = useState(ChartType.LINE);
 
-  const passRateData = data?.rule?.insights.passRateData ?? undefined;
+  const passRateData = data?.reportingRule?.insights.passRateData ?? undefined;
 
   const getDataInTimeWindow = useCallback(
     (
       arr?: NonNullable<
-        NonNullable<GQLReportingRulePassRateAnalyticsQuery['rule']>['insights']
+        NonNullable<
+          GQLReportingRulePassRateAnalyticsQuery['reportingRule']
+        >['insights']
       >['passRateData'],
     ) =>
       filterNullOrUndefined(arr ?? []).filter((elemWithDate) => {
--- a/server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
+++ b/server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
@@ -90,7 +90,7 @@ export class ClickhouseReportingAnalyticsAdapter implements IReportingAnalyticsA
     return rows.map((row) => ({
       totalMatches: Number(row.totalMatches),
       totalRequests: Number(row.totalRequests),
-      date: new Date(row.date).toJSON(),
+      date: row.date,
     }));
   }
 
--- a/server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
+++ b/server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
@@ -1,5 +1,6 @@
 import type { IDataWarehouse } from '../../../storage/dataWarehouse/IDataWarehouse.js';
 import type SafeTracer from '../../../utils/SafeTracer.js';
+import { getUtcDateOnlyString } from '../../../utils/time.js';
 import {
   type IReportingAnalyticsAdapter,
   type ReportingRulePassingContentSample,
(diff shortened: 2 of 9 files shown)


client/src/webpages/dashboard/rules/info/insights/sample_details/RuleInsightsSampleDetailResults.tsx
```diff
@@ -86,6 +87,7 @@ export function outcomeIcon(outcome?: GQLConditionOutcome) {
 
 export default function RuleInsightsSampleDetailResults(props: {
```
taobojlen, line 88: The changes in this file make `RuleInsightsSampleDetailResults` work for both proactive rules and report rules.

client/src/index.tsx
```diff
@@ -81,6 +81,13 @@ const client = new ApolloClient({
           return `${object.contentId}.${object.ruleId}.${object.ts}`;
         },
       },
+      ReportingRule: {
+        fields: {
+          insights: {
+            merge: true,
```
taobojlen, line 87: This tells Apollo Client that with these two queries:

```jsonc
// Chart query
insights: {
  passRateData: [...]
}

// Samples query
insights: {
  latestVersionSamples: [...]
}
```

it can merge them in its cache:

```jsonc
insights: {
  passRateData: [...],
  latestVersionSamples: [...],
}
```

server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
```diff
@@ -111,28 +120,40 @@ export class ClickhouseReportingAnalyticsAdapter implements IReportingAnalyticsA
       params.push(...itemIds);
     }
 
+    if (itemTypeIds && itemTypeIds.length > 0) {
```
taobojlen, line 123: without this change, when opening the sample details panel in the UI, we'd look up a sample based on the item ID only. but two items of different types can have the same ID! so we need to filter by item type, too.

server/plugins/warehouse/queries/ClickhouseReportingAnalyticsAdapter.ts
```diff
@@ -111,28 +120,40 @@ export class ClickhouseReportingAnalyticsAdapter implements IReportingAnalyticsA
...
+      conditions.push(
+        `item_type_id IN (${itemTypeIds.map(() => '?').join(', ')})`,
+      );
+      params.push(...itemTypeIds);
+    }
+
+    if (executionTimestamp) {
```
taobojlen, line 130: similar to above, the same item can be evaluated by the same rule many times. so by filtering on the timestamp, we can make sure we're looking up the exact evaluation that we want in the sample details panel.

server/test/integ/reporting-rule-insights.integ.test.ts
```diff
@@ -0,0 +1,192 @@
...
+  afterAll(async () => {
+    try {
+      await userCleanup?.();
+      await orgCleanup?.();
+    } finally {
+      await harness?.shutdown();
+    }
+  }, 30_000);
+
+  test('returns the selected report execution rather than another execution for the same item', async () => {
```
taobojlen: this is a pretty wordy test file. but i do like having integration tests. let me know if you think it's unnecessary!

taobojlen:
@cubic-dev-ai review this PR

server/test/integ/reporting-rule-insights.integ.test.ts
```diff
@@ -0,0 +1,192 @@
...
+     VALUES (
+       'Integration test rule', ${q(opts.ruleId)}, ${dateTime(opts.ruleVersion)},
+       'LIVE', ${q(opts.orgId)}, 'integration-test',
+       '{"conjunction":"AND","conditions":[],"result":{"outcome":"PASSED"}}',
+       1, ${dateTime(opts.executedAt)}, toDate(${dateTime(opts.executedAt)}), [],
+       ${q(jsonStringify(opts.itemData))}, ${q(opts.itemId)}, 'User',
+       ${q(opts.itemTypeId)}, 'USER', '{}', '{}', '1', 'original'
+     )`,
+    opts.harness.deps.Tracer,
+    [],
```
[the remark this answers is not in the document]
taobojlen: Not worth doing for a test.

client/src/webpages/dashboard/rules/info/insights/ReportingRuleInsightsSamplesTable.tsx
```diff
@@ -255,6 +270,8 @@ export default function ReportingRuleInsightsSamplesTable(props: {
         creatorId: sample.creatorId,
         creatorTypeId: sample.creatorTypeId,
         itemData: sample.itemData,
+        executionTimestamp:
+          sample.ts instanceof Date ? sample.ts.toISOString() : sample.ts,
```
juanmrad: An empty string is currently treated as an omitted date because it is falsy. That removes the timestamp filter and returns the newest execution, which could show a different sample than requested.

Could we distinguish an omitted date from an invalid one?
```ts
const executionTimestamp =
  input.date == null ? undefined : new Date(input.date);
```

This allows the existing invalid-date guard to reject `""` rather than silently falling back to the newest execution.

client/src/webpages/dashboard/rules/info/insights/ReportingRuleInsightsActionsChart.tsx
juanmrad: this is pre-existing, but we should fix it. The analytics rows represent whole days at midnight. A preset’s first day equals the start boundary and is excluded.

juanmrad (changes requested):
I was looking at the graph and the underlying query we use to populate it, and I think the front-end label is not representing the data p[ropertly.

The query represents rule matches, not executed actions. `totalMatches` comes from `sum(num_passes)`. A Background rule can contribute matches while executing zero actions, and a Live match can execute multiple actions. One report can also evaluate multiple thread items.

I think we may want to at the very least for this pr change the way we display it to be accurate such as "Rule Matches" rather than "Actions". We may need to add different queries and views to the dashboard to distinguish actual actions.

PR #998, opened 2026-08-11
Fix investigation rule result selection

taobojlen:

This is a small bug my LLM found. In the investigation tool, when looking up an item, we show a list of rules that it matched on. You can click each row to see some more details in a modal.

However, in the modal that appears, we were always showing the results from the *latest* version of that item. However, the item might've been updated and earlier versions might have different results on the same rule! This video shows what happened before:

https://github.com/user-attachments/assets/5399f3ff-058b-48d9-8a95-ab336a903d0b

This PR fixes this bug, so the modal shows the correct data:

https://github.com/user-attachments/assets/a0d187b4-15e4-486a-b9bb-0c9e06c59673

## Tests

Tested manually (see videos). Also added tests (see the diff).

## (Optional) Rollout Plan

N/A

## Checklist

### DIFF
--- a/client/src/webpages/dashboard/investigation/ItemInvestigationRuleResults.tsx
+++ b/client/src/webpages/dashboard/investigation/ItemInvestigationRuleResults.tsx
@@ -24,41 +24,48 @@ import Table from '../components/table/Table';
 
 import {
   GQLConditionOutcome,
+  GQLInvestigationItemsQuery,
+  GQLItemType,
   useGQLInvestigationItemsQuery,
 } from '../../../graphql/generated';
 import { ReadonlyDeep } from '../../../utils/typescript-types';
-import { LookbackVersion } from '../rules/info/insights/RuleInsightsSamplesTable';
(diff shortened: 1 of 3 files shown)


client/src/webpages/dashboard/investigation/ItemInvestigationRuleResults.tsx
```diff
@@ -274,28 +280,27 @@ export default function ItemInvestigationRuleResults(props: {
...
       {modalInfo.visible && (
         <div className="p-4">
-          <RuleInsightsSampleDetailResults
-            ruleId={modalInfo.ruleId}
-            itemIdentifier={itemIdentifier}
-            itemSubmissionDate={submissionTime}
-            lookback={LookbackVersion.LATEST}
-          />
+          {modalInfo.result ? (
+            <RuleInsightsSampleDetailResultsImpl
```
taobojlen, line 284: I don't love using an `Impl` component directly, but we already do this in `RuleTestModal` and I am reluctant to start a wider refactoring here.

PR #1002, opened 2026-08-11
Fix React warning in recent decision summary

taobojlen:

This fixes a React warning about a missing `key`.

## Tests

Manually tested.

## (Optional) Rollout Plan

N/A

## Checklist


juanmrad (approved):
👍🏻

PR #1003, opened 2026-08-12
Fix merged reports sorting

taobojlen:

The merged reports table was using the wrong sort functions, so sorting the table was broken. Now fixed.

## Tests

See regression test.

Also tested manually.

CleanShot 2026-08-18 at 12 21 26@2x

## (Optional) Rollout Plan

N/A

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved sorting for merged reports by Reported By, Reported For, Reason, and Report Time.
  * Sorting now follows the values displayed in the table, with support for ascending and descending order.

* **Tests**
  * Added coverage to verify sorting behavior across all supported report columns.

### DIFF
--- a/client/src/webpages/dashboard/mrt/manual_review_job/MergedReportsComponent.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/MergedReportsComponent.tsx
@@ -223,7 +223,7 @@ export default function MergedReportsComponent(props: {
         ) : (
           '—'
         ),
-        reason: report.reason?.trim() ? report.reason : '—',
+        reason,
         reportTime: parseDatetimeToReadableStringInCurrentTimeZone(
           report.reportedAt,
         ),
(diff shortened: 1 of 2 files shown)


client/src/webpages/dashboard/mrt/manual_review_job/MergedReportsComponent.tsx
```diff
@@ -141,6 +162,8 @@ export default function MergedReportsComponent(props: {
         hasReporter && report.displayInfo?.typeName
           ? `${report.displayInfo.typeName}: `
           : '';
+      const reportedFor = policy?.name ?? '—';
+      const reason = report.reason?.trim() || '—';
```
juanmrad, line 166: you are declaring this here, but line 226 also does the same. we could just use reason there no?

juanmrad (approved):
minor comment, but otherwise LGTM

client/src/webpages/dashboard/mrt/manual_review_job/MergedReportsComponent.tsx
taobojlen: good idea, done!

PR #1010, opened 2026-08-12
Fix manual review pie legend toggling

taobojlen:

Fixes a small bug on e.g. `/dashboard/manual_review/analytics`. If you showed a pie chart, then toggling a series from the legend would not actually hide it from the pie chart like you'd expect. This PR fixes that.

## Tests

Tested manually and added a unit test.

## (Optional) Rollout Plan

N/A

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pie chart behavior when categories are hidden.
  - Hidden categories now display zero values while preserving legend order.
  - Zero-value slices no longer show labels or label lines.
  - Restoring all categories returns the chart to its original values.

- **Tests**
  - Added coverage for hidden-category handling and value restoration.

### DIFF
--- a/client/src/webpages/dashboard/mrt/visualization/ManualReviewDashboardInsightsChart.test.ts
+++ b/client/src/webpages/dashboard/mrt/visualization/ManualReviewDashboardInsightsChart.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from 'vitest';
+
+import { getPieChartData } from './ManualReviewDashboardInsightsChart';
+
+describe('getPieChartData', () => {
+  it('zeros a hidden category and restores its value without changing legend order', () => {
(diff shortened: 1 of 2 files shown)


julietshen:
I tested this locally and I get a stray 0 label when toggling off a segment, is that expected?

Screenshot 2026-08-18 at 10 38 38 AM

reitblatt (dismissed):
@taobojlen Can you please fix the extra '0' label that @julietshen flagged?

taobojlen:
@julietshen @reitblatt thanks for the feedback (sorry Juliet I missed your comment!)

now fixed; here's a demo: https://cleanshot.com/share/Lkhr87qC

julietshen (approved):
LGTM!

juanmrad (approved):

Issue #386, opened 2026-05-01
Show decisions made from Bulk Action and Investigations in Recent Decisions table


jess-upscrolled:
Currently, when a moderator takes an action on an item outside of a decision, it does not show up anywhere other than the in Investigation UI.  The distinction between an “Action taken” and “decision made” wasn’t clear to our moderation team and required special documentation and instruction.  Furthermore, it makes it difficult for managers supervise moderators since they have no central place they can see actions moderators took outside of a decision.

It would be nice to be able to see recent actions moderators took in addition to recent decisions on a review job.  Perhaps this could be shown in the “Recent decisions” page.

This is also bad UX when moderators want to use the bulk action UI, but they cannot see the results of their actions anywhere.

julietshen:
Thanks @jess-upscrolled ! When you say that a moderator takes an action on an item outside of a decision, where are you referring to?

It makes sense to show in Recent Decisions, but I think we need more information on where these types of decisions are being made.

jess-upscrolled:
@julietshen I'm referring to actions taken from the "bulk action" or "investigation" UI.

julietshen:
@calebmcquaid do you have bandwidth to work on this?

calebmcquaid:
@julietshen Yes, I can take a look at this 👍

jess-upscrolled:
@julietshen @calebmcquaid any updates on this issue?

selena-lustig:
Hi @jess-upscrolled I'm a new PM for Roost helping out on Coop. @calebmcquaid , were you able to take a look?

calebmcquaid:
@jess-upscrolled @selena-lustig I did some investigating on this and I think I have a path forward: actions currently live in Clickhouse and decisions live in PG and sometimes a decision fires an action. What I'm proposing is to wire up a "getActionsByModerator" fetch to Clickhouse and fire that alongside the `getRecentDecisions` call we currently make. I will then combine these 2 into a "All/Decisions/Actions" feed on the current Decisions feed. This will include a filter to remove any non-moderator/automated actions and combine bulk actions into a single line

julietshen:
This makes sense to me! I like that automated actions will show up but can be filtered out

jess-upscrolled:
@calebmcquaid @julietshen that sounds great to us.  Let me know if you need any support for getting this out

taobojlen:
see also my comment on https://github.com/roostorg/coop/issues/387#issuecomment-5397114085

[...]

PR #1026, opened 2026-08-18
[386] refactor: cursor-paged decisions read for the activity feed

calebmcquaid:

First of four for #386. Split out because it's the only one touching code that already serves users.

`getRecentDecisions` is offset-paged. This issue needs a feed merged across two databases, which offset can't do. So: this extracted the shared query into a private builder, pulled the row creation into a shared function, and added `getDecisionsForActivityFeed`.

`id` is a uuid column and the cursor casts to match. This prevents decisions who share a timestamp from getting dropped or repeated at the beginning or end of pages.

## Tests

None yet, but there are some in the stacked PRs

## Summary by CodeRabbit

* **New Features**
  * Added cursor-based pagination for activity feed decisions.
  * Activity feed results maintain consistent descending chronological ordering, including reliable handling of matching timestamps.
  * Existing filters, user search, and permission-based visibility continue to apply.
  * Activity feed pages support up to 200 decisions.

* **Bug Fixes**
  * Improved pagination reliability to prevent duplicate or missing decisions when navigating between pages.
  * Preserved existing offset-based pagination behavior for recent decisions.


juanmrad (commented):
----

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -247,156 +258,192 @@ export default class DecisionAnalytics {
...
-                      eb.or(
-                        policyIds.map((policyId) =>
-                          eb(
-                            sql<string>`decision_component->>'policies'`,
-                            'like',
-                            `%"${policyId}"%`,
+    return (
+      this.pgQuery
+        .selectFrom('manual_review_tool.manual_review_decisions')
+        .select([
```
juanmrad, line 265: There is no limit on this query which makes it unbounded and potentially cause performance issues. Similarly `getDecisionsForActivityFeed` takes limit straight from the caller with no cap. 

Once the GraphQL resolver wires this up, someone could pass `limit: 1000000` and pin the read replica. Clamp it, e.g. `const limit = Math.min(opts.limit, 200);`

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -247,156 +258,192 @@ export default class DecisionAnalytics {
...
+    const { page } = input;
+    const limit = 100;
+    const decisions = await this.buildRecentDecisionsQuery({
+      userPermissions,
+      orgId,
+      input,
+    })
+      .orderBy('created_at', 'desc')
+      .limit(limit)
+      .offset(page * limit)
```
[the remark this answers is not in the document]
juanmrad: Agreed.

juanmrad (commented):
agreed with coderabbit also we may want to clamp on limits to preven unbounded queries

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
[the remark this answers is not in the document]
taobojlen: I do think it's worth using the new cursor-paged function everywhere as a follow-up. Because decisions can be created frequently, then limit/offset pagination doesn't work super well -- the next page might include decisions that were already in the previous page. Worth creating a ticket for that at least!

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -526,6 +573,57 @@ export default class DecisionAnalytics {
...
+  decision_components: ManualReviewDecisionComponent[];
+  related_actions: ManualReviewDecisionRelatedAction[];
+  created_at: Date;
+  decision_reason: string | null;
+  job_id: string;
+};
+
+/**
+ * Projects a raw `manual_review_decisions` row into the shape both
+ * `getRecentDecisions` and `getDecisionsForActivityFeed` return. Moved
```
taobojlen: Please don't add comments like "Moved verbatim out of...". That's really a comment from the agent to you, the developer, not a comment for future readers of this code.

Comments should not describe what you did in your PR, but how and why the code is the way it is, only when it's not obvious from the code itself.

server/services/manualReviewToolService/modules/DecisionAnalytics.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+      const last = firstPage[firstPage.length - 1];
+      // The cursor's `id` is a real uuid pulled from a previous row, exactly
+      // as a real caller would pass it. If the `::uuid` cast on the cursor
+      // predicate were missing or wrong, this call raises
+      // `22P02 invalid input syntax for type uuid` instead of returning.
+      const secondPage = await mrtService.getDecisionsForActivityFeed({
+        userPermissions: [],
+        orgId: org.id,
+        input: { page: 0 },
```
taobojlen: it's odd that this function takes a `page` input that's ignored. can you remove that input since it's unused?

taobojlen (changes requested):

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -247,156 +258,192 @@ export default class DecisionAnalytics {
...
+      input,
+    })
+      .$if(cursor !== undefined, (qb) =>
+        qb.where(
+          sql`(created_at, id)`,
+          '<',
+          // `id` is uuid. The cast is load-bearing: without it Postgres
+          // infers the bind type from the column and a non-uuid string
+          // raises 22P02.
+          sql`(${cursor!.ts}, ${cursor!.id}::uuid)`,
```
taobojlen: I just looked into the indexes we have on this table. For some reason we have two b-tree indexes on `created_at`.

But can we add a new index on `(org_id, created_at DESC, id DESC)` to help this new cursor-based pagination? Worth doing now since this table tends to get pretty large over time.

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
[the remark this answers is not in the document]
calebmcquaid: I added the `id` tie breaker. I will add a ticket for the cursor paging

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -247,156 +258,192 @@ export default class DecisionAnalytics {
...
         ),
       )
```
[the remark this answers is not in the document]
calebmcquaid: Done

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: I clamped it at 200.

Passed a limit to the builder

server/services/manualReviewToolService/modules/DecisionAnalytics.test.ts
calebmcquaid: Removed

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: Removed

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -247,156 +258,192 @@ export default class DecisionAnalytics {
...
+   */
+  async getDecisionsForActivityFeed(opts: {
+    userPermissions: UserPermission[];
+    orgId: string;
+    input: RecentDecisionsFilterInput;
+    // The DECISIONS side of a per-store cursor. The caller must never pass
+    // the actions side's id here: `id` is uuid, and a non-uuid string raises
+    // 22P02 invalid input syntax for type uuid.
+    cursor?: { ts: Date; id: string };
+    limit: number;
```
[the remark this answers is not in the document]
calebmcquaid: Done

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -1183,6 +1183,16 @@ export class ManualReviewToolService {
     return this.decisionAnalytics.getRecentDecisions(opts);
   }
 
+  async getDecisionsForActivityFeed(opts: {
+    userPermissions: UserPermission[];
+    orgId: string;
+    input: RecentDecisionsFilterInput;
+    cursor?: { ts: Date; id: string };
+    limit: number;
+  }) {
```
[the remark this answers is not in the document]
calebmcquaid: Done

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: I added these in 2 migrations

db/src/scripts/api-server-pg/2026.09.09T13.50.00.drop_duplicate_mrt_decisions_created_at_index.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+DROP INDEX IF EXISTS manual_review_tool.mrt_decisions_created_at_idx;
```
taobojlen, line 3: can you use `DROP INDEX CONCURRENTLY` here to avoid locking the entire table?

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -274,12 +278,21 @@ export default class DecisionAnalytics {
       .execute();
   }
 
-  async getRecentDecisions(opts: {
+  /**
+   * Shared query body for `getRecentDecisions` and
+   * `getDecisionsForActivityFeed`: the select list, every filter predicate,
+   * the `userSearchString` branch, and the NCMEC permission gate. Stops
```
taobojlen, line 284: Probably best not to mention the specific callers in the code comment. It can get out of sync quickly. (I think I mentioned this somewhere but can't find it now!)

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -288,162 +301,197 @@ export default class DecisionAnalytics {
...
-    }));
+    return decisions.map(mapDecisionRow);
+  }
+
+  /**
+   * Cursor-paged twin of `getRecentDecisions`, for the merged activity feed.
+   *
+   * Kept separate rather than folded into `getRecentDecisions`: that query is
+   * offset-paged and consumed elsewhere, and changing its contract would
+   * break callers this feature has no reason to touch.
```
taobojlen, line 456: This second paragraph of the code comment is unnecessary. It explains why its contract wasn't changed -- but a future reader doesn't need to know that.

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -288,162 +301,197 @@ export default class DecisionAnalytics {
...
+   * `created_at`-only bound would drop or repeat decisions sharing an
+   * instant.
+   */
+  async getDecisionsForActivityFeed(opts: {
+    userPermissions: UserPermission[];
+    orgId: string;
+    input: Omit<RecentDecisionsFilterInput, 'page'>;
+    // The DECISIONS side of a per-store cursor. The caller must never pass
+    // the actions side's id here: `id` is uuid, and a non-uuid string raises
+    // 22P02 invalid input syntax for type uuid.
```
taobojlen, line 469: I don't think a reader will know what 22P02 is. It would be better to say "raises a Postgres error".

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
```diff
@@ -288,162 +301,197 @@ export default class DecisionAnalytics {
...
+      orgId,
+      input,
+      limit,
+    });
+    const pagedQuery = cursor
+      ? baseQuery.where(
+          sql`(created_at, id)`,
+          '<',
+          // `id` is uuid. The cast is load-bearing: without it Postgres
+          // infers the bind type from the column and a non-uuid string
```
taobojlen, line 485: how would a non-UUID string be passed in here anyway?

server/services/manualReviewToolService/modules/DecisionAnalytics.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        break;
+      }
+      const last = page[page.length - 1];
+      cursor = { ts: last.createdAt, id: last.id };
+    }
+    return ids;
+  };
+
+  testWithDecisions()(
+    'pages a real cursor across the id-cast boundary without erroring',
```
taobojlen, line 112: IMO this test would be better framed as just "paginates using a cursor". that's the core behavior we want! the "id-cast boundary" is an implementation detail.

taobojlen (approved):
Some comments; the only change I'd like to see before merging is using `DROP INDEX CONCURRENTLY`, others are non-blocking!

db/src/scripts/api-server-pg/2026.09.09T13.50.00.drop_duplicate_mrt_decisions_created_at_index.sql
calebmcquaid: Done!

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: Removed

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: Removed

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: Adjusted

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
calebmcquaid: Good callout

server/services/manualReviewToolService/modules/DecisionAnalytics.test.ts
calebmcquaid: Renamed and dropped the comment

server/services/manualReviewToolService/modules/DecisionAnalytics.ts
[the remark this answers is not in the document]
cassidyjames: Filed as https://github.com/roostorg/coop/issues/1212

cassidyjames (approved):

PR #1029, opened 2026-08-18
[386] feat: merge review decisions and manual actions into one

calebmcquaid:

Third of four for #386. Stacked on #B.

`ModerationActivityFeed` merges PG review-job decisions with manual Clickhouse actions into one list.

Two things to call out:

The cursor keeps a separate position per store. Decision ids are uuids,  action ids are strings like `manual-action-run:abc`. One shared cursor throws a PG type error on any page ending on an action, and casting to text doesn't help because JS, CH and PG each sort strings differently.

## Tests

24 unit tests, full server suite passes, typechecks standalone.

### DIFF
--- a/server/graphql/modules/moderationActivity.test.ts
+++ b/server/graphql/modules/moderationActivity.test.ts
@@ -1,0 +1,35 @@
+import { typeDefs } from './moderationActivity.js';
+
+describe('moderationActivity schema', () => {
+  it('exposes both queries', () => {
+    expect(typeDefs).toContain('recentModerationActivity(');
+    expect(typeDefs).toContain('manualActionItems(');
+  });
+
+  it('models rows as an interface so the client discriminates on __typename', () => {
+    expect(typeDefs).toContain('interface ModerationActivityRow');
+    expect(typeDefs).toContain(
+      'type ReviewJobDecisionRow implements ModerationActivityRow',
+    );
+    expect(typeDefs).toContain(
+      'type ManualActionRow implements ModerationActivityRow',
+    );
+  });
+
+  it('caps page sizes', () => {
+    expect(typeDefs).toContain('limit: Int');
+  });
+
+  it('uses the opaque Cursor scalar rather than a bare string', () => {
+    expect(typeDefs).toContain('cursor: Cursor');
+    expect(typeDefs).toContain('nextCursor: Cursor');
+  });
+
+  it('does not expose offset on manualActionItems', () => {
+    const inputBlock = typeDefs.slice(
+      typeDefs.indexOf('input ManualActionItemsInput'),
+      typeDefs.indexOf('type ManualActionItem'),
+    );
+    expect(inputBlock).not.toContain('offset');
+  });
+});
--- a/server/graphql/modules/moderationActivity.ts
+++ b/server/graphql/modules/moderationActivity.ts
@@ -1,0 +1,285 @@
+import { type ItemInvestigationService } from '../../services/itemInvestigationService/index.js';
+import { type ManualReviewToolService } from '../../services/manualReviewToolService/index.js';
+import { type ActivityFeedFilterInput } from '../../services/moderationActivityFeed/index.js';
+import { UserPermission } from '../../services/userManagementService/index.js';
+import { filterNullOrUndefined } from '../../utils/collections.js';
+import {
+  type GQLManualActionRowResolvers,
+  type GQLModerationActivityRowResolvers,
+  type GQLQueryResolvers,
+  type GQLRecentManualReviewDecisionType,
+  type GQLReviewJobDecisionRowResolvers,
+} from '../generated.js';
+import { forbiddenError, unauthenticatedError } from '../utils/errors.js';
+
+/** Matches the decisions feed's page size so the two merge evenly. */
+const DEFAULT_PAGE_SIZE = 100;
+const MAX_PAGE_SIZE = 200;
+const DEFAULT_ITEM_PAGE_SIZE = 100;
+/**
+ * Covers any run Bulk Actioning can produce — its 1000-item limit is enforced
+ * client-side only (`BulkActioningDashboard`), and `bulkExecuteActions` caps
+ * nothing server-side, so a programmatic caller can still exceed this. The
+ * panel keeps its truncation notice for that case rather than pretending the
+ * list is always complete.
+ */
+const MAX_ITEM_PAGE_SIZE = 1000;
+
+const typeDefs = /* GraphQL */ `
+  input RecentModerationActivityInput {
+    "Opaque cursor from a previous page. Absent for the newest page."
+    cursor: Cursor
+    "Rows per page. Defaults to 100, capped at 200."
+    limit: Int
+    view: ActivityView
+    startTime: DateTime
+    endTime: DateTime
+    reviewerIds: [ID!]
+    policyIds: [ID!]
+    queueIds: [ID!]
+    decisions: [RecentManualReviewDecisionType!]
+    userSearchString: String
+  }
+
+  enum ActivityView {
+    ALL
+    DECISIONS
+    ACTIONS
+  }
+
+  "Fields every row in the merged activity feed carries, regardless of kind."
+  interface ModerationActivityRow {
+    id: ID!
+    ts: DateTime!
+    reviewerId: ID
+  }
+
+  type ReviewJobDecisionRow implements ModerationActivityRow {
+    id: ID!
+    ts: DateTime!
+    reviewerId: ID
+    jobId: String
+    queueId: ID
+    itemId: ID
+    itemTypeId: ID
+    decisions: [ManualReviewDecisionComponent!]!
+    decisionReason: String
+  }
+
+  """
+  One operation a moderator ran outside a review job, from Bulk Actioning or
+  Investigation. These produce no manual review decision, so they are otherwise
+  invisible outside the actioned item's own history. A bulk run writes one
+  execution per item per action; this collapses those into a single row.
+  """
+  type ManualActionRow implements ModerationActivityRow {
+    id: ID!
+    ts: DateTime!
+    reviewerId: ID
+    correlationId: ID!
+    itemTypeId: ID
+    actionIds: [ID!]!
+    policyIds: [ID!]!
+    actorNote: String
+    itemCount: Int!
+    failedCount: Int!
+  }
+
+  type ModerationActivityPage {
+    rows: [ModerationActivityRow!]!
+    "Null means the end of the feed."
+    nextCursor: Cursor
+  }
+
+  input ManualActionItemsInput {
+    correlationId: ID!
+    "max(ts) of the operation, from its feed row. Bounds the partition scan."
+    occurredAt: DateTime!
+    "Defaults to 100, capped at 500."
+    limit: Int
+  }
+
+  type ManualActionItem {
+    itemId: ID!
+    itemTypeId: ID
+    failed: Boolean!
+  }
+
+  type ManualActionItemsPage {
+    items: [ManualActionItem!]!
+    totalCount: Int!
+  }
+
+  type Query {
+    recentModerationActivity(
+      input: RecentModerationActivityInput!
+    ): ModerationActivityPage!
+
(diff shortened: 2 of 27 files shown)


taobojlen:
> Third of four for https://github.com/roostorg/coop/issues/386. Stacked on #B.

there's no such thing as #B -- it's very helpful if you write PR descriptions in your own words so we avoid things like this!

when i see LLM-written stuff like this, especially stuff that has small mistakes, it has the effect that i don't trust *any* of the code or comments in the PR because i don't know if you've read over it yourself and can stand 100% behind it. i really don't mean to be rude! but i think it's important because it makes PR reviews much harder -- i don't know if i'm reviewing caleb's contribution or an LLM's contribution, you know? i trust the former a lot more :)

server/graphql/modules/moderationActivity.test.ts
```diff
@@ -0,0 +1,35 @@
+import { typeDefs } from './moderationActivity.js';
+
+describe('moderationActivity schema', () => {
```
taobojlen, line 3: these tests are kind of weird -- they're basically testing that some strings contain some substrings?

i don't think we need these!

server/graphql/modules/moderationActivity.ts
```diff
@@ -0,0 +1,285 @@
...
+import {
+  type GQLManualActionRowResolvers,
+  type GQLModerationActivityRowResolvers,
+  type GQLQueryResolvers,
+  type GQLRecentManualReviewDecisionType,
+  type GQLReviewJobDecisionRowResolvers,
+} from '../generated.js';
+import { forbiddenError, unauthenticatedError } from '../utils/errors.js';
+
+/** Matches the decisions feed's page size so the two merge evenly. */
```
taobojlen, line 15: if these numbers need to match those from the decision feed, can we please use the same constants so there's no risk of drift?

server/graphql/modules/moderationActivity.ts
```diff
@@ -0,0 +1,285 @@
...
+const MAX_PAGE_SIZE = 200;
+const DEFAULT_ITEM_PAGE_SIZE = 100;
+/**
+ * Covers any run Bulk Actioning can produce — its 1000-item limit is enforced
+ * client-side only (`BulkActioningDashboard`), and `bulkExecuteActions` caps
+ * nothing server-side, so a programmatic caller can still exceed this. The
+ * panel keeps its truncation notice for that case rather than pretending the
+ * list is always complete.
+ */
+const MAX_ITEM_PAGE_SIZE = 1000;
```
taobojlen, line 26: i don't understand this. if there's a limit that's only enforced client-side, then what is this constant for? when the comment says "the panel", what is it referring to?

server/graphql/modules/moderationActivity.ts
```diff
@@ -0,0 +1,285 @@
...
+
+type DecisionFilterEntry = NonNullable<
+  ActivityFeedFilterInput['decisions']
+>[number];
+
+/**
+ * Maps the GraphQL oneof-style decision filter to the service's tagged-union
+ * shape. Mirrors the equivalent mapping in `manualReviewTool.ts`'s
+ * `getRecentDecisions` resolver — kept separate rather than shared, since
+ * touching that resolver is out of scope here.
```
taobojlen, line 140: please review your code and its comments before handing it off. this comment does not make sense to commit to the codebase, since it's clearly directed at the developer making changes, rather than future readers of the code.

taobojlen: anyway, for this function, we also have `getRecentDecisions` and `getSkipsForRecentDecisions` in `manualReviewTool.ts` that do basically the same thing. i do think it's worth extracting this into a shared helper (can be done in a PR upstack).

server/graphql/modules/moderationActivity.ts
```diff
@@ -0,0 +1,285 @@
...
+        queueIds: input.queueIds ?? undefined,
+        decisions: toActivityFeedDecisionFilter(input.decisions),
+        userSearchString: input.userSearchString ?? undefined,
+      },
+      view: input.view ?? 'ALL',
+      limit: Math.min(
+        Math.max(1, input.limit ?? DEFAULT_PAGE_SIZE),
+        MAX_PAGE_SIZE,
+      ),
+      // Already decoded by the `Cursor` scalar; `parseActivityCursor` handles
```
taobojlen, line 209: as a reader of this comment my first question is "what is `parseActivityCursor`?".

IMO no need to reference it here.

server/graphql/modules/moderationActivity.ts
```diff
@@ -0,0 +1,285 @@
...
+  },
+
+  async manualActionItems(_, { input }, context) {
+    const user = context.getUser();
+    if (user == null) {
+      throw unauthenticatedError('Authenticated user required');
+    }
+    // Manual actions are only ever taken from Investigation or Bulk Actioning,
+    // so viewing them requires the permission that gates Investigation itself.
+    // EXTERNAL_MODERATOR — read-only access for external moderation partners —
```
taobojlen, line 222: this comment does not need to reference `EXTERNAL_MODERATOR` at all, and only adds a risk of drift.

server/graphql/modules/moderationActivity.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    return context.services.ModerationActivityFeed.getManualActionItems({
+      orgId: user.orgId,
+      correlationId: input.correlationId,
+      occurredAt: new Date(input.occurredAt),
+      limit: Math.min(
+        Math.max(1, input.limit ?? DEFAULT_ITEM_PAGE_SIZE),
+        MAX_ITEM_PAGE_SIZE,
+      ),
+      // `offset` stays internal, not part of the public schema:
```
taobojlen, line 239: can you explain this comment?

if offset is always 0, how does one paginate?

server/services/moderationActivityFeed/activityCursor.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      cursor,
+    );
+  });
+
+  it('treats an absent cursor as the newest page', () => {
+    expect(parseActivityCursor(undefined)).toBeUndefined();
+    expect(parseActivityCursor(null)).toBeUndefined();
+  });
+
+  it('rejects a malformed cursor rather than silently restarting the feed', () => {
```
taobojlen, line 43: how might a malformed cursor occur?

i think that it's preferable to error loudly in that case, rather than silently swallow the unexpected data.

server/services/moderationActivityFeed/activityCursor.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Position in the merged activity feed — one per store, not one shared.
+ *
+ * Decision ids are `uuid`; action ids are `manual-action-run:<uuid>` strings.
+ * A shared id would be bound against a uuid column and rejected, and casting to
+ * text gives three different orderings (JS UTF-16, ClickHouse UTF-8, Postgres
+ * collation). Each side pages from its own last-returned row instead.
+ *
+ * `null` means "start from the newest" for that store.
+ */
+export type ActivityCursor = {
```
taobojlen, line 21: hmm, i think we could do this a bit simpler, using a single timestamp as the cursor for both stores. something like

```ts
type ActivityCursor = {
  timestamp: Date;
  decision_id: string | null;
  action_id: string | null;
}
```

i.e. the timestamp is the main cursor shared across both postgres + clickhouse, and the ID fields are just in case of multiple rows with the same timestamp.

would that work?

server/services/moderationActivityFeed/mergeActivityRows.ts
```diff
@@ -0,0 +1,94 @@
...
+ * merging two stores that share no index.
+ */
+export function mergeActivityRows(
+  decisions: readonly ActivityRow[],
+  actions: readonly ActivityRow[],
+  limit: number,
+  incoming: ActivityCursor | undefined,
+): { rows: ActivityRow[]; nextCursor: JsonValue | null } {
+  const merged = [...decisions, ...actions].sort(byTimeThenKindThenIdDesc);
+  const hasMore = merged.length > limit;
```
taobojlen, line 35: won't this always be true? if `decisions` and `actions` are both length `limit + 1`, as per the docstring, then `hasMore` will always be true here?

server/services/moderationActivityFeed/ModerationActivityFeed.test.ts
```diff
@@ -0,0 +1,188 @@
+import { UserPermission } from '../userManagementService/index.js';
+import { ModerationActivityFeed } from './ModerationActivityFeed.js';
+
+/**
+ * Manual actions require VIEW_INVESTIGATION. Every role except
+ * EXTERNAL_MODERATOR holds it, so this is the ordinary case.
```
taobojlen, line 6: no need to reference EXTERNAL_MODERATOR here i think

server/services/moderationActivityFeed/ModerationActivityFeed.ts
```diff
@@ -0,0 +1,163 @@
...
+ * supply; the placeholder below is inert because `getDecisionsForActivityFeed`
+ * never reads `page`.
+ */
+export type ActivityFeedFilterInput = Omit<RecentDecisionsFilterInput, 'page'>;
+
+/** Placeholder for the `page` field `getDecisionsForActivityFeed` never reads. */
+const UNUSED_PAGE = 0;
+
+/** Merged-view lookback. See the spec's "Time window" section. */
+const MERGED_VIEW_WINDOW_MS = 30 * 24 * 60 * 60 * 1000;
```
taobojlen, line 25: this comment makes no sense as we do not commit specs to this repo. (anyway code comments should not require reading a separate spec to understand).

taobojlen: but anyway, do we need this time window?

server/services/moderationActivityFeed/ModerationActivityFeed.ts
```diff
@@ -0,0 +1,163 @@
...
+  return (
+    (input.queueIds?.length ?? 0) > 0 || (input.decisions?.length ?? 0) > 0
+  );
+}
+
+/**
+ * Manual actions are only ever taken from Investigation or Bulk Actioning, so
+ * seeing them requires the permission that gates Investigation itself.
+ *
+ * This is not a formality. `EXTERNAL_MODERATOR` — described in
```
taobojlen, line 42: again, let's stop referencing this role, as it may change.

server/services/moderationActivityFeed/ModerationActivityFeed.ts
```diff
@@ -0,0 +1,163 @@
...
+  userPermissions: readonly UserPermission[],
+): boolean {
+  return userPermissions.includes(UserPermission.VIEW_INVESTIGATION);
+}
+
+/**
+ * The Recent Decisions feed, merged from two stores.
+ *
+ * Review-job decisions live in Postgres and manual moderator actions live in
+ * ClickHouse. Nothing outside this module knows that.
```
taobojlen, line 60:
```suggestion
 * ClickHouse.
 ```

unnecessary LLM fluff

File: server/services/moderationActivityFeed/ModerationActivityFeed.ts [hunk 15]
```diff
@@ -0,0 +1,163 @@
...
+  ) {}
+
+  async getPage(opts: {
+    userPermissions: UserPermission[];
+    orgId: string;
+    input: ActivityFeedFilterInput;
+    view: ActivityView;
+    limit: number;
+    /** Already decoded by the `Cursor` scalar; absent for the newest page. */
+    cursor?: unknown;
```
taobojlen, line 75: why `unknown` type here?

server/services/moderationActivityFeed/ModerationActivityFeed.ts
```diff
@@ -0,0 +1,163 @@
...
+      })),
+      limit,
+      decoded,
+    );
+  }
+
+  /**
+   * Every item id one manual action run touched.
+   *
+   * Callers MUST check `VIEW_INVESTIGATION` first — see `canViewManualActions`.
```
taobojlen, line 148: do you find this comment easy to read?

taobojlen (changes requested):

PR #1032, opened 2026-08-18
Handle zeros in table filtering

taobojlen:

This fixes a bug found by CodeRabbit. Our table numeric filtering didn't work for zeros. Now it does.

## Tests

See the test.

## (Optional) Rollout Plan

N/A

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Numeric range filters now correctly apply lower and upper bounds set to zero.
  - Filtering results no longer incorrectly include negative or positive values when zero is specified as a boundary.

- **Tests**
  - Added coverage verifying that zero-valued minimum and maximum range filters return only matching zero values.


selena-lustig:
@julietshen This seems pretty low risk to merge

juanmrad (approved):
👍🏻

PR #1037, opened 2026-08-18
fix: race condition between "orgUsers" and a "useEffect" in ManualReviewQueueForm

jess-upscrolled:

There was a race condition between when the "useEffect" that sets the initial values of "ManualReviewQueueForm" runs and when useGQLQueueFormDataQuery resolves orgUsers.

If the useEffect runs before orgUsers is set, the react state variable "moderatorsWithAccess" gets set to an empty array (due to missing orgUsers).  Since the useEffect was missing the dependency, it would never run again.  Thus "moderatorsWithAccess" would end up empty when submitted to the API server.  This causes any unrelated updates to wipe this field.

The following PR adds memorization orgUsers and some related data and fixes the dependencies of the "useEffect" hook

## Tests

Tested the "Update Manual Review Queue" locally

## Checklist

  Unsure if this requires a CHANGELOG.md or docs change as its a bug fix.  Might be worth mentioning in case other adopters are facing a similar confusing bug

  N/A

  N/A

  N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manual review queue initialization when reviewer access changes.
  * Ensured reviewer information is consistently sorted and updated across user and organization-wide access settings.


client/src/webpages/dashboard/mrt/ManualReviewQueueForm.tsx
```diff
@@ -307,8 +319,7 @@ export default function ManualReviewQueueForm() {
     setClearReportsTriggerActionIds([
       ...(queue.clearReportsTriggerActionIds ?? []),
     ]);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [queue, queue?.autoCloseJobs]);
+  }, [queue, queue?.autoCloseJobs, sortedUsers]);
```
julietshen: non-blocking nit: I think this would reset the entire form if there's any org-user data change, not just moderators. This is a low probability since this only refetches the two queue queries after its own mutation but it could potentially take out any unsaved edits 

maybe split the moderator initialization which needs `sortedUsers` into its own effect from the queue-field initialization?

julietshen (approved):
LGTM!

client/src/webpages/dashboard/mrt/ManualReviewQueueForm.tsx
jess-upscrolled: @julietshen good catch.  I believe it would reset the form it that were to happen.  I've split the moderator initialization into its own effect

julietshen:
awesome thank you!

PR #1045, opened 2026-08-20
Remove Google Cloud Translation API

### LINKED ISSUE #1044: Remove Google Cloud Translation API
## Summary

While cleaning up old SaaS signals that created a confusing user experience in the UI, I saw that the Google Cloud Translation API is also "coming soon". This is a paid API and because it was "coming soon" in the SaaS origins, remains hidden in the UI behind a flag. 

The intended purpose is to translate content and be part of the rules systems, but today is a paid API wired through a hidden global key, which can be a trap for self-hosted adopters. 

## Proposal

Remove the code for Google Cloud Translation API 

### Alternatives Proposed
I thought about leaving it, as translation is often very helpful - however, there is a Google Translate extension and Chrome users have it built in. There are other ways to translate content independent of Coop, and given the open source / free principles of what should be included out-of-the-box I decided against it.

julietshen:
Fixes #1044

This is a paid API that was hidden in the UI behind a flag as "coming soon"; it's a legacy of the SaaS era. 

## Summary

This cleans up the Google Cloud Translation API. It removes:
- The signal class and its IoC registration, plus the internal GOOGLE_CLOUD_TRANSLATE_MODEL signal type and its entries in the type maps
- The `ENGLISH_TRANSLATION` derived field: GraphQL enum value, the recipe in the derived-fields service, the display-name case, and the client rule-form mapping. VIDEO_TRANSCRIPTION (Whisper-backed) is untouched.
- also removed `GOOGLE_TRANSLATE_API_KEY` from .env.example and the env type declarations
- the @google-cloud/translate dependency

https://claude.ai/code/session_01Wx3nmYtSe8Cb37QQ8WY3mx

## Summary by CodeRabbit

- **Removed Features**
  - Removed English translation as a derived-field option.
  - Removed Google Cloud Translation integration and related configuration.
  - Video transcription remains available as the supported derived-field output type.


server/.env.example
```diff
@@ -132,7 +132,6 @@ SENDGRID_API_KEY=SG.REAL_API_KEY_HERE
 GOOGLE_PLACES_API_KEY=
 READ_ME_JWT_SECRET=
 LAUNCHDARKLY_SECRET=
```
ThisIsMissEm, line 134: @julietshen seeing this, it may be worth going back through and seeing what feature flags remain in a separate PR, and unflagging or removing those things.

ThisIsMissEm (approved):
LGTM

PR #643, opened 2026-05-29
Add #342 background-jobs integration test (and unbreak rule-anomaly query)

julietshen:
Closes #342 (per the [planning comment](https://github.com/roostorg/coop/issues/342#issuecomment-4576284547) on that issue).

## Summary

Adds an integ test for two of the three jobs called out by #342: `RefreshMRTDecisionsMaterializedViewJob` and `DetectRulePassRateAnomaliesJob`. `RetryFailedNcmecDecisionsJob` is deferred to a follow-up — its invariants live partly in NCMEC's HTTP response and need an outbound-HTTP simulator we don't have today.

Each scenario runs the job twice against the real stack and asserts the no-double-fire, no-skip invariants directly.

## What surfaced along the way

Building the rule-anomaly test required fixing `getRuleAnomalyDetectionStatistics` first — the query was carried over verbatim from the pre-#133 Snowflake era and fails against ClickHouse in **five distinct ways**, all of which compound to silently make every rule's computed alarm state `INSUFFICIENT_DATA` in production:

1. `array_size(passes_distinct_user_ids)` — Snowflake `ARRAY_SIZE` has no ClickHouse equivalent. Column is a JSON `String`, not a native `Array`. Use `JSONLength(...)`.
2. `SYSDATE()` — not a ClickHouse function. Replace with `now64(3)` (UTC `DateTime64(3)`).
3. `ts_start_inclusive >= ?` with a `Date` bind — the warehouse adapter formats Dates as ISO 8601 with `Z\, which ClickHouse rejects when implicitly converting to `DateTime64(3)`. Wrap with `parseDateTime64BestEffort(?)`.
4. `row.RULE_ID`, `row.NUM_PASSES`, etc — Snowflake defaults to uppercase identifiers; ClickHouse returns column names as written. Every column access was reading `undefined`, so`getCurrentPeriodRuleAlarmStatuses` collapsed every rule into a single `{ undefined: INSUFFICIENT_DATA }` bucket.
5. `passCount` /`passingUsersCount` / `runsCount` cast as `number` — ClickHouse `Int64` / `UInt64` columns deserialise to JS `BigInt`. `binomialTest` throws on `1n` even though it accepts `1`. Coerce with `Number(...)`.

Net: the rule-anomaly detection job is effectively a no-op against the current ClickHouse setup until this lands. The integ test demonstrates an end-to-end alarm transition with the fix in place.

## AGENTS.md callouts

Per `coop/AGENTS.md` > Human-approval-required actions:
[...]

PR #1056, opened 2026-08-23
Fix rule history startDate filter dropping other rules' versions

juanmrad:

Fix bug uncovered by #643 integration tests that was not fixed befor emerging. `getSimplifiedRuleHistory` applied the "version in effect at startDate" cut to a flat list of every rule. The neighbor check treated the next *row* as the next *version of the same rule*, so an older seed rule later in the list could drop a newer rule's history.

That meant that `DetectRulePassRateAnomaliesJob` then saw no min version, discarded a full week of ClickHouse stats, and stored `INSUFFICIENT_DATA` instead of `ALARM`. 
This was uncovered by the `background-jobs.integ.test.ts` failure on main.

We now filter per rule id (same change in the reporting copy of the helper). If history is missing, keep warehouse stats. Added a regression test: newer rule followed by an older rule in the raw list.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected rule history filtering so each rule retains the version active at the selected start date and all subsequent versions.
  - Preserved newer rules when older rules appear later in history.
  - Ensured combined rule history results are returned in chronological order.
  - Marked alarm statuses as insufficient data when minimum anomaly-detection version information is unavailable.
- **Tests**
  - Added regression coverage for multi-rule history filtering, chronological ordering, and insufficient-data alarm statuses.

### DIFF
--- a/server/services/reportingService/reportingRuleHistoryHelpers.test.ts
+++ b/server/services/reportingService/reportingRuleHistoryHelpers.test.ts
@@ -1,0 +1,56 @@
+import { getSimplifiedRuleHistory } from './reportingRuleHistoryHelpers.js';
+
+describe('reporting getSimplifiedRuleHistory', () => {
+  test('filters startDate per rule and returns retained versions chronologically', async () => {
+    const mockGetRawHistory = async () => [
+      {
+        id: 'old-seed',
+        name: 'Old seed',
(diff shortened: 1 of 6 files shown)


calebmcquaid (approved):
I believe Copilot is missing the ordering happening above. CodeRabbit suggestion may be legit but I'll leave up to you.

Issue #704, opened 2026-06-04
Document deployment considerations for popular hosting providers


cassidyjames:
We should cover specific settings/considerations for popular hosting providers, like AWS, Digital Ocean, or Azure.

Issue #683, opened 2026-06-03
Recent-decisions chart hardcodes "Child Safety" as the policy label for NCMEC rows


julietshen:
## Summary

The Recent Decisions chart renders the literal string `Child Safety` as the policy label for every `SubmitNCMECReportDecisionComponent` row, instead of looking up the org's actual child-safety policy.

## Where

- `client/src/webpages/dashboard/mrt/ManualReviewRecentDecisions.tsx:399-401`

## Why this matters

- Wrong for any org whose child-safety policy isn't named `Child Safety`.
- Inconsistent with how every other policy is displayed (via the org's policy name, not a hardcoded constant).
- Distorts the recent-decisions chart for orgs trying to attribute decisions to specific policies.

## Proposed fix

- Look up the org's child-safety policy (by id or by a configured flag on the policy) and render its name.
- If no child-safety policy is configured, render a neutral fallback (e.g. the decision-type label) rather than a misleading constant.

---

🤖 Filed with [Claude Code](https://claude.com/claude-code)

JagadeeshChandra12:
Hi @julietshen,

I'd like to work on this issue. I'm currently investigating the affected component and will submit a PR once I have a fix ready.

Thanks!

JagadeeshChandra12:
Hi @julietshen,

I've been investigating this issue and found the hardcoded `"Child Safety"` label in `ManualReviewRecentDecisions.tsx`.

From tracing the data flow:

* `SubmitNCMECReportDecisionComponent` currently exposes only `type` and `reportedMedia`.
* The frontend policy lookup relies on policy IDs.
* NCMEC jobs appear to retain `policyIds` in the backend job payload, but those IDs are not exposed through the GraphQL `SubmitNCMECReportDecisionComponent` type.

Before I proceed with a PR, should the intended fix be to expose the associated `policyIds` on `SubmitNCMECReportDecisionComponent` and use those for policy-name lookup, or is there another source that should be used to identify the child-safety policy?

Thanks!

PR #1058, opened 2026-08-24
Hover unblurs a blurred video the same way as an image

### LINKED ISSUE #524: Hovering over a blurred video does not deblur
Blurred videos don't deblur when hovering over them, which only lets two workarounds be allowed:
- turn off blurring (not a great option)
- maximise the video

forest-savage1234:
# Hover unblurs a blurred video the same way as an image

Closes roostorg/coop#524.

Images already use Tailwind `hover:blur-none`. Videos only drop blur when `playing` is true, so hover does nothing unless the reviewer hits play or turns blurring off. ThatKoffe named hover.

This puts `group` on the outer player wrapper and `group-hover:blur-none` on the blurred inner div so hovering the play button or the video unblurs, matching the image control. Play-to-unblur is unchanged. Wellness docs still say play-to-unblur only. Juliet pointed at #194; that is a different still-open migration, not this PR.

## Tests

Dest check, Ubuntu Node v24.18.0, vitest 4.1.10:

```
✓ src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobContentBlurableVideo.test.tsx (1 test) 272ms
Test Files  1 passed (1)
     Tests  1 passed (1)
```

The test renders a blurred video and asserts `group` on the wrapper plus `group-hover:blur-none` and `blur-sm` on the inner player. No live job with a video was hovered.

## (Optional) Rollout Plan

Not this PR. Local UI hover class only.

## Checklist

User-facing. CHANGELOG has the Review Console line. History / db scripts / signals are not this PR. Wellness docs were not rewritten.

## Files

1. `client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobContentBlurableVideo.tsx`
2. `client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobContentBlurableVideo.test.tsx`
3. `CHANGELOG.md`

## Out of scope

- Wellness docs rewrite (`docs/user/review-console.md` still says play-to-unblur)
- Coop #768 (Unblur All Media / iframes)
- Coop #194 (content-review-filters migration)
- Coop #683 / #628 / #551 / #202 / #704 / #703
- Changing `BLUR_LEVELS` or wellness defaults
- CASE / CAC / CaseLinker

## How to review

On a job with a blurred video, hover the player. Blur should lift and return when the cursor leaves, without pressing play. Images should still behave as they do.

I am a volunteer. Thank you for the time. I am trying to become more useful on this work, so I welcome a critical look. If this is the wrong cut, or you want me to stand down, say so and I will recut from notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Review Console
 
+- Fixed hovering a blurred video not lifting blur the way images already do; the player now unblurs on hover without requiring play (#524)
 - Fixed a "Job submission failed" error that prevented reviewers from clearing jobs whose item had an unparseable `Created At` value; the decision now records instead of failing (#913)
 - Fixed a "Something Went Wrong" page when opening a job or queue whose `Created At` value was unparseable; the affected date now shows `Unknown` instead of blanking the view (#916)
 - Fixed "Oldest Task Age" on the MRT queues dashboard showing the newest job's age instead of the oldest (#909)
--- a/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobContentBlurableVideo.test.tsx
+++ b/client/src/webpages/dashboard/mrt/manual_review_job/ManualReviewJobContentBlurableVideo.test.tsx
@@ -1,0 +1,31 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import '@testing-library/jest-dom/extend-expect';
+
+import ManualReviewJobContentBlurableVideo from './ManualReviewJobContentBlurableVideo';
+
+vi.mock('react-player', () => ({
+  default: function MockPlayer() {
+    return <div data-testid="player" />;
+  },
+}));
+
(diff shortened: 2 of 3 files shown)


forest-savage1234:
Thank you for the time on this review. This only unblurs a video on hover the same way as an image. If that is the wrong cut, say so and I will recut from notes.

taobojlen:
thanks for this contribution @forest-savage1234 !

i think we should probably focus on #194, which would conflict with this change here.

forest-savage1234:
Thank you—agreed. Closing this in favor of #194 so the shared `content-review-filters` migration defines the behavior rather than adding another local path. I’ll keep the branch for reference.

forest-savage1234:
Thank you—agreed. Closing this in favor of #194 so the shared `content-review-filters` migration defines the behavior rather than adding another local path. I’ll keep the branch for reference.Thank you—agreed. Closing this in favor of #194 so the shared `content-review-filters` migration defines the behavior rather than adding another local path. I’ll keep the branch for reference.

PR #528, opened 2026-05-20
[Permissions] Capability-based authz refactor. gate org secrets, deleteUser, and ReadMe JWT on MANAGE_ORG

juanmrad:
## Context & Requests for Reviewers

Foundation work for #406 (granular permissions). 

Replace the remaining role-string authz callsites with `user.getPermissions().includes(...)` so server-side checks no longer branch on `user.role`. After this PR there are zero resolver-level role-string checks.

Add a new `MANAGE_ROLES` capability granted only to `ADMIN` today — the seam future PRs will gate the role-editing API on so admins can delegate role editing without granting full `MANAGE_ORG`.

Close several server-side authz gaps where any authenticated user in an org could read or modify org-level secrets. All now require `MANAGE_ORG`.

The refactor is intentionally a no-op for every role today (verified by `permissioning.test.ts`); only the security fixes change observable behavior, and only for non-admins.

## Summary by CodeRabbit

* **Security Improvements**
  * Stricter permission gating across org secrets, API keys, integrations, user deletion, retroaction/backtests, routing, org settings; new MANAGE_ROLES permission added.

* **UI Changes**
  * Save/manage controls and "Skip to next job" now enforce permission checks with clearer disabled states/tooltips and minor text/layout tweaks.

* **Bug Fixes**
  * Prevented unauthorized access to sensitive fields and improved safe handling of varying data shapes.

* **Tests**
  * Expanded test coverage for permission-gated paths and permission/role behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/roostorg/coop/pull/528?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1066, opened 2026-08-25
fix: seed system roles

### LINKED ISSUE #1061: Seed system roles on org creation and backfill orgs created since May 2026
The seed migration file inserts the 7 system roles for every org that existed the moment the migration ran, but for any orgs created afterward the roles do not get created. This should probably be part of the `create-org-and-user.ts` bootstrap file and a backfill migration for any orgs created since May 2026.
(I wrote this I am a human!)

Context: because of this, newer orgs have zero `public.roles` rows and their users have `role_id = NULL`. Today this is masked by fallbacks (static role defaults, static permission map, and lazy materialization on first role edit), so permissions still work. But it means role rows and `users.role_id` are unreliable, which blocks referencing roles by `role_id` (e.g. queue access by role, Groups, deletable custom roles).
(Claude wrote this)

taobojlen:

Fixes #1061 

In https://github.com/roostorg/coop/pull/528, we started working on more granular permissions. Roles + permissions are now persisted in the database.

However, new orgs created since then did *not* get the new DB-backed roles and permissions. Instead, they rely on defaults in the code, and the permissions only get created on-demand when the defaults are edited. This means that there are really 2 different implementations of permissions!

We can clean this up by getting every org to use the DB-backed roles and permissions so there's just one way to do it. This PR:

- Adds a data migration that creates the default roles + permissions for all orgs
- Updates the new-org script so that new orgs are created with the default roles + permissions

Once this is merged, we no longer need the non-DB, code `isFallback` path for permissions.

## Tests

I tested this manually by creating an org on main, editing a default role, then checking out this branch and running the data migration, and verifying the permissions in the DB.

## (Optional) Rollout Plan

N/A

## Checklist

## Summary by CodeRabbit

* **New Features**
  * System roles and their default permissions are now automatically created for each organization.
  * Existing users and invitations are linked to their corresponding system roles.
  * New organizations receive role setup automatically during creation.
  * Role provisioning is safe to repeat and preserves existing role assignments.

### DIFF
--- a/db/src/scripts/api-server-pg/2026.08.25T15.56.24.seed_missing_system_roles.sql
+++ b/db/src/scripts/api-server-pg/2026.08.25T15.56.24.seed_missing_system_roles.sql
@@ -1,0 +1,110 @@
+BEGIN;
+
+WITH role_defaults(key, display_name, description) AS (
+    VALUES
+        ('ADMIN', 'Admin', 'Manages all users in the org and has every permission.'),
+        ('RULES_MANAGER', 'Rules Manager', 'Can modify and run rules and actions, but cannot manage users.'),
+        ('MODERATOR_MANAGER', 'Moderator Manager', 'Can view MRT, edit queues, and manage moderators.'),
+        ('MODERATOR', 'Moderator', 'Reviews queues they have been granted access to.'),
+        ('CHILD_SAFETY_MODERATOR', 'Child Safety Moderator', 'Reviews child safety jobs, including NCMEC.'),
+        ('ANALYST', 'Analyst', 'Reads rules and insights; can run backtests and edit non-live rules.'),
+        ('EXTERNAL_MODERATOR', 'External Moderator', 'Read-only MRT access for external moderation partners.')
+),
+inserted_roles AS (
+    INSERT INTO public.roles (org_id, key, display_name, description, is_system)
+    SELECT
+        o.id,
+        r.key,
+        r.display_name,
+        r.description,
+        true
+    FROM public.orgs AS o
+    CROSS JOIN role_defaults AS r
+    ON CONFLICT (org_id, key) DO NOTHING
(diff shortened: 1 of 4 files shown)


db/src/scripts/api-server-pg/2026.08.25T15.56.24.seed_missing_system_roles.sql
```diff
@@ -0,0 +1,110 @@
...
+    AND r.key = u.role
+    AND r.is_system = true;
+
+UPDATE public.invite_user_tokens AS t
+SET role_id = r.id
+FROM public.roles AS r
+WHERE t.role_id IS NULL
+    AND r.org_id = t.org_id
+    AND r.key = t.role
+    AND r.is_system = true;
```
julietshen, line 108: Not blocking!

BUT i think these two backfill UPDATEs are the only part of your change that touches existing rows, and they aren't covered by tests right now. the test file uses `seedSystemRolesForOrg`, which doesn't include this logic. Could we add a test that seeds an org, inserts a user and an invite token with a legacy `role` and `role_id IS NULL`, then exercises the backfill and confirms both get linked to the matching system role?

julietshen (approved):
LGTM! Maybe we should add a little heads up somewhere about how renaming a key
requires a data migration updating all three columns at the same time or things break?

db/src/scripts/api-server-pg/2026.08.25T15.56.24.seed_missing_system_roles.sql
juanmrad: Normally I would agree with this assessment, but in practice adding tests to migrations is mostly dead code later own. This is something better tested manually ensuring this migration works as expected and that is all.

juanmrad (approved):
Thank you for this. Totally scaped my mind 😓

PR #1067, opened 2026-08-25
Remove legacy user role implementation

taobojlen:

Follow-up from #1061. Now that all orgs have DB-backed roles and permissions, we don't need the fallback code, and we don't need the legacy `role` column in the database. Instead, we can always use `role_id` -- an actual foreign key that refers to the role in the database.

## Tests

I ran this locally, using an org first created on `main`, and tested that I can click around (+ edit permissions) without issues.

## (Optional) Rollout Plan

Because this removes a GraphQL field, any live users may get some errors in their browsers until they refresh their page. I think this is acceptable for this self-hosted software. The alternative is to deprecate the field and stop using it, and then only later (when everyone has updated to that version), *then* fully remove it. I don't think the juice is worth the squeeze there.

## Checklist

## Summary by CodeRabbit

* **Role Management**
  * Roles now use persisted organization-specific records for consistent permissions and assignments.
  * Role identifiers are always available when viewing or editing roles.
  * The “Using default permissions” notice is no longer displayed.

* **Data Reliability**
  * User and invitation role assignments remain preserved when roles are renamed or permissions change.
  * Invalid or missing role assignments are rejected instead of being silently replaced with defaults.

### DIFF
--- a/db/src/scripts/api-server-pg/2026.08.25T15.56.25.remove_legacy_role_columns.sql
+++ b/db/src/scripts/api-server-pg/2026.08.25T15.56.25.remove_legacy_role_columns.sql
@@ -1,0 +1,76 @@
+BEGIN;
+
+-- Prevent role assignments from changing between preflight validation and DDL.
+LOCK TABLE public.roles, public.users, public.invite_user_tokens IN ACCESS EXCLUSIVE MODE;
+
+DO $$
+BEGIN
+    IF EXISTS (
+        SELECT 1
+        FROM public.users AS u
+        LEFT JOIN public.roles AS r ON r.id = u.role_id
+        WHERE u.role_id IS NULL
+            OR r.id IS NULL
+            OR r.org_id IS DISTINCT FROM u.org_id
+            OR r.is_system IS DISTINCT FROM true
+            OR r.key IS NULL
+            OR r.key NOT IN (
+                'ADMIN',
+                'RULES_MANAGER',
+                'MODERATOR_MANAGER',
+                'MODERATOR',
+                'CHILD_SAFETY_MODERATOR',
+                'ANALYST',
+                'EXTERNAL_MODERATOR'
+            )
+    ) THEN
+        RAISE EXCEPTION 'public.users contains an invalid role_id: role_id must be non-null, reference public.roles(id), belong to the same org, and identify a system role with a recognized key';
+    END IF;
+
+    IF EXISTS (
+        SELECT 1
+        FROM public.invite_user_tokens AS t
+        LEFT JOIN public.roles AS r ON r.id = t.role_id
+        WHERE t.role_id IS NULL
+            OR r.id IS NULL
+            OR r.org_id IS DISTINCT FROM t.org_id
+            OR r.is_system IS DISTINCT FROM true
+            OR r.key IS NULL
+            OR r.key NOT IN (
+                'ADMIN',
+                'RULES_MANAGER',
+                'MODERATOR_MANAGER',
+                'MODERATOR',
+                'CHILD_SAFETY_MODERATOR',
+                'ANALYST',
+                'EXTERNAL_MODERATOR'
+            )
+    ) THEN
+        RAISE EXCEPTION 'public.invite_user_tokens contains an invalid role_id: role_id must be non-null, reference public.roles(id), belong to the same org, and identify a system role with a recognized key';
+    END IF;
+END
+$$;
+
+ALTER TABLE public.users
+    ALTER COLUMN role_id SET NOT NULL;
+
+ALTER TABLE public.invite_user_tokens
+    ALTER COLUMN role_id SET NOT NULL;
+
+ALTER TABLE public.users
+    DROP CONSTRAINT users_role_id_fkey,
+    ADD CONSTRAINT users_role_id_fkey
+        FOREIGN KEY (role_id) REFERENCES public.roles(id) ON DELETE RESTRICT;
+
+ALTER TABLE public.invite_user_tokens
+    DROP CONSTRAINT invite_user_tokens_role_id_fkey,
+    ADD CONSTRAINT invite_user_tokens_role_id_fkey
+        FOREIGN KEY (role_id) REFERENCES public.roles(id) ON DELETE RESTRICT;
+
+ALTER TABLE public.users
+    DROP COLUMN role;
+
+ALTER TABLE public.invite_user_tokens
+    DROP COLUMN role;
+
+COMMIT;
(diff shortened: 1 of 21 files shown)


julietshen (approved):
LGTM!

the `DROP COLUMN` part might need to be called out in the rollout note to make it super clear that the DB change is forward-only.

juanmrad (changes requested):
blocking while I gather my thoughts for a moment to explain some problem I think this may have.

db/src/scripts/api-server-pg/2026.08.25T15.56.25.remove_legacy_role_columns.sql
```diff
@@ -0,0 +1,76 @@
...
+ALTER TABLE public.invite_user_tokens
+    DROP CONSTRAINT invite_user_tokens_role_id_fkey,
+    ADD CONSTRAINT invite_user_tokens_role_id_fkey
+        FOREIGN KEY (role_id) REFERENCES public.roles(id) ON DELETE RESTRICT;
+
+ALTER TABLE public.users
+    DROP COLUMN role;
+
+ALTER TABLE public.invite_user_tokens
+    DROP COLUMN role;
```
juanmrad: These two `DROP COLUMN` statements shouldn't be in this migration. During deployment it's very normal to:

```
Download New Tag -> Run Migrations -> Deploy Server
```

Between "Run Migrations" and "Deploy Server", the old code is still running and doing `SELECT role` / `INSERT INTO ... (role, ...)` against these tables. Dropping the column out from under it will crash the server during that window.

Let's remove both `DROP COLUMN` statements from this migration and keep the columns around (unused) for one release cycle. 

Can you open an issue to track dropping them in the next release? That gives everyone time to deploy this version first.

juanmrad: Given we won't drop this the migration should instead do:
```sql
ALTER TABLE public.users ALTER COLUMN role DROP NOT NULL;
ALTER TABLE public.invite_user_tokens ALTER COLUMN role DROP NOT NULL;
```

To avoid the new code breaking trying to insert here where the data is required to not be null

server/graphql/datasources/rolePersistence.ts
```diff
@@ -127,64 +110,58 @@ export async function kyselyListRolesForOrg(
...
+        .onRef('roles.org_id', '=', 'users.org_id'),
+    )
+    .select((eb) => [
+      'roles.id as roleId',
+      eb.fn.countAll<string>().as('count'),
+    ])
+    .where('users.org_id', '=', orgId)
+    .where('roles.org_id', '=', orgId)
+    .where('users.approved_by_admin', '=', true)
+    .where('users.rejected_by_admin', '=', false)
```
juanmrad: This changes behavior. 

The old query counted all non-rejected users (including pending), this one counts only approved AND non-rejected. 

The userCount in the role editor will drop for orgs with pending users. 🤔 This may be ok, but would like @julietshen to confirm. Basically should count of user's that are part of a role be All users ( pending + active ) or just active. It may be good to be All as we also wouldn't want an admin deleting a role that is assigned to a user in pending state and cause issues. 

Your migration adds `ON DELETE RESTRICT` to the FK, so deleting a role with any assigned user (pending or active) would fail at the DB level, but the UI count showing 0 would make it look safe to delete.

juanmrad (changes requested):
One critical issue on the migration side and another smaller on the persistence count side.

db/src/scripts/api-server-pg/2026.08.25T15.56.25.remove_legacy_role_columns.sql
taobojlen: Yeah I've intentionally accepted this kind of thing in other PRs (e.g. removing GraphQL fields)! I figured that for most self-hosters, HA is not the focus, and so brief transient errors while upgrading is acceptable. Especially since we a) don't know which versions of Coop people are running, and b) there's no guarantee that people upgrade one version at a time -- they could easily go from, say, 1.0.0 to 1.2.0, skipping 1.0.x and 1.1.x releases.

In terms of the cost/benefit, that's where I've landed since it's a lot of overhead for us to track otherwise. But happy to make this change! Coming right up...

I'd be curious to hear your thoughts on: how long we should keep e.g. legacy columns around for. That'll set the expected upgrade cadence for adopters.

server/graphql/datasources/rolePersistence.ts
taobojlen: you're right! the method is called `countApprovedUsersByRole` so I *assume* the intent is to only count approved users... but changing behaviour here is unnecessary so i've undone that.

server/graphql/modules/roles.ts
```diff
@@ -15,16 +15,14 @@ import {
 
 const typeDefs = /* GraphQL */ `
   type Role {
-    "Persisted public.roles.id, or null when the row is materialized lazily on first save."
-    id: ID
+    "Persisted public.roles.id."
+    id: ID!
```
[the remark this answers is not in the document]
taobojlen: cc @juanmrad , curious to hear what you think about this (related to our discusison above)!

db/src/scripts/api-server-pg/2026.08.25T15.56.25.remove_legacy_role_columns.sql
juanmrad: one version at least. it is sensible for us to ask people to upgrade one version at a time or at least flag it. it would be up to adopters to follow that recommendation. we can definitively talk more about this tho

server/graphql/modules/roles.ts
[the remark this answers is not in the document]
juanmrad: This one is a bit more interesting. On the DB table it would break everything. The role field is read on every user load every `SELECT` from `public.users` in `userKyselyPersistence.ts` includes role in `USER_COLUMNS`. So if the column is dropped while old code is running, it's not just role management that breaks it will break login, every authenticated request, user listing, SAML resolution, basically anything that loads a user. That's a full outage, not a partial one.

While here being GraphQL it would only affect during the deploy of front-end/back-end while they are not in sync and only affect this specific query not every single one. So on here it would be more acceptable to remove it, but at the same time just to keep things clean it would be better to deprecate it and then remove on the future release.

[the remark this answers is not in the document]
taobojlen: I'm going to keep it as-is (i.e. removed here), then, just to reduce overhead tasks!

db/src/scripts/api-server-pg/2026.08.25T15.56.25.remove_legacy_role_columns.sql
taobojlen: @juanmrad makes sense. the PR is ready for re-review, let me know what you think!

taobojlen: Follow-up here: https://github.com/roostorg/coop/issues/1141

juanmrad (approved):
👍🏻 looks good now

PR #1069, opened 2026-08-26
fix: Regression on queue creation

### LINKED ISSUE #1074: Creating queue fails with "name already exists" error
### What happened

I was testing something locally with a brand new org (single admin, no queues/rules/etc. yet) and went to create my first queue. I got an error saying:

> A manual review queue with that name already exists in this organization.

…Except it doesn't! This is a fresh org without any queues.

When debugging, I noticed I'd left the default selection (just the admin user) for the "who can review this queue" field when I submitted. When I removed that, submitting worked fine.

### Repro

1. Create a fresh org + admin user
2. Log in, go to queues, create a new queue
3. Leave the default reviewer selection as-is and submit
4. Get the "queue with that name already exists" error on what would be the very first queue for this org

jess-upscrolled:
Fixes #1074 

When moving `setModeratorsWithAccess` to its own effect, I didn't add the short-circuit behavior the original effect had when `queue` is undefined (i.e., queue creation).  This causes `moderatorsWithAccess` to be populated with admins whereas previously it was initially empty.  

While this may be a bit of a UX-win since it more correctly shows the future state of access to the queue, it causes a regression due an existing bug with createManualReviewQueue when one or more of the userIds is also the currently logged in user (or really any case where userIds is duplicated).

The following PR fixes the UI regression and patches createManualReviewQueue so that it always de-duplicates userIds before attempting to insert into users_and_accessible_queues

Note, semi-related: admins/mod-managers don't appear to need to be added to users_and_accessible_queues in order to view/update/delete queues.  Ideally, we should probably strip admins/mod-managers from moderatorsWithAccess before sending to the server.  That way we never populate people who have implicit access into users_and_accessible_queues which will make explicitlyAssignedReviewers more accurate to its naming/intention.  I avoided this approach since it would constitutes a change in behavior and data model which could have unintended side-effects.

## Tests

Manually tested creating queue.  

Before e6aa07fad9dd324fd2d1c58276dafde3fe8e8181:
- Admins are **not** auto-populated in the form
- Creating a new queue succeeds

After e6aa07fad9dd324fd2d1c58276dafde3fe8e8181
- Admins are auto-populated in form
- creating a new queue fails

After patch
- Admins are **not** auto-populated in the form
- Creating a new queue succeeds.

## Checklist

N/A, fixing a regression

N/A

N/A

N/A

## Summary by CodeRabbit

* **New Features**
  * Added handle-time reporting, with filtering by date range, queue, and reviewer.
  * Handle-time results can be grouped by queue or reviewer.
  * Added assignment and job creation timestamps to manual review decisions.
* **Bug Fixes**
  * Improved manual review queue setup when queue data is unavailable.
  * Prevented duplicate users, including the current user, from being added to queues.
  * Added validation to prevent invalid date ranges in handle-time reports.

### DIFF
--- a/client/src/webpages/dashboard/mrt/ManualReviewQueueForm.tsx
+++ b/client/src/webpages/dashboard/mrt/ManualReviewQueueForm.tsx
@@ -321,8 +321,11 @@ export default function ManualReviewQueueForm() {
   }, [queue, queue?.autoCloseJobs]);
 
   useEffect(() => {
+    if (!queue) {
+      return;
+    }
     setModeratorsWithAccess(sortedUsers.map((it) => it.id));
-  }, [sortedUsers]);
+  }, [queue, sortedUsers]);
 
   if (queueQueryError || error) {
     // eslint-disable-next-line @typescript-eslint/only-throw-error
--- a/server/graphql/modules/manualReviewTool.ts
+++ b/server/graphql/modules/manualReviewTool.ts
@@ -2353,12 +2353,15 @@ const Mutation: GQLMutationResolvers = {
       clearReportsScope,
       clearReportsTriggerActionIds,
     } = params.input;
+
+    // createManualReviewQueue expects userIds to be unique so as to not violate the DB primary key for users_and_accessible_queues
+    const userIdsWithCurrentUser = Array.from(new Set([...userIds, user.id]));
     try {
       const queue =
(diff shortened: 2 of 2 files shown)


cassidyjames (approved):
Confirmed this fixes the issue. Thanks!

PR #1057, opened 2026-08-24
Show a spinner while the recent-decisions CSV is building

forest-savage1234:
# Show a spinner while the recent-decisions CSV is building

Closes #551.

## Context

The Recent Decisions Download button can make up to 100 paginated requests before building `decisions.csv`. Its Ant `loading` prop was wired to `allDecisionsLoading`, which belongs to the table refresh query, so the download itself provided no progress feedback.

This change:

- adds a download-specific `isDownloadingDecisions` state;
- keeps the Download button loading until the CSV handoff finishes;
- clears the state in `finally`;
- shows an error toast if generation fails; and
- leaves the existing pagination and CSV construction behavior unchanged.

Per review feedback, this branch is rebuilt on current `main` after the repository's line-ending normalization change. It contains only the component and its focused test; the earlier changelog edit is removed.

## Tests

- `npx prettier --check src/webpages/dashboard/mrt/ManualReviewRecentDecisions.tsx src/webpages/dashboard/mrt/ManualReviewRecentDecisions.test.tsx`
- focused Vitest run with the repository's transforms and native config loader:
  - 1 test file passed
  - 2 tests passed
- `git diff --check`

The tests cover both a delayed successful download (spinner persists through completion) and a rejected query (error toast is shown and loading clears).

## Checklist

- [ ] User-facing changelog/docs update — omitted following maintainer guidance that this small loading-state fix does not need a changelog entry.
- [ ] Content/action/rule/policy model history changes — not applicable.
- [ ] Database schema changes — not applicable.
- [ ] New signal error classification — not applicable.

## Files

1. `client/src/webpages/dashboard/mrt/ManualReviewRecentDecisions.tsx`
2. `client/src/webpages/dashboard/mrt/ManualReviewRecentDecisions.test.tsx`

## How to review

Click Download on Recent Decisions. The button should show Ant's loading state until `decisions.csv` starts. If the paginated query fails, an error toast should appear and the button should become usable again. Refresh Table remains on its existing loading state.

PR #1071, opened 2026-08-26
chore: Normalize line-endings with .gitattributes

cassidyjames:
If someone uses Windows or has different defaults than Unix/Linux, we end up with mixed line-endings which causes issues with diffs (like in #1057). Git gives us a way to specify preferred line endings with `.gitattributes`, so let's do that and have Git automatically handle it for us.

This is the same as what we did in Osprey: https://github.com/roostorg/osprey/pull/451


julietshen (approved):
good catch!

PR #1115, opened 2026-09-02
docs: Flag that wellness settings don't apply to iframes by default

cassidyjames:
As raised in #768

## Summary by CodeRabbit

- **Documentation**
  - Clarified that wellness settings apply by default to images and videos.
  - Documented that embedded content is excluded unless the deployment implements a content proxy.


julietshen (approved):
LGTM

PR #1122, opened 2026-09-04
Add AgentScan PR labeling workflow

taobojlen:
This is a small proposal: add a GitHub Action to run [AgentScan](https://agentscan.tools/) on PRs. This tool checks a contributor's GitHub history, and if it detects signs of the user being an AI agent, it adds a small label to the PR. So it's purely informational, but it'd be a nice additional data point when we get PRs from new contributors.

The reason is that PRs from AI agents are largely not worth reviewing, IMO! Reviewing PRs is a decent workload, and if there isn't a human on the other end, there's very little benefit -- it's much simpler for us to prompt an agent ourselves.

## Summary by CodeRabbit

* **Chores**
  * Added automated security scanning for pull requests targeting the main branch.
  * Scan results can now be applied as pull request labels.

### DIFF
--- a/.github/workflows/agentscan.yml
+++ b/.github/workflows/agentscan.yml
@@ -14,8 +14,6 @@ jobs:
     if: github.repository == 'roostorg/coop'
     runs-on: ubuntu-latest
     permissions:
-      contents: read
-      issues: write
       pull-requests: write
     steps:
       - name: Cache AgentScan analysis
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
It's worth trying out imho! Depending on our feeling after using it for a while, we should also propose to the Osprey and community repos and maybe write something about it in the ROOST-wide community docs. (It would be nice to be able to add GitHub Actions to an org…)

PR #1136, opened 2026-09-07
Update to Debian Trixie

### LINKED ISSUE #1135: Fix CI errors on `check_generated_graphql`
The workflow https://github.com/roostorg/coop/actions/workflows/apply_pr_checks.yaml?query=check_generated_graphql has been failing for a lot of prs that have unrelated changes. This is disruptive as well as signals an error that is unrelated to pr changes. 

We should investigate the root cause and fix it.

taobojlen:

CI is failing on main, e.g. [here](https://github.com/roostorg/coop/actions/runs/34079272163/job/101611385271).

The issue is that Debian Bullseye is now past its end-of-life so its package registries aren't being updated. This PR updates us to Trixie, the current version of Debian.

Fixes #1135.

## Tests

See CI passing here! I also startup up Coop manually to verify I can click around.

## (Optional) Rollout Plan

N/A

### DIFF
--- a/.github/workflows/apply_pr_checks.yaml
+++ b/.github/workflows/apply_pr_checks.yaml
@@ -213,6 +213,12 @@ jobs:
         name: Get Migrations Logs
         run: docker compose logs migrations
 
+      - name: Verify drop_dbs service
+        # Exercise package installation and teardown after the tests have
+        # finished using this job's disposable databases. CI's target DB is
+        # postgres, so connect to template1 to avoid dropping the current DB.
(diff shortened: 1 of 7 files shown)


juanmrad (approved):
sure this is an ok patch while I work on a proper fix #1135

taobojlen:
@juanmrad this didn't actually work fully; turns out there were even more packages with issues! looks like we need to update debian. i'm on it (unless you're alraedy ahead of me?)

juanmrad:
cosing as we moved to bookworm to keep in line with Node base image being bookworm also. we can explore trixie in the future.

Issue #1110, opened 2026-09-02
e2e CI job sometimes fails


cassidyjames:
Seems like a race condition, because re-running the failed CI job has always worked. But it seems weird. According to my friend Claude, it could be because the Redis queue isn't always actually ready when the test is running.

cassidyjames:
Dug into this a bit more. Redis is the only backing service in docker-compose.yaml without a healthcheck, so starting backing services in the e2e workflow doesn't wait for it to be ready before the server starts. The last few failures I saw of the e2e job looked the same, too.

Adding a healthcheck to the redis service seems like the fix.

PR #1137, opened 2026-09-07
Fix flaky E2E tests

taobojlen:

This fixes an occassional flake in our E2E tests in CI: we now build the server up front.

Before, we were running a live-reloading watcher, which means that changes on the filesystem (like Playwright writing an artifact) could trigger restarts and thus break network requests.

May be related to #1110 

## Tests

See CI passing!

## (Optional) Rollout Plan

N/A

## Summary by CodeRabbit

* **Chores**
  * Updated end-to-end test setup to build the server before launching it.
  * Server startup now uses the transpiled production output during end-to-end testing.


juanmrad (approved):
👍🏻

PR #1138, opened 2026-09-07
Fix `check_generated_graphql` and move to Debian 12 as 11 is EOL.

juanmrad:

Debian 11 LTS ended 31 August 2026. `check_generated_graphql` (and any other job that `apt-get`s inside `node:24.20.0-bullseye-slim`) started failing on `apt-get install git` with exit 100. Frontend CI stayed green because that Dockerfile never calls apt.

This PR retags every Coop Node image from `node:24.20.0-bullseye-slim` to `node:24.20.0-bookworm-slim` (same Node 24.20.0). `nodejs-instrumentation` already used `node:24.20.0`, which is Bookworm.

GraphQL codegen CI used to build the server Docker image just to run `npm run generate`. It now runs on the GitHub runner the same way formatting does: `install Node`, `npm ci`, `regenerate types`, and fail if those files changed.

`git` was in the production server image only so that Docker job could run git status. That install is removed.

## Summary by CodeRabbit

- **Chores**
  - Updated production and development container images to Debian 12 (Bookworm), including Node.js and NGINX-based images.
  - Streamlined automated GraphQL generation and formatting checks by running them directly in the Node.js environment.
  - Removed the dedicated Docker-based GraphQL code generation service.

- **Documentation**
  - Updated the changelog and development guides with the new container images and revised local CI commands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ For more information about each release including git tags and artifacts, see [R
 - Settings "Other" tab renamed to "Partial Items" and its settings relocated ([#965](https://github.com/roostorg/coop/pull/965) by [@golden-fox07](https://github.com/golden-fox07))
 - Queue deletion is refused while routing rules still reference the queue ([#808](https://github.com/roostorg/coop/pull/808) by [@reitblatt](https://github.com/reitblatt))
 - Long text fields in the review console collapse behind a "Read more" control ([#903](https://github.com/roostorg/coop/pull/903) by [@taobojlen](https://github.com/taobojlen))
-- Production Node images moved from Debian 11 (bullseye) to Debian 12 (bookworm)
+- Production Node images moved from Debian 11 (bullseye) to Debian 12 (bookworm) ([#1138](https://github.com/roostorg/coop/pull/1138) by [@juanmrad](https://github.com/juanmrad))
 
 ### Removed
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ For more information about each release including git tags and artifacts, see [R
 - Settings "Other" tab renamed to "Partial Items" and its settings relocated ([#965](https://github.com/roostorg/coop/pull/965) by [@golden-fox07](https://github.com/golden-fox07))
 - Queue deletion is refused while routing rules still reference the queue ([#808](https://github.com/roostorg/coop/pull/808) by [@reitblatt](https://github.com/reitblatt))
 - Long text fields in the review console collapse behind a "Read more" control ([#903](https://github.com/roostorg/coop/pull/903) by [@taobojlen](https://github.com/taobojlen))
+- Production Node images moved from Debian 11 (bullseye) to Debian 12 (bookworm) ([#1138](https://github.com/roostorg/coop/pull/1138) by [@juanmrad](https://github.com/juanmrad))
```
taobojlen, line 24: I kind of wonder if we should remove CHANGELOG.md entirely or at least add some guidelines to AGENTS.md. Right now we're using it very inconsistently!

taobojlen (approved):
Thank you!! And ty for fixing the codegen CI job; it was weird that we ran that the way we did.

CHANGELOG.md
juanmrad: Yeah I've had some thoughts on this. part of me feels this should just be auto generated based on prs merged and not part of the prs. I update it when I feel there is something of high importance to explain on the change log which is good to do as part of the pr, but then also sometimes forget to update lol.

PR #1151, opened 2026-09-09
Harden access control for queues/jobs

### LINKED ISSUE #1150: Authenticated users can enumerate all queues and read all jobs
### What happened

Reported by @serendipty01: authenticated users can enumerate queues and read the jobs regardless of whether they have permissions according to the queue's settings.

### Version/commit

7074f0222d6b593e45470231be2ac8379c2ec32f

### Affected browser(s)

_No response_

### Relevant logs/errors

```shell

```

### Anything else

_No response_

### Screenshots

_No response_

### Checklist

cassidyjames:
From @serendipty01:

> Any authenticated user, regardless of role or queue membership, could enumerate every MRT queue in an org and read or dequeue/lock every job in them -- including CSAM/NCMEC-designated queues -- because Org.mrtQueues, Query.manualReviewQueue, Query.getTotalPendingJobsCount, and Mutation.dequeueManualReviewJob all resolved queues via the *Dangerously*BypassPermissioning helpers with no permission or membership check.
> 
> Switch those four call sites to getReviewableQueuesForUser, the existing membership-aware lookup already used correctly elsewhere (e.g. RoutingRule.destinationQueue, user.ts's reviewableQueues resolver). Add the missing auth check on ManualReviewQueue.jobs (defense-in-depth; not independently reachable today).
> 
> The *Dangerously*BypassPermissioning primitives themselves are kept -- RoutingRule.destinationQueue uses the same bypass correctly, gated behind EDIT_MRT_QUEUES.

Fixes #1150

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

## Summary by CodeRabbit

* **New Features**
  * Manual review jobs can now be retrieved with a lock token for secure review workflows.
  * Job requests containing more than 10 IDs are rejected; empty ID lists return no jobs.

* **Bug Fixes**
  * Manual review queues, jobs, and favorites are restricted to queues the signed-in user can access.
  * Unauthorized queue access is rejected instead of allowing jobs to be viewed, dequeued, or favorited.
  * Existing jobs are limited to selected review queues.
  * Organization queue listings respect user permissions and organization membership.
  * Unauthenticated access returns an authentication error.

### DIFF
--- a/server/graphql/modules/manualReviewTool.resolver.test.ts
+++ b/server/graphql/modules/manualReviewTool.resolver.test.ts
@@ -1,0 +1,199 @@
+import { UserPermission } from '../../services/userManagementService/index.js';
+import { resolvers } from './manualReviewTool.js';
+
+type ResolverFn = (
+  parent: unknown,
+  args: unknown,
+  ctx: unknown,
+) => Promise<unknown>;
+
+const Query = resolvers.Query as Record<
+  'getTotalPendingJobsCount' | 'manualReviewQueue',
+  ResolverFn
+>;
+const Mutation = resolvers.Mutation as Record<
+  'dequeueManualReviewJob',
+  ResolverFn
+>;
+const ManualReviewQueue = resolvers.ManualReviewQueue as Record<
+  'jobs',
+  ResolverFn
+>;
+
+function makeCtx(opts: {
+  reviewableQueueIds: string[];
+  user?: {
+    id: string;
+    orgId: string;
+    permissions: readonly UserPermission[];
+  } | null;
+}) {
+  const user =
+    opts.user === undefined
+      ? { id: 'user-1', orgId: 'org-1', permissions: [UserPermission.VIEW_MRT] }
+      : opts.user;
+
+  const getReviewableQueuesForUser = jest.fn(async () =>
+    opts.reviewableQueueIds.map((id) => ({ id, orgId: 'org-1', name: id })),
+  );
+  const getAllQueuesForOrgAndDangerouslyBypassPermissioning = jest.fn(
+    async () => {
+      throw new Error('resolver must not bypass permissioning (#1150)');
+    },
+  );
+  const getQueueForOrgAndDangerouslyBypassPermissioning = jest.fn(async () => {
+    throw new Error('resolver must not bypass permissioning (#1150)');
+  });
+  const getTotalPendingJobCountForQueues = jest.fn(async () => 7);
+  const dequeueNextJob = jest.fn(async () => null);
+  const getAllJobsForQueue = jest.fn(async () => []);
+
+  const ctx = {
+    getUser: () =>
+      user == null
+        ? null
+        : {
+            id: user.id,
+            orgId: user.orgId,
+            getPermissions: () => user.permissions,
+          },
+    services: {
+      ManualReviewToolService: {
+        getReviewableQueuesForUser,
+        getAllQueuesForOrgAndDangerouslyBypassPermissioning,
+        getQueueForOrgAndDangerouslyBypassPermissioning,
+        getTotalPendingJobCountForQueues,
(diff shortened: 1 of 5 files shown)


server/graphql/modules/manualReviewTool.ts
```diff
@@ -1748,6 +1748,11 @@ const NcmecManualReviewJobPayload: GQLNcmecManualReviewJobPayloadResolvers = {
 
 const ManualReviewQueue: GQLManualReviewQueueResolvers = {
   async jobs(queue, { ids: jobIds, limit }, context) {
+    const user = context.getUser();
+    if (user == null) {
+      throw unauthenticatedError('User required.');
+    }
```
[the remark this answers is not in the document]
cassidyjames: Addressed in #1152

server/services/manualReviewToolService/modules/QueueOperations.test.ts
```diff
@@ -188,6 +188,63 @@ describe('QueueOperations', () => {
     },
   );
 
+  // Regression: #1150 -- MRT queue resolvers used to resolve
```
taobojlen, line 191: IMO: LLMs write a lot of code comments about how code *used* to be, and it's usually not all that helpful for future readers. they can always git blame if needed!

server/services/manualReviewToolService/modules/QueueOperations.test.ts
```diff
@@ -188,6 +188,63 @@ describe('QueueOperations', () => {
...
+  // queues via *Dangerously*BypassPermissioning helpers with no permission or
+  // membership check, so any authenticated user could read (and dequeue/lock)
+  // every queue in the org, including CSAM/NCMEC queues. They now call
+  // getReviewableQueuesForUser instead; these lock in its filtering.
+  const invoker = (
```
taobojlen, line 196: nit: this function makes it ever so slightly harder to read the tests, i think, because it's an extra layer of indirection for little benefit

server/graphql/modules/manualReviewTool.resolver.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  const getReviewableQueuesForUser = jest.fn(async () =>
+    opts.reviewableQueueIds.map((id) => ({ id, orgId: 'org-1', name: id })),
+  );
+  const getAllQueuesForOrgAndDangerouslyBypassPermissioning = jest.fn(
+    async () => {
+      throw new Error('resolver must not bypass permissioning (#1150)');
+    },
+  );
+  const getQueueForOrgAndDangerouslyBypassPermissioning = jest.fn(async () => {
+    throw new Error('resolver must not bypass permissioning (#1150)');
```
taobojlen, line 45: similar to another comment; i don't know if it's all that useful to reference a specific github issue given that git blame exists, but that's just my 2 cents and not a strongly held opinion!

taobojlen (approved):
some minor nits but nothing blocking!

server/services/manualReviewToolService/modules/QueueOperations.test.ts
cassidyjames: Agree; I think this is worth explicitly calling out in AGENTS.md

server/graphql/modules/manualReviewTool.resolver.test.ts
cassidyjames:
```suggestion
    throw new Error('resolver must not bypass permissioning');
```

server/services/manualReviewToolService/modules/QueueOperations.test.ts
cassidyjames:
```suggestion
```

juanmrad:
@coderabbitai review

juanmrad (approved):

PR #1152, opened 2026-09-09
Additional hardening for queue access controls

cassidyjames:
Follow-up to #1150

ManualReviewQueue.jobs checked that someone was logged in, but not that the caller could actually review the queue it was handed; pendingJobCount and oldestJobCreatedAt had no check at all. Queue objects can reach these resolvers from parents that don't establish membership themselves, so we could still leak jobs and queue metadata.

- User.favoriteMRTQueues keeps a favorite after access is revoked, and Org.users is readable by any authenticated org member, so `org { users { favoriteMRTQueues { jobs } } }` needed no permissions at all.

- RoutingRule.destinationQueue bypasses queue permissions for EDIT_MRT_QUEUES holders, while getReviewableQueuesForUser returns nothing without VIEW_MRT, so that permission combination could reach a queue it had no queues for.

Add assertQueueIsReviewable and apply it to all three fields.

The reviewable-queue lookup is memoized on the GraphQL context rather than called per field to avoid two full queue fetches per queue when loading the dashboard.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved access control for manual review queues and related data.
  - Restricted existing-job searches to queues the viewer can review.
  - Prevented unauthorized access to queue jobs, counts, timestamps, reviewers, hidden actions, and report-clearing actions.
  - Improved handling when no reviewable queues are available.

- **Tests**
  - Expanded coverage for queue authorization and queue-scoped job searches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ For more information about each release including git tags and artifacts, see [R
 - Scylla healthcheck probing gossip instead of CQL readiness ([#1041](https://github.com/roostorg/coop/pull/1041) by [@reitblatt](https://github.com/reitblatt))
 - Podman Compose setup failing to resolve the backend ([#981](https://github.com/roostorg/coop/pull/981) by [@juanmrad](https://github.com/juanmrad))
 - DB Migrator rejecting non-standard Scylla ports ([#878](https://github.com/roostorg/coop/pull/878) by [@jess-upscrolled](https://github.com/jess-upscrolled))
-- Review queue and job access control hardening ([#1151](https://github.com/roostorg/coop/pull/1151) by [@serendipty01](https://github.com/serendipty01) and [@cassidyjames](https://github.com/cassidyjames)
+- Review queue and job access control hardening ([#1151](https://github.com/roostorg/coop/pull/1151) by [@serendipty01](https://github.com/serendipty01) and [@cassidyjames](https://github.com/cassidyjames))
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ For more information about each release including git tags and artifacts, see [R
 - Scylla healthcheck probing gossip instead of CQL readiness ([#1041](https://github.com/roostorg/coop/pull/1041) by [@reitblatt](https://github.com/reitblatt))
 - Podman Compose setup failing to resolve the backend ([#981](https://github.com/roostorg/coop/pull/981) by [@juanmrad](https://github.com/juanmrad))
 - DB Migrator rejecting non-standard Scylla ports ([#878](https://github.com/roostorg/coop/pull/878) by [@jess-upscrolled](https://github.com/jess-upscrolled))
-- Manual review queues and jobs being readable by users without access to that queue ([#1151](https://github.com/roostorg/coop/pull/1151) by [@serendipty01](https://github.com/serendipty01), closes [#1150](https://github.com/roostorg/coop/issues/1150))
+- Manual review queue and job access control hardening ([#1151](https://github.com/roostorg/coop/pull/1151) by [@serendipty01](https://github.com/serendipty01) and [@cassidyjames](https://github.com/cassidyjames)
 
 ### Security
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ For more information about each release including git tags and artifacts, see [R
 - Scylla healthcheck probing gossip instead of CQL readiness ([#1041](https://github.com/roostorg/coop/pull/1041) by [@reitblatt](https://github.com/reitblatt))
 - Podman Compose setup failing to resolve the backend ([#981](https://github.com/roostorg/coop/pull/981) by [@juanmrad](https://github.com/juanmrad))
 - DB Migrator rejecting non-standard Scylla ports ([#878](https://github.com/roostorg/coop/pull/878) by [@jess-upscrolled](https://github.com/jess-upscrolled))
+- Review queue and job access control hardening ([#1151](https://github.com/roostorg/coop/pull/1151) by [@serendipty01](https://github.com/serendipty01) and [@cassidyjames](https://github.com/cassidyjames)
```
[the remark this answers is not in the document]
cassidyjames:
```suggestion
- Review queue and job access control hardening ([#1151](https://github.com/roostorg/coop/pull/1151) by [@serendipty01](https://github.com/serendipty01) and [@cassidyjames](https://github.com/cassidyjames))
```

server/graphql/modules/manualReviewTool.resolver.test.ts
```diff
@@ -180,20 +308,232 @@ describe('MRT queue/job resolvers are membership-scoped', () => {
...
+      ).rejects.toThrow('User does not have access to this queue');
+      expect(getPendingJobCount).not.toHaveBeenCalled();
+    });
+
+    it('oldestJobCreatedAt refuses a queue the caller cannot review', async () => {
+      const { ctx, getOldestJobCreatedAt } = makeCtx({
+        reviewableQueueIds: ['q-1'],
+      });
+      await expect(
+        ManualReviewQueue.oldestJobCreatedAt(
```
taobojlen, line 403: our graphQL API is a bit weird here, which means it's a lot of extra work to do these permission checks!

a standard graphQL would look more like

```
query {
  queue(id: $ID!) {
    pendingJobCount
    oldestJobCreatedAt
    # and so on...
  }
}
```

then you'd only need to do the permission check once, in the `queue` resolver.

anyway, just a thought, i realize it's out of scope for this PR.

server/graphql/modules/manualReviewTool.ts
```diff
@@ -1746,12 +1748,72 @@ const NcmecManualReviewJobPayload: GQLNcmecManualReviewJobPayloadResolvers = {
   },
 };
 
+/**
+ * The queue-scoped fields below each resolve one queue at a time, but the MRT
+ * dashboard asks for them on every reviewable queue at once. The GraphQL
+ * context is a fresh object per request, so memoizing on it keeps the
+ * reviewability lookup to one query per request instead of one per queue.
```
taobojlen, line 1755: this would be better resolved by refactoring the API (see my other comment).

IMO we could also exclude the memoization here, for simplicity, and then refactor the API in a follow-up.

server/graphql/modules/manualReviewTool.ts
```diff
@@ -1746,12 +1748,72 @@ const NcmecManualReviewJobPayload: GQLNcmecManualReviewJobPayloadResolvers = {
...
+  Promise<Set<string>>
+>();
+
+// The cache is populated before the first await, so concurrent field resolvers
+// in the same request share one in-flight lookup rather than racing.
+async function getReviewableQueueIds(
+  context: Context,
+  user: GraphQLUserParent,
+) {
+  const cached = reviewableQueueIdsByRequest.get(context);
```
taobojlen, line 1768: hmm, how does cache invalidation work here? what if the server runs for a long time, and permissions change -- we'd need to invalidate this cache in that scenario, right?

i'd suggest removing the caching for now and refactoring the API instead. (you can assign that to me if you agree!)

taobojlen (changes requested):
i think there may be a cache invalidation bug here. but there's also a broader question about whether we can avoid the complexity of caching by making our API more standard graphQL-like!

cassidyjames:
@taobojlen thanks for the review! I think some of this got handled directly in #1151, and there's a bunch of conflicts here now, so I'm going to close this PR. Please feel free to open any additional improvements or refactors separately. :)

PR #1156, opened 2026-09-10
Add poutine supply-chain scan for GitHub Actions workflows

reitblatt:
## Context & Requests for Reviewers

Adds [poutine](https://github.com/boostsecurityio/poutine) (BoostSecurity) as a CI supply-chain scanner for our GitHub Actions pipelines, wired into CI and gated so it stays green in steady state.

Poutine is complementary to the `zizmor` job we already run rather than duplicating it.

**zizmor** analyzes workflow files in this repo: template injection,
credential persistence, cache poisoning, unpinned actions. Broad rule
coverage, fast, auto-fix, and it runs as a blocking PR gate.

**poutine** analyzes those files *reach*. It resolves 
workflows and actions transitively, so it flags an unpinned
action pulled in through a dependency (e.g. the [Shai-Hulud](https://hivesecurity.gitlab.io/blog/shai-hulud-github-actions-supply-chain-attack/))
attack, which is invisible to a tool that only reads our own YAML.

poutine is only advisory and does not block merges. Findings land in the code scanning tab.

This + zizmor should address CI scanning in #213 

### What's here

- **`boostsecurityio/poutine-action`** — upstream's official action does the scanning. 
- **Pinned to `main@badd750a`, not the latest tag.** The newest tag (`v1.1.4`, Apr 2026) still builds from `poutine:1.1.4`; `main@badd750a` builds from `poutine:1.1.6@sha256:722a8e09…`.  Pinning tag refs vs hashes is literally what these tools warn you about...
- **`.github/workflows/poutine.yml`** — Runs on a weekly cron (new poutine rules/advisories can flag workflows nobody has touched) and `workflow_dispatch`.

### Current baseline

0 findings at `warning` or `error`. 

7 **note**-level `github_action_from_unverified_creator_used` findings:
* `dorny/paths-filter` (×3), 
* `dorny/test-reporter` (×2), 
* `zizmorcore/zizmor-action`, 
* `MatteoGabriele/agentscan-action` 

All already commit-SHA pinned. 

## Tests

- Ran poutine locally: zero `warning`/`error` findings; the gate's `jq` filter returns 0.

## (Optional) Rollout Plan

None: CI-only change.

## Checklist

_Only check items that apply to this PR; leave the rest unchecked._

- [ ] **If you changed anything user-facing** (i.e. user interface or APIs):
  Did you update related docs?

- [ ] **If the change is notable** (refer to [Keep a Changelog](https://keepachangelog.com/en/2.0.0/) conventions):
[...]

PR #1168, opened 2026-09-10
ci: track the github-actions ecosystem in Dependabot

reitblatt:
Adds a `github-actions` ecosystem to `.github/dependabot.yml`, mirroring `roostorg/osprey`, so action pins are maintained automatically instead of by periodic manual passes. All updates land in one grouped weekly PR with a 7-day cooldown.

### The semver-major decision (please read)

Every other ecosystem in this file ignores `version-update:semver-major`. **This block deliberately does not.**

Reasoning: action majors are overwhelmingly runner-runtime bumps — `actions/upload-artifact` v5 -> v6 and `actions/cache` v4 -> v5 exist almost entirely to move from Node 20 to Node 24, which GitHub's runners force on us anyway. If majors were ignored here, the exact situation this stack of PRs is fixing (a `node20` pin sitting on `main` emitting a deprecation annotation on every run) would silently recur, and the config would be self-defeating. Actions are also SHA-pinned and every bump is reviewed in a PR, so a bad major cannot land silently.

The trade-off is accepted noise: an occasional major that needs a call-site change (e.g. `actions/checkout` v7 blocking fork-PR checkouts) will show up as a red Dependabot PR rather than never being offered. That is the cheaper failure mode.

I updated the file's header comment so "majors are disabled across every ecosystem" is no longer inaccurate.

### Also documented

A note that Dependabot bumps the action SHA only: tool versions passed *into* an action still need a human — the `version:` input of `zizmorcore/zizmor-action`, and the image digest baked into `boostsecurityio/poutine-action` (arriving in #1156).

### CHANGELOG

This PR carries the single `### CI & infrastructure` entry for the whole pin-bump stack; the individual bump PRs are CI-only and do not touch `CHANGELOG.md`. It will conflict trivially with #1156, which introduces the same heading.

### Verification

- SHA re-derived from the release tag via `git/ref/tags` -> `git/tags` (annotated tags dereferenced), not copied from the tag ref.
- `zizmor` 1.25.2 (the version this repo pins): no findings.

Split out of a single pin-bump pass; sibling PRs cover the other actions.

## Summary by CodeRabbit

* **Chores**
  * Added weekly automated monitoring for GitHub Actions dependencies.
  * Enabled major-version updates for GitHub Actions dependencies.
  * Grouped GitHub Actions dependency updates and applied consistent labeling.

* **Documentation**
  * Updated the unreleased changelog with the new dependency monitoring and action version updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,10 @@ For more information about each release including git tags and artifacts, see [R
 - Passwords are hashed with Argon2id instead of bcrypt at cost factor 5 ([#901](https://github.com/roostorg/coop/pull/901) by [@serendipty01](https://github.com/serendipty01), closes [#900](https://github.com/roostorg/coop/issues/900))
 - Minimum password length raised to 15 and enforced server-side ([#1065](https://github.com/roostorg/coop/pull/1065), [#1094](https://github.com/roostorg/coop/pull/1094) by [@serendipty01](https://github.com/serendipty01))
 
+### CI & infrastructure
+
+- Dependabot now tracks the `github-actions` ecosystem, and every action pin was bumped to its current release
+
 ## [1.0.2] - 2026-06-30
 
 This release addresses reported security advisories, improves NCMEC CyberTipline reporting, and includes front-end quality-of-life improvements.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ For more information about each release including git tags and artifacts, see [R
 
 ### CI & infrastructure
 
-- Dependabot now tracks the `github-actions` ecosystem, and every action pin was bumped to its current release
+- Dependabot now tracks the `github-actions` ecosystem, and every action pin was bumped to its current release ([#1157](https://github.com/roostorg/coop/pull/1157)–[#1168](https://github.com/roostorg/coop/pull/1168) by [@reitblatt](https://github.com/reitblatt))
 
 ## [1.0.2] - 2026-06-30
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,10 @@ For more information about each release including git tags and artifacts, see [R
 - Passwords are hashed with Argon2id instead of bcrypt at cost factor 5 ([#901](https://github.com/roostorg/coop/pull/901) by [@serendipty01](https://github.com/serendipty01), closes [#900](https://github.com/roostorg/coop/issues/900))
 - Minimum password length raised to 15 and enforced server-side ([#1065](https://github.com/roostorg/coop/pull/1065), [#1094](https://github.com/roostorg/coop/pull/1094) by [@serendipty01](https://github.com/serendipty01))
 
+### CI & infrastructure
+
+- Dependabot now tracks the `github-actions` ecosystem, and every action pin was bumped to its current release ([#1157](https://github.com/roostorg/coop/pull/1157)–[#1168](https://github.com/roostorg/coop/pull/1168) by [@reitblatt](https://github.com/reitblatt))
+
```
cassidyjames: CHANGELOG.md should follow [Keep a Changelog](https://keepachangelog.com/en/2.0.0/), which means [curated](https://keepachangelog.com/en/2.0.0/#curation) notable changes, _only_ under `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, or `Security` headings.

I don't think this change is notable enough to be included, but if so, it probably belongs under the `Security` heading.

```suggestion
```

cassidyjames (approved):

PR #1189, opened 2026-09-13
Add a server-side content resolver for manual review jobs

### LINKED ISSUE #1188: Support refreshing expiring media URLs for active manual reviews
## Problem

Manual review jobs can remain queued longer than the signed URL stored with their media. When the URL expires, the underlying object still exists but the reviewer can no longer render it.

Making the original URL longer-lived is not always possible. S3, GCS and other storage providers often sign URLs with temporary credentials.

## Proposed solution

Add an optional server-side content resolver for active manual reviews.

When the review page loads:

1. The existing job query includes the review lock token.
2. Coop loads the job from a queue the authenticated reviewer can access.
3. If the lock token belongs to that reviewer, Coop passes the server-loaded job to the configured resolver.
4. The resolver returns a copy of the job with refreshed media URLs.
5. The existing client renders the resolved job.

Deployments would register the resolver through `getBottle({ manualReviewContentResolver })`. Without a resolver, Coop would return the stored job unchanged.

```ts
type ManualReviewContentResolver = (input: {
  orgId: string;
  queueId: string;
  reviewerId: string;
  job: ManualReviewJobOrAppeal;
}) => Promise<{
  job: ManualReviewJobOrAppeal;
  resolvedContentCount: number;
}>;
```

Storage-specific code would remain outside Coop. For example, a deployment could replace an expired S3 URL with a fresh short-lived URL using its current role credentials.

## Safety and compatibility

- Resolution runs only for an authenticated reviewer with the active review lock.
- The resolver receives the job loaded by Coop, not a URL supplied separately by the browser.
- Submitted item data cannot select resolver endpoints.
- Resolver output must retain the same job and organization.
- Resolver errors or invalid output fall back to the stored job.
- Shared and historical views keep their current behavior.
- Trace metadata may record the reviewer, job, queue and resolved-content count, but not media URLs.
- The contract does not depend on a specific cloud or storage provider.

## Acceptance criteria

- Existing deployments behave as they do today when no resolver is configured.
- Reloading an active review requests fresh content resolution.
- The current media components render the resolved job without another client request.
- Tests cover authorization, no-op behavior, successful resolution, failure fallback and invalid output.

## Out of scope

- Storage-provider implementations
- Refreshing media without a page reload
- Proving that the browser rendered every media byte

sunilatlas:
# Add a server-side content resolver for manual review jobs

## Problem

Manual review jobs may outlive signed media URLs stored in their payload. When a reviewer opens the job later, the media can no longer be rendered even though the underlying object still exists.

## Change

This adds an optional `ManualReviewContentResolver` extension to `getBottle`.

When an active review page loads, the existing job query sends its review lock token. Coop resolves content only when the token belongs to the authenticated reviewer in the same organization. The resolver receives the job loaded by the server, not a URL supplied separately by the browser.

Deployments without a resolver keep the existing behavior. Resolver errors and invalid results fall back to the stored job.

The review client does not add another request or change its media components. Reloading the page runs resolution again through the existing job query.

## Safety

- Submitted item data cannot select a resolver endpoint.
- Resolver output must keep the same job and organization.
- Shared links, history pages and mismatched review sessions use the stored job unchanged.
- Trace attributes include job, queue, reviewer and resolved-content count, but not media URLs.
- Storage-specific behavior remains outside Coop.

## Validation

- GraphQL client and server types regenerated
- Server and client typechecks pass
- Focused resolver tests: 8 passed
- Client suite: 237 passed, 2 skipped
- Server production build passes
- Client production build passes
- Changed files pass ESLint and formatting checks

## Out of scope

- AWS, GCS or another storage-provider implementation
- Refreshing content without a page reload
- Proving that the browser rendered every media byte

Closes #1188

## Summary by CodeRabbit

* **New Features**
  * Manual review jobs can securely resolve review content using a lock token.
  * Resolved content counts are available during review processing.
  * Review content resolution supports safe fallback to stored content when resolution fails.
  * Review job requests now support up to 10 job IDs.

* **Bug Fixes**
  * Improved authorization for locked review content.
  * Review locks are extended during authorized content resolution.

* **Tests**
  * Added coverage for authorization, lock handling, resolution, validation, fallback behavior, and errors.


server/services/manualReviewContentResolver.ts
```diff
@@ -0,0 +1,79 @@
...
+  hasActiveLock: () => Promise<boolean>;
+}) {
+  if (
+    opts.lockToken !== opts.reviewerId ||
+    opts.jobOrgId !== opts.reviewerOrgId
+  ) {
+    return false;
+  }
+  return opts.hasActiveLock();
+}
```
juanmrad, line 52:
```suggestion
export async function canResolveManualReviewContent(opts: {
  jobOrgId: string;
  reviewerOrgId: string;
  hasActiveLock: () => Promise<boolean>;
}) {
  if (opts.jobOrgId !== opts.reviewerOrgId) {
    return false;
  }
  return opts.hasActiveLock();
}
```

There is no need to check reviewer id here. BullMQ's extendLock already rejects mismatched tokens. if the token doesn't match the job's actual lock, it returns `false` and `canResolveManualReviewContent` returns false. The org check stays since that's not validated by BullMQ.

server/graphql/modules/manualReviewTool.ts
```diff
@@ -1763,12 +1763,35 @@ const ManualReviewQueue: GQLManualReviewQueueResolvers = {
...
+        context.services.ManualReviewToolService.resolveContentForReview({
+          job,
+          queueId,
+          reviewerId: user.id,
+          reviewerOrgId: user.orgId,
+          lockToken,
+          isAppealsQueue: queue.isAppealsQueue,
+        }),
+      ),
+    );
```
juanmrad, line 1806: This can be unbounded. in practice I believe JobIds only has 1, but this could also lead to problems if too many are passed. we could bound this to maybe 10 or so at most to avoid issues.

juanmrad (changes requested):
Thanks for contributing left two comments let me know what you think. I can merge once addressed :)

server/services/manualReviewContentResolver.ts
sunilatlas: Thanks. I think we need to keep this check. lockToken is client-provided, and Coop uses the reviewer’s user ID as the token, so it is not a secret. extendLock only proves that the supplied token owns the lock. Comparing it with the authenticated reviewer binds that lock to the caller.

server/graphql/modules/manualReviewTool.ts
sunilatlas: ACK. Updated PR.

juanmrad (approved):
LGTM

PR #1190, opened 2026-09-13
Remove unreachable RedisStore from the cache library

ThisIsMissEm:

`RedisStore` has never been constructed in this repository's history. `cached()` in server/utils/caching.ts is the only place that builds a `Cache`, and it hardcodes `MemoryStore`; I haven't been able to find any usages of `RedisStore` in the project nor its history.

The only reference anywhere was the barrel re-export in `lib/cache/index.ts`.

Deleting it removes 596 lines of TypeScript and two Lua scripts, and takes two dependencies with it:

- `pipeline-segment`, imported only by RedisStore
- `copyfiles` and the `copy-assets` script, which existed solely to copy those two `.lua` files into `transpiled/`

It was also the only place in `server/` importing ioredis's compiled internals (`ioredis/built/Script.js`, `ioredis/built/transaction.js`). Those are not public API and carry no semver guarantee, so this removes an unguarded coupling to another package's build output.

No Redis-side cleanup is required: no cache keys were ever written and no Lua was ever loaded. Redis's script cache is volatile regardless — not persisted to RDB/AOF, cleared on restart. After this change `server/` has no `SCRIPT LOAD`, `evalsha`, or `defineCommand` call sites at all; BullMQ manages its own Lua independently.

No CHANGELOG entry: unreachable internal code, no behavioural change.

## Tests

Removal of dead code, so no tests necessary.

## (Optional) Rollout Plan

I don't think a rollout plan is applicable as this code appears to have never been exercised.

## Checklist

  Did you update related docs? Minor docs change related to the removal of `copy-assets` which is no longer needed.

  Did you update CHANGELOG.md?

## Summary by CodeRabbit

- **Chores**
  - Removed the Redis-backed cache store and its associated cleanup scripts.
  - Simplified build and startup commands by removing the separate static-asset copying step.
  - Removed unused supporting packages related to asset copying and Redis cache processing.

- **Documentation**
  - Updated the development quick-start guide to reflect the shorter setup process and revised step numbering.


juanmrad (approved):
LGTM

juanmrad (approved):
LGTM

taobojlen:
CleanShot 2026-09-14 at 17 41 16@2x

beautiful

julietshen:
wow beauty

Issue #671, opened 2026-06-02
Remove PR checklist action in favor of PR template


cassidyjames:
We have this checklist action:

https://github.com/roostorg/coop/blob/dc279b7b9ce88f7232413678f115d762b8054da6/.github/workflows/checklist.yaml#L1-L19

…and its associated config stored as JSON in the workflows directory:

https://github.com/roostorg/coop/blob/dc279b7b9ce88f7232413678f115d762b8054da6/.github/workflows/checklistConfig.json#L1-L27

It doesn't work for PRs originating from forks, since it uses a GitHub token to write a comment. It seems like this would just be better handled by updating the PR template, anyway, rather than adding an action run.

PR #420, opened 2026-05-08
[Kysely] Demolish server/models/, flip Rule codegen mappers, drop sequelize from server/

juanmrad:
## Context & Requests for Reviewers

Final server-side step of the Sequelize -> Kysely migration. The production runtime, fixtures, and bin scripts have all been off Sequelize for a while. This PR removes the last type-only references, the IoC wiring, and the `sequelize` dependency itself.

After this PR: `grep -R "from 'sequelize'\|Dependencies\['Sequelize'\]\|deps\.Sequelize\|container\.Sequelize" server/` returns zero hits, `server/models/` no longer exists, and `server/package.json` no longer pulls in `sequelize` (or its `clshooked` transitive).

### Deletions
- 11 Sequelize model files: `OrgModel`, `UserModel`, `PolicyModel`, `rules/{ActionModel,BacktestModel,ItemTypeModel,RuleLatestVersionModel,RuleModel}`, `banks/{LocationBankModel,LocationBankLocationModel}`, plus `models/index.ts` and `models/sequelizeSetup.ts`.
- `models/errors.ts`.
- The `server/models/` directory itself.

### IoC cleanup (`server/iocContainer/index.ts`)
- Dropped `makeDb` import.
- Dropped `Sequelize`, `OrgModel`, `RuleModel`, `ActionModel`, `PolicyModel`, `ItemTypeModel`, `LocationBankModel`, `LocationBankLocationModel` from `Dependencies`.
- Dropped the 8 `bottle.factory(...)` registrations.
- Dropped `'Sequelize'` from `servicesThatCanBeShutdown` and removed the per-model entries from the closable-services exclude list.

PR #1191, opened 2026-09-13
Remove PR checklist item relating to models

ThisIsMissEm:

`server/models` was removed in the migration from sequelize to kysely. We may still want a line for ensure that versioned tables are correctly modified in lockstep, but the existing PR checklist item was incorrect.

What we could actually do here is remove the need for the checklist item, and move it to a CI check instead: For every table carrying a versioning_trigger, assert its _history table has an identical column set.

It's one query against `pg_trigger` / `pg_get_triggerdef` + `information_schema.columns`, which runs after the migrations are already execute in the migrations compose service, it covers all ten tables automatically including ones added later, and it fails the PR rather than reminding someone. Claude estimates "maybe twenty lines of SQL".


ThisIsMissEm:
Checked whether the concern behind this item is still live before it disappears.

**It isn't currently, so this is safe to merge as-is.** Since the initial schema (`2025.12.01T00.00.00.initial-schema.sql`), no migration has added or dropped a column on any temporally-versioned base table, so every `_history` table is still in lockstep with its parent.

Two things worth recording for whoever picks this up, though.

**The item was incomplete as well as stale.** It named four models, but there are eight tables carrying a `versioning_trigger`: `rules`, `actions`, `item_types`, `policies`, plus the join tables `rules_and_actions`, `rules_and_item_types`, `rules_and_policies`, and `actions_and_item_types`.

**The underlying hazard is real, but it moved.** `versioning()` copies rows into `<table>_history` on every write. If someone adds a column to a versioned table without adding it to the matching `_history` table, the trigger breaks or silently drops the value. Back when Sequelize models were the schema source of truth, "did you change a model file?" was a reasonable proxy for that. Since #420 removed `server/models/` and Kysely took over — with the database maintaining history itself — the only thing still needing human attention is a *column* change, and those live in `db/src/scripts/**`, not in model files.

Which means the existing migrations item is arguably the right home for it:

> - [ ] **If you changed `db/src/scripts/**` and used `CREATE TABLE`, `ADD COLUMN`, or `ALTER COLUMN`:** …

with something appended like:

> If the table is temporally versioned (it has a `versioning_trigger`), make the same column change to its `_history` table.

No strong opinion on whether that's worth adding — flagging it mainly so the reminder isn't lost silently. @cassidyjames your call.

ThisIsMissEm:
Correcting myself — my earlier comment was wrong on the central point, and in the direction that undersells the case for keeping a reminder.

I claimed no migration since the initial schema had added a column to a versioned table. That was a bad check on my part: a single-line grep against multi-line `ALTER TABLE` statements, and scoped to `rules_history` rather than all history tables. **Three migrations have done it:**

- `2026.05.21T16.50.30.add_ip_address_field_role_to_item_types.sql`
- `2026.06.14T02.33.06.add-action-parameters-to-rules-and-strikes.sql`
- `2026.06.25T20.44.03.add_email_field_role_to_item_types.sql`

The conclusion still holds — everything is in lockstep — but for the opposite reason to the one I gave. Not because it never comes up, but because it comes up every couple of months and each author remembered.

It's also more than the history table. Adding one column to `item_types` meant touching six objects:

```
ALTER TABLE public.item_types              ADD COLUMN
ALTER TABLE public.item_types_history      ADD COLUMN
DROP / CREATE VIEW public.item_type_latest_versions
DROP / CREATE MATERIALIZED VIEW public.item_type_versions
(plus ownership ALTERs on both recreated objects)
```

That exact sequence has been done twice for `item_types` alone. `rules_and_actions` was simpler — base table plus history, no views.

And correcting my count: it's **ten** versioned tables, not eight. I missed `manual_review_tool.routing_rules` and `manual_review_tool.appeals_routing_rules`, whose triggers register with schema-qualified quoted names.

So I'd revise "no strong opinion" to: there's a reasonable case for a reminder existing, and for it naming the view and materialized-view steps rather than just the history table. Removing the old item is still right regardless — it pointed at `server/models/`, which is the one place not involved.

juanmrad (approved):
👍🏻

cassidyjames:
@ThisIsMissEm we moved to the PR checklist in favor of a CI run that had a bot comment on things for #671, buuuut I'm not opposed to doing things better in CI, instead. I'm also fine to remove this checklist item in the interim! I'll open your suggestion as a new issue so it doesn't block this but doesn't get lost, either.

PR #1193, opened 2026-09-14
Remove usage of cached from MRT queues

ThisIsMissEm:

Building on #1190, this refactors the `QueueOperations` class to not use `cached` in favour of an in-memory collection of queues and workers, via `KeyedResourceRegistry` which is a facade on top of a basic `Map`. We weren't actually caching data with `cached`, and `cached` is already in memory, instead it was being used for some side-effects:

1. The number of `Worker` and `Queue` objects held in memory was limited to 128 per type (`bullQueues`, `bullWorkers`, `bullAppealQueues`, `bullAppealWorkers` — so a total of 512)
2. The `freshUntilAge: 600` directive automatically cleaned up `Worker` or `Queue` instances after 10 minutes of existence.

Both of these side-effects meant that there were potentially some bugs lurking here:
- Claimed-but-unactioned jobs would not return to the queue once the queue was evicted from memory, as the stalled job checker would be lost with closing the worker. The claimed-but-unactioned jobs would only be returned to the queue once something next requests that queue
- No Worker or Queue instance survived more than ten minutes, which when moderators are actively working queues could be a real problem.
- Leakage of queues and workers in tests (we weren't tearing them down properly)
- Leakage of queues and workers in production as the underlying implementation was actually overwriting the instance each time it expired, without calling the `onItemEviction` method (which closed the worker or queue)

Further, I've taken the chance to improve `makeTransactionalTestWithFixture` allowing it to return an optional `cleanup` method that runs before the server is stopped and cleaned up, ensuring we're not leaking queues between tests (at least within `QueueOperations.test.ts`) — that's what uses the org-wide queue cleanup provided by `obliterateAllQueuesForOrg`.

I've measured the retained connections to redis and we aren't leaking redis connections with this change, since nothing calls `run` and the `autorun` option is disabled on the Worker. The only thing that is retained rather than evicted now is the stalled job checker which is a simple timer and we actually do want it to remain (I think).

One thing we loose here is that deleted queues can still exist cross-process boundaries, which was I think accidentally solved by the usage of the LRU/max-age, not a deliberate design decision. 

## Longer term rationale

We had this custom half-used cache implementation, which really doesn't make sense for us to be maintaining as there are plenty of packages for doing caching, such as [bentocache](https://bentocache.dev) (which is actually the module underlying Adonis.js's cache module).

This is the only call site which we couldn't move across to an external cache package, hence taking the time to clean it up.

## Tests

Newly added code has test coverage and I've amended some tests to ensure they correctly clean up resources created during the test. Whilst working on this, I did ensure to run the entire test suite to ensure no regressions were introduced.

## (Optional) Rollout Plan

I do not think there's a need for any rollout plan, as `cached` is in-memory anyway.

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md? Yes, added a note about the claimed but unactioned jobs returning to the queue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where claimed but unactioned jobs could stop re-entering queue rotation when an organization had more than 128 active review queues.
  - Improved queue cleanup to preserve review queue records while safely clearing associated processing resources.

- **Reliability**
  - Improved handling of concurrent queue access and long-lived queue resources, reducing unexpected resource eviction and related processing interruptions.


server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -444,7 +456,7 @@ export default class QueueOperations {
     if (queueId === defaultQueueId) {
       throw makeUnableToDeleteDefaultQueueError({ shouldErrorSpan: true });
     }
-    const queue = await this.getOrCreateBullQueue({ orgId, queueId });
+    const queue = await this.bullQueues.get({ orgId, queueId });
```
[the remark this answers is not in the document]
ThisIsMissEm: I've removed the `deleteAllQueuesForOrg` method since it was actually broken, now we're just cleaning up the queues in redis here, since the tests use transactions/savepoints which conflicted with doing a delete in the cleanup.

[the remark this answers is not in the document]
ThisIsMissEm: okay, so we've resolved the original issue in f5d853d

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -170,51 +191,42 @@ export default class QueueOperations {
...
+  /**
+   * Drops the Bull Queue/Worker handles for a queue that no longer exists.
+   */
+  async #forgetBullResources(key: QueueKey) {
+    await Promise.all([
+      this.bullQueues.remove(key),
+      this.bullWorkers.remove(key),
+      this.bullAppealQueues.remove(key),
+      this.bullAppealWorkers.remove(key),
+    ]);
```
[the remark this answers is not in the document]
ThisIsMissEm: The call sites are right, but they all resolve bullQueues, which bounds the impact more than the finding suggests.

The two registries whose resources hold a recurring startStalledCheckTimer() are bullWorkers and bullAppealWorkers, and both are only reachable through dequeueNextJobWithLock / dequeueNextAppealJobWithLock, which call checkQueueExists first. bullAppealQueues is likewise only reached via #getBullAppealQueue, which also checks. So what this race can orphan is a Queue — an object on the shared Redis connection, with no timer and no connection of its own.

The proposed barrier also wouldn't remove the orphan: get() would await the teardown promise and then fall through to create(), registering the same entry one tick later. The only variant that prevents it is refusing to create after removal, and that's ruled out here — obliterateAllQueuesForOrg clears Redis but deliberately leaves the queue rows in place, so a later get() for that queue is correct and must produce a fresh handle.

The gap worth tracking is a different one: the registry is per-process, so a queue deleted by another process leaves its Worker — timer and all — registered here indefinitely. #forgetBullResources can't see that, and neither could a teardown promise. That needs cross-process invalidation, which is out of scope here; opening a separate issue.

Leaving the registry as-is.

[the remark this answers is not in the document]
ThisIsMissEm: @coderabbit can you look at the comment above?

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -569,9 +589,39 @@ export default class QueueOperations {
...
+  async obliterateAllQueuesForOrg(orgId: string) {
+    const queues =
+      await this.getAllQueuesForOrgAndDangerouslyBypassPermissioning(orgId);
+
+    // Serial: an org's queues share one Redis connection.
+    for (const queue of queues) {
+      const key = { orgId, queueId: queue.id };
+      const bullQueue = await this.bullQueues.get(key);
+      await bullQueue.obliterate({ force: true });
+      await this.#forgetBullResources(key);
```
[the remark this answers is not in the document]
ThisIsMissEm: `obliterateAllQueuesForOrg` has no production caller — it's fixture cleanup for tests that roll back their own transaction. If obliterate() rejects there, Redis is broken and the test should fail loudly rather than press on; the handles get dropped by queueOps.close() at shutdown moments later. Leaving as-is.

server/test/harness/transactionalTest.ts
```diff
@@ -13,26 +13,39 @@
...
 type ServerVars = Pick<MockedServer, 'deps' | 'request'>;
 
 export function makeTransactionalTestWithFixture<
   T extends Record<string, unknown>,
->(makeFixtures: (server: ServerVars) => Promise<T> | T) {
+>(makeFixtures: (server: ServerVars) => Promise<Fixture<T>> | Fixture<T>) {
   return makeTestWithFixture<ServerVars & T>(async () => {
     const server = await makeMockedServer();
-    const fixtures = await makeFixtures({
+    const { cleanup: cleanupFixtures, ...fixtures } = await makeFixtures({
```
taobojlen, line 25: nit: what do you think of calling this `nonPostgresCleanup` or something to make it clear to callers (and LLMs) they they do not need to manually delete postgres data at the end of each test?

server/test/harness/transactionalTest.ts
```diff
@@ -13,26 +13,39 @@
...
       deps: server.deps,
       request: server.request,
     });
     return {
       deps: server.deps,
       request: server.request,
-      ...fixtures,
+      ...(fixtures as unknown as T),
```
juanmrad, line 32: normally avoid having double casting where possible :/ but I guess at least we can add the reason:

```
// TypeScript can't narrow Omit<Fixture<T>, 'cleanup'> back to T
```

Or if we can potentially avoid it by not destructure cleanup out at all. Just spread the whole fixture result and override cleanup at the end of the object literal:
```ts
const fixtureResult = await makeFixtures({
  deps: server.deps,
  request: server.request,
});
return {
  deps: server.deps,
  request: server.request,
  ...fixtureResult,
  async cleanup() {
    try {
      await fixtureResult.cleanup?.();
    } finally {
      await server.rollback();
      await server.shutdown();
    }
  },
};
```

Up to you not a huge issue.

juanmrad (approved):
minor comment, but otherwise LGTM

server/utils/keyedResourceRegistry.ts
```diff
@@ -0,0 +1,111 @@
...
+    const id = this.#keyToString(key);
+    const existing = this.#entries.get(id);
+    if (existing !== undefined) {
+      return existing;
+    }
+
+    const created = this.#create(key).catch((err: unknown) => {
+      // Don't cache a rejection: a transient failure to build the resource
+      // shouldn't poison the key for the lifetime of the process.
+      if (this.#entries.get(id) === created) {
```
taobojlen, line 51: oh this is interesting. we're defining `created` on line 48, and then we refer to it inside its own `catch` handler.

how does this work? if the `#create` promise errors, what's the value of `created` at this point?

server/utils/keyedResourceRegistry.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const key = { orgId: 'org', queueId: 'q1' };
+
+    const [a, b] = await Promise.all([registry.get(key), registry.get(key)]);
+
+    expect(a).toBe(b);
+    expect(create).toHaveBeenCalledTimes(1);
+  });
+
+  it('never evicts, however many keys are registered', async () => {
+    // Regression for the original bug: these resources were held in a cache
```
taobojlen, line 47: Let's not reference things like "the original bug" in code comments; future readers won't know what that's referring to!

server/utils/keyedResourceRegistry.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    );
+
+    expect(registry.size).toBe(200);
+    for (const resource of resources) {
+      expect(resource.close).not.toHaveBeenCalled();
+    }
+  });
+
+  it('keeps a resource indefinitely rather than expiring it', async () => {
+    // Regression: the previous implementation attached a 600s TTL, so job
```
taobojlen, line 66: IMO this test is probably unnecessary; it's a regression test but the implementation is totally new!

taobojlen (approved):
LGTM!

server/test/harness/transactionalTest.ts
ThisIsMissEm: Ideally we should be testing database code in isolation from actual services, and tests should be responsible for ensuring the database is left in a clean state.

taobojlen: Hmm, I think I disagree that individual tests should be responsible for leaving the DB in a clean state! That was how things worked when I started working on Coop but it was very inconsistently done -- it's quite hard to remember to do, and hard to get right. I added the transactional test case so now it's done automatically via DB transactions/rollbacks. (But maybe I am misunderstanding what you're saying)!

PR #1195, opened 2026-09-14
Fix scylla crashing when docker has both IPv4 and IPv6 available

ThisIsMissEm:

Whilst working on #1193, I noticed that scylla was failing to startup for me, receiving "too many argument" errors, this turned out to be due to docker desktop for mac was running a dual-stack networking setup (both IPv4 and IPv6), as I'm not sure if those containers actually support IPv6 and we only reference IPv4, I've disabled IPv6 in the networking stack.

## Tests

Tested by running the stack and checking that scylla started up correctly.

## (Optional) Rollout Plan

I don't think there's a rollout plan for this, it all depends on how people are actually deploying coop in practice.

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the container network configuration to disable IPv6 by default, preventing Scylla startup failures caused by invalid entrypoint arguments.


juanmrad (approved):
👍🏻

PR #1196, opened 2026-09-14
Pass --build to every docker compose run in AGENTS.md

ThisIsMissEm:

Whilst working on #1193, I encountered an issue with claude running the test suite, unbeknownst to me, it was actually reusing the existing container from a previous run, this meant that when I thought I'd run all the tests following changes, I'd actually run the previous containers code & tests, not the latest changes.

> `docker compose run` only builds when no image exists yet, so the documented commands reuse a stale image and report on code that isn't in the container. That falsifies the CI section's claim that each command's exit code matches the corresponding CI step's.

## Tests

n/a, changes agents.md

## (Optional) Rollout Plan

n/a

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

## Summary by CodeRabbit

- **Documentation**
  - Updated Docker Compose command examples to include the `--build` flag.
  - Clarified that Compose may otherwise run tests against a stale image.
  - Updated testing, CI, and individual-check guidance to consistently rebuild images before execution.


AGENTS.md
```diff
@@ -116,11 +118,11 @@ CI runs entirely via GitHub Actions (`.github/workflows/apply_pr_checks.yaml`).
...
-docker compose run --rm backend npm run lint
-docker compose run --rm backend npm run build
-docker compose run --rm client npm run lint
-docker compose run --rm client npm run build
-docker compose run --rm test
+docker compose run --rm --build backend npm run lint
+docker compose run --rm --build backend npm run build
+docker compose run --rm --build client npm run lint
+docker compose run --rm --build client npm run build
+docker compose run --rm --build test
```
ThisIsMissEm, line 125: Arguably this should be kept the same, though it is weird to build development tools into a baked docker image..

juanmrad: I was thinking about this as I was looking at the pr. And the `--build` is only needed because the compose services use `COPY` instead of volume mounts. 

For local dev, mounting the source directories would keep the container in sync without rebuilding. That avoids the `--build` tax entirely and makes the feedback loop faster. I think we can update `docker-compose.yaml` to mount `./server` and `./client` into the respective services instead?

ThisIsMissEm: I'm not sure if that'd actually help here, due to how volume mounts end up working with the node_modules directories and with `.dockerignore` — really I don't think we should be running lint, build, or test within the container typically in developer workflows: we already have those tools all existing locally via `npm install` — with a volume mount we'd end up with macos deps running inside a linux container which would obviously break.

The chances of drift here between the CI and local copy is next to none, bar maybe platform specific issues, since the only additional software in the container is dumb-init. Platform specific issues would be things we'd need to get fixed anyway in order to make the project contributor friendly.

juanmrad: I think I understand the concern now. Mounting `./server` over `/app` would overlay the image’s existing `/app`, including its Linux-installed `node_modules`. We could preserve those with a separate `/app/node_modules` volume or narrower source mounts, but that requires additional Compose design.

I don’t think `.dockerignore` affects runtime bind mounts. It only controls the Docker build context. Either way, with the current setup the source is baked into the images, so adding `--build` makes sense.

juanmrad (approved):

PR #1197, opened 2026-09-14
Prevent invalid worker or job name from being executed

ThisIsMissEm:

Previously it was possible to invoke `bin/run-worker-or-job.ts` on the server with an invalid worker or job name, which would attempt to do something and then crash due to an undefined function. As this is a known set, and we already declare those in `server/iocContainer/services/workersAndJobs.ts`, we can just hardcode a list and call it a day.

I should note: the dockerfile as it stands today for `coop-worker` is invalid, and would have never run successfully. I'm not sure what the deployment mechanism is intended to be here, since none is documented, see: #1198

## Tests

No prior test coverage, no easy way to add it.

## (Optional) Rollout Plan

n/a

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

Arguable if I need a docs/changelog command, I don't think this change is particularly notable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Worker and job names are now validated before background processes start.
  * Invalid names display an error with the available options and exit without starting application services.
  * The readiness check now consistently reports the service as healthy without failing due to high CPU usage. Database connectivity checks remain unchanged.


ThisIsMissEm:
@coderabbitai can you review now? The one minute has elapsed.

server/iocContainer/services/workersAndJobs.ts
```diff
@@ -48,3 +48,12 @@ export function registerWorkersAndJobs(bottle: Bottle<Dependencies>) {
     makeRetryFailedNcmecDecisionsJob,
   );
 }
+
+export const WORKERS = ['ItemProcessingWorker'];
```
taobojlen, line 52: I'd prefer not to hard-code these values here -- it'll be very easy for them to get out of sync with the code.

Instead, can we dynamically populate these consts? Perhaps something like:

```ts
const workerAndJobFactories = {
  ItemProcessingWorker: makeItemProcessingWorker,
  RunUserRulesJob: makeRunUserRulesJob,
  // etc.
} satisfies Record<string, AnnotatedFactory<WorkerOrJob>>;

export const WORKER_AND_JOB_NAMES = Object.keys(workerAndJobFactories);

export function registerWorkersAndJobs(bottle: Bottle<Dependencies>) {
  for (const name of WORKER_AND_JOB_NAMES) {
    register(bottle, name, workerAndJobFactories[name]);
  }
}
```

taobojlen (changes requested):

server/bin/run-worker-or-job.ts
```diff
@@ -2,12 +2,25 @@
 import _ from 'lodash';
 
 import getBottle, { type Dependencies } from '../iocContainer/index.js';
+import { JOBS, WORKERS } from '../iocContainer/services/workersAndJobs.js';
 import { logErrorJson } from '../utils/logging.js';
 import { type WorkerOrJob } from '../workers_jobs/index.js';
 
+// Validate before building the container, so a mistyped name exits without
+// opening Postgres, Redis and Scylla connections just to reject the argument.
```
taobojlen, line 10: Is it the end of the world if a connection gets opened and dropped again? Given that this is going to get hit quite rarely I think probably not?

server/iocContainer/services/workersAndJobs.ts
ThisIsMissEm: I thought about that but wanted to try to keep it fairly minimal to stay away from any changes to IoC

server/bin/run-worker-or-job.ts
ThisIsMissEm: No, it just adds extra latency so do the quick check first is the laziest way to implement

server/iocContainer/services/workersAndJobs.ts
ThisIsMissEm: Ideally we'd not actually have this file and instead we'd have a tool to run workers + scheduled jobs that would handle everything, which iirc is what these all almost are (see the issue in pr description)

juanmrad: I agree with tao here on avoiding the hardcoded lines. we already have a place where they are defined. If we move to something else like a tool over this file I think that is ok, but for now we have this and should keep it working well.

taobojlen: I think that in this case, it's better to make the diff slightly larger but avoid introducing a new footgun!

ThisIsMissEm: @juanmrad it does keep working well — I was trying to avoid changing the bottle/IoC setup here, if we prefer the way @taobojlen proposed we can do that, but I wanted to not touch the existing registration code.

taobojlen (approved):
Thank you! 🙏

PR #1202, opened 2026-09-15
[581] Add actions and policy selection on additional content items

### LINKED ISSUE #581: Display available actions for "additional items"
Image

Sometimes content is found which is violative in this section- however can not be moderated.

juanmrad:

Closes #581.

https://www.loom.com/share/d74648b3420442e1a8f219082a71168d

Screenshot 2026-09-14 at 10 41 17 PM

Additional content on a review job (report extras and user-history items) had no way to take action. Reviewers could only act on the primary reported item.

This adds per-item action and policy selection on those additional content cards. Enqueued actions show as compact chips (name, policy, parameters, remove). Submit sends only items the reviewer marked with an action — not every additional item on the job — and publishes them with saved parameters and the latest item submission when available.

## Tests

- `client`: `ContentRelatedItemComponent.test.tsx`, `manualReviewTool.test.ts`
- `server`: `JobDecisioning.test.ts` (`relatedActionPublishPayloads`, `actionableRelatedActions`)
- Manual: enqueue/remove actions on additional items, pick policy, edit parameters, submit; confirm unmarked items are not actioned

## Checklist

  Did you update related docs?

- [x] **If the change is notable** (refer to Keep a Changelog conventions):
  Did you update CHANGELOG.md?

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting policies and applying actions to additional reported content items in the review console.
  * Added the ability to apply actions directly to selected thread messages.
  * Added controls to review, edit, and remove queued actions and their parameter values.
  * Preserved custom decision parameters when submitting related actions.

* **Bug Fixes**
  * Improved handling of queued related actions to prevent stale selections and exclude incomplete actions.
  * Corrected invalid date handling in review decision processing.

* **Documentation**
  * Updated the unreleased changelog with the expanded review console capabilities.


juanmrad:
@cubic-dev-ai review this PR

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -38,8 +40,83 @@ export type ManualReviewDecisionRelatedAction = {
...
+): ManualReviewDecisionRelatedAction[] {
+  return relatedActions.filter(
+    (relatedAction) =>
+      relatedAction.actionIds.some((id) => id.length > 0) &&
+      relatedAction.itemIds.some((id) => id.length > 0) &&
+      relatedAction.itemTypeId.length > 0,
+  );
+}
+
+export function sanitizeRelatedActionParameterPayloads(
```
taobojlen: I'd maybe call this `validateRelatedActionParameterPayloads`; "sanitize" makes me think it's a security-related thing and it's calling other functions called `validateFoo`

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -262,27 +349,93 @@ export default class JobDecisioning {
...
+          if (hasEmptyPolicyCustomAction) {
+            throw makeMissingRequiredPolicyForDecisionError({
+              shouldErrorSpan: true,
+            });
+          }
+          const relatedPolicyIds = [
+            ...new Set(
+              relatedActionsToPublish.flatMap((relatedAction) =>
+                relatedAction.policyIds.filter(
+                  (policyId) => policyId.length > 0,
```
taobojlen: Thinking out loud: this file has a lot of complexity from morphing one object shape into another. When I see stuff like `foos.filter(foo => foo.length > 0)`, it makes me wonder whether this kind of validation should really be happening at the API layer rather than in the business logic! (This is probably not something to fix here but worth mentioning).

server/graphql/modules/manualReviewTool.ts
```diff
@@ -2449,7 +2450,20 @@ const Mutation: GQLMutationResolvers = {
           jobId,
           lockToken,
           decisionComponents: decisionPayloads,
-          relatedActions: [...relatedItemActions],
+          relatedActions: actionableRelatedActions(
+            relatedItemActions.map((relatedAction) => ({
```
taobojlen: something like [lodash.pick](https://lodash.info/doc/pick) might make this a little more elegant

taobojlen (commented):
The code looks good to me! @juanmrad I just have one UX question... Right now, you can select an action on an additional item, but it doesn't appear in the right sidebar under "Other actions" until you also select a policy. So if I select an action on an additional item, but no policy, then the behaviour is a bit unclear to me as a user: does the action get applied, or does it get silently dropped? I.e. is the policy selection required, too?

It'd be great if we could reflect this more in the UI. So for example, if the user has selected an action only, but a policy is required, then we could disable the Submit button (and maybe show a tooltip on hover explaining why). Or, if the policy is *not* required, then maybe we could show the action under "Other actions" as soon as it's selected, even without a policy.

Does this make sense? Happy to record a video of what I mean if that's useful!

client/src/webpages/dashboard/mrt/manual_review_job/v2/ContentRelatedItemComponent.tsx
```diff
@@ -1,18 +1,459 @@
...
+export function AdditionalReportedContentItems(
+  props: {
+    items: readonly RelatedContentItem[];
+    unblurAllMedia: boolean;
+    excludeItemIds?: readonly string[];
+  } & RelatedContentActionProps,
+) {
+  const { items, excludeItemIds, unblurAllMedia, ...actionProps } = props;
+  const excluded = new Set(excludeItemIds ?? []);
+  const visibleItems = items.filter((item) => !excluded.has(item.id));
```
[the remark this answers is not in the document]
juanmrad: Fixed: additional items are now excluded with a composite `typeId:itemId` key so two items that share an ID across types are not collapsed.

server/services/manualReviewToolService/modules/JobDecisioning.ts
```diff
@@ -262,27 +349,93 @@ export default class JobDecisioning {
...
+                (policyId) =>
+                  policyId.length === 0 || !foundPolicyIds.has(policyId),
+              ),
+          );
+          if (relatedHasUnknownPolicy) {
+            throw makeMissingRequiredPolicyForDecisionError({
+              shouldErrorSpan: true,
+            });
+          }
         }
```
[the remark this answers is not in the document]
juanmrad: Fixed: unknown related policy IDs are now looked up and rejected even when `requires_policy_for_decisions` is off. Empty policies are still only required when that setting is enabled.

client/src/webpages/dashboard/mrt/manual_review_job/v2/ContentRelatedItemComponent.tsx
```diff
@@ -1,18 +1,459 @@
...
+                  const selected = arrayFromArrayOrSingleItem(policyIds);
+                  if (
+                    requirePolicySelectionToEnqueueAction &&
+                    selected.length === 0
+                  ) {
+                    return;
+                  }
+                  enqueue(pendingAction, policyIds);
+                  if (allowMoreThanOnePolicySelection) {
+                    setPendingPolicyIds(selected);
```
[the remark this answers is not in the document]
juanmrad: This pending-chip path is gone. Additional-item actions are enqueued immediately (so they show under Other Actions), and re-selecting policies on the queued chip already forwards the saved `customMrtApiParamDecisionPayload`.

server/services/manualReviewToolService/modules/JobDecisioning.ts
juanmrad: Renamed to `validateRelatedActionParameterPayloads`.

server/services/manualReviewToolService/modules/JobDecisioning.ts
juanmrad: Agreed this filtering would sit more cleanly at the API boundary, but I am leaving it here for this PR as you suggested.

server/graphql/modules/manualReviewTool.ts
juanmrad: Left the explicit mapping. `pick` would not copy the readonly arrays, and the optional parameter payload still needs a conditional spread.

juanmrad:
@taobojlen Policy is required when the org setting is on. Additional-item actions now enqueue immediately (so they show under Other Actions even before a policy is picked), chips and the sidebar mark **Policy required**, and Submit stays disabled with a tooltip until those policies are selected. The server still rejects related actions that are missing a required policy, and it now also rejects unknown policy IDs when policies are optional.

juanmrad:
Screenshot 2026-09-24 at 8 53 28 AM

PR #1204, opened 2026-09-15
Address #1203 Optimizing oldest job by creating a heap

### LINKED ISSUE #1203: Optimize `getOldestJobCreatedAt` with a Redis Mini-Heap Cache
## Problem

This is a quick follow for #718

`getOldestJobCreatedAt` scans up to 10,000 jobs from BullMQ's `prioritized` sorted set on every dashboard page render to find the minimum `createdAt`. For FIFO queues this is O(1) (first `waiting` job), but for prioritized queues (`NUM_REPORTS`) it transfers up to 10k job payloads from Redis on every call.

The dashboard's queue list resolves `oldestJobCreatedAt` as a GraphQL field per queue, so a page with 10 prioritized queues = 100k jobs scanned per render.

## Proposed Solution

Maintain a small cached heap (3 entries) of the oldest pending jobs per queue in Redis.

### Cache Key

```
{orgId}:mrt-oldest-jobs:{queueId} → JSON array of [{ jobId, createdAt }, ...]
```

Stored sorted oldest-first, capped at 3 entries.

### Read Path (Dashboard)

- Read `[0].createdAt` from the cached heap — O(1).
- On cache miss (cold start or expired), do a full scan, populate the heap, return `[0]`.

### Write Path — Enqueue

- Almost always a no-op: new jobs are newer than all 3 cached entries.
- Only insert into the heap when the queue has fewer than 3 pending jobs.

### Write Path — Dequeue

- If the dequeued job's ID matches any entry in the heap, remove it.
- If the dequeued job was `[0]` (the oldest), `[1]` becomes the new oldest — dashboard immediately has a correct answer with no scan.
- Trigger an **async background rescan** to refill the heap back to 3 entries. This never blocks the dequeue or the dashboard.

### Write Path — Admin Job Removal

- Same as dequeue: if the removed job is in the heap, evict it and backfill async.

### Write Path — Queue Deletion

- Delete the cache key.

### Edge Cases

- **3 rapid dequeues before rescan completes**: Unlikely since each dequeue requires a reviewer decision. If it does happen, fall back to a synchronous scan on the next dashboard read (same as today's behavior).
- **Priority recompute sweep**: Does not affect `createdAt` no cache invalidation needed.
- **Race between background rescan and concurrent dequeue**: The rescan writes the full heap atomically (single `SET`). A concurrent dequeue that removes an entry the rescan just added is fine the next dequeue check will evict it.

## Scope

- Only needed for prioritized queues (`NUM_REPORTS`). FIFO queues already get the oldest job in O(1) via `getWaiting(0, 0)`.
- Self-contained change in `QueueOperations.ts` + a new small cache helper.
- No migration, no schema change, no GraphQL change.

juanmrad:
Fixes #1203  and addresses some issues found during review:

- `toBullPriority` now floors at 1 instead of 0 BullMQ treats priority 0 as unprioritized, which would silently demoting heavily-reported items to the back of the wait list.
- `changePriority` calls now check job state first only waiting/prioritized/delayed jobs support it, active jobs are skipped.
- `getOldestJobCreatedAt` uses an in-memory mini-heap (3 entries) instead of scanning up to 10k prioritized jobs on every call. The heap is invalidated on dequeue/remove and lazily rescanned as to keep a relatively fresh number of oldest job that can be updated as items are removed.
- `numReports` stored as a number and formatted at render time which was breaking numeric sort in the preview table.
- Removed `normalizeJobSortType` as it was just a nullish coalesce with extra steps.

Priority is conditionally spread into BullMQ options instead of passing undefined explicitly. Legacy job format (id vs itemId) handled in the recompute sweep.

Tests added for the heap cache, dequeue eviction, bulk delete invalidation, and the active-job guard.

## Summary by CodeRabbit

- **Bug Fixes**
  - Report counts now display with localized number formatting while preserving correct numeric sorting.
  - Priority recalculations no longer alter jobs that are already active or otherwise unavailable for reprioritization.
  - Priority values exceeding the supported range now clamp to the correct minimum priority.
  - Queue age information remains accurate as jobs are removed, completed, or cleared.
- **Performance**
  - Repeated queue age lookups are faster through improved caching.


julietshen:
Can you add a description so this PR is easier to review, please? :)

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -167,6 +171,11 @@ export default class QueueOperations {
   >;
   private readonly transactionWithRetry: KyselyTransactionWithRetry<ManualReviewToolServicePg>;
 
+  // Mini-heap cache: keeps the N oldest pending jobs per queue so
+  // getOldestJobCreatedAt can return in O(1) instead of scanning up to
+  // 10k prioritized jobs. Keyed by `${orgId}:${queueId}`.
+  readonly #oldestJobHeaps = new Map<string, OldestJobEntry[]>();
```
taobojlen, line 177: This is an in-memory structure, so what happens if there are multiple instances of Coop running (say in k8s or some other way)? E.g. if another instance calls `removeJobByJobIdUnsafe`, then `oldestJobHeaps` in other instances will become out of date, right? I think this data structure may need to live in something like Redis perhaps?

server/services/manualReviewToolService/modules/QueueOperations.ts
```diff
@@ -1765,34 +1819,86 @@ export default class QueueOperations {
...
+    const { orgId, queueId, isAppealsQueue } = opts;
+    const heapKey = `${orgId}:${queueId}`;
+    const cached = this.#oldestJobHeaps.get(heapKey);
+
+    if (cached != null && cached.length > 0) {
+      return cached[0].createdAt;
+    }
+
+    // Cache miss — do the full scan and populate the heap.
+    const heap = await this.#scanOldestJobs({ orgId, queueId, isAppealsQueue });
```
taobojlen, line 1831: If this is a cache over the oldest-job calculation, and our oldest-job calculation is still incorrect (because it only checks 10k jobs), then is it really worth adding?

I'd be very curious to see some metrics to quantify how long it takes to scan 10k jobs, because this *might* be an unnecessary optimization. Hard to say without numbers!

taobojlen (changes requested):

PR #1206, opened 2026-09-15
Prefer refreshed content after dequeuing review jobs

sunilatlas:
## Problem

PR #1189 resolves temporary review content in the no-cache job query. Immediately after dequeue, the review component still prefers the original mutation response whenever it exists, so the first claimed review can render the stored URL instead of the refreshed query result. Reloading the page then uses the refreshed result.

## Change

Select review jobs in this order:

1. Closed job supplied by the closed-review flow.
2. Refreshed query result.
3. Dequeue mutation result while the query is loading.

This keeps the existing loading fallback and makes the first active review use resolved content as soon as the query returns.

## Validation

- Added focused tests for closed-job priority, refreshed-query priority, dequeue fallback, and empty state.
- Full client and server typechecking passes.
- Client lint and production build pass.
- Prettier and diff checks pass.

Follow-up to #1189.

## Summary by CodeRabbit

* **Refactor**
  * Improved the internal handling of manual review job selection without changing user-visible behavior.

* **Tests**
  * Added coverage for job selection priority, including closed, queried, dequeued, and unavailable jobs.


juanmrad (approved):
👍🏻

PR #1208, opened 2026-09-15
[196] Add reported NCMEC media to a hash bank

### LINKED ISSUE #196: Add reported images to HMA bank
As discussed in the April 2 issue triage call: Add image to the HMA bank If CSAM is detected and reported.

maarkN:

Closes #196

Media reported to NCMEC as CSAM never reached a hash bank, so the same content had to be caught again if it came back. Following the answers on the issue, orgs can now pick a bank in **Settings → NCMEC**, and each classified item is added to it once NCMEC accepts the report (production only).

196-dropdown-lists-org-banks

---

### A few things worth a look:

- **Same design as `#sendUserPreservationRequest`.** The banking is a private method in `NcmecReporting` with the same `withRetries` setup, and a failure fails the report the same way. I knew this means no unit tests for it (it's covered by integration tests), but I preferred to keep the existing pattern. Happy to move both to a queue in a follow-up if it becomes a bottleneck.
- **HMA `metadata` goes in the JSON body now,** and `addContentToBank` also supports `note`. The endpoint docstring in HMA says "query param", but the code only reads the body ([curation.py#L302-L336](https://github.com/maarkN/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py#L302-L336)), so the metadata was being dropped.
- **The bank must belong to the org.** The mutation checks it, and a bank from another org gets the same error as a bank that doesn't exist.

This adds a migration (nullable `reported_media_hash_bank_id` on `ncmec_org_settings`, FK to `hash_banks` with `ON DELETE SET NULL`). Could a maintainer sign off on it, as `AGENTS.md` asks?

## Tests

- **Unit:** `hmaService/index.test.ts` (metadata in the body, `note` up to 255 characters) and `ncmec.resolver.test.ts` (ownership, same error for another org's bank, id `0`).
- **Integration:** `ncmec-submission.integ.test.ts` covers a production report adding each item, no bank selected, a test submission, and HMA returning 500 (the report fails after retries, the preservation request still goes out, and no media URL ends up in the stored error).
- **UI:** picked a bank, saved and reloaded; cleared it, saved and reloaded.

Integration tests pass locally with a fresh image (`docker compose run --rm test npm run test:integ`: 7 suites, 21 tests).

## (Optional) Rollout Plan

The new column is nullable, so nothing changes until an org picks a bank.

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

- [x] **If you changed `db/src/scripts/**` and used `CREATE TABLE`, `ADD COLUMN`, or `ALTER COLUMN`:**

## Summary by CodeRabbit

* **New Features**
  * Added an NCMEC setting to select a hash bank for media from accepted production reports.
  * Selected media is added to the configured hash bank in the background, with retries for individual items.
  * Reports remain successful even if background hash-bank processing is temporarily unavailable.
* **Documentation**
  * Updated NCMEC integration guidance and local development instructions for the new setting and background processing.


server/services/ncmecService/ncmecService.ts
```diff
@@ -38,6 +38,7 @@ export class NcmecService {
     readonly tracer: Dependencies['Tracer'],
     readonly itemInvestigationService: Dependencies['ItemInvestigationService'],
     readonly getItemTypeEventuallyConsistent: Dependencies['getItemTypeEventuallyConsistent'],
+    readonly hmaService: Dependencies['HMAHashBankService'],
```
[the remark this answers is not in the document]
maarkN: This is on purpose for now. I keep the same design as `#sendUserPreservationRequest`, which already works this way: the report row is saved first, and a later failure turns the submission into `FAILURE`. Reviewers now get a specific message saying NCMEC accepted the report but banking failed. Making this resumable needs its own `pending` state or a _**queue**_, so I'd rather do it for both calls in a follow-up.

server/services/ncmecService/ncmecReporting.ts
```diff
@@ -2072,6 +2142,18 @@ export default class NcmecReporting {
               reportId: parseInt(reportId),
             });
           }
+
+          if (
+            ncmecConfig?.reported_media_hash_bank_id != null &&
+            reportParams.media.length > 0 &&
+            isTest === false
+          ) {
+            await this.#addReportedMediaToHashBank({
```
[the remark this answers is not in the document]
maarkN: Same point as the CodeRabbit comment above: I kept the preservation request's design.

server/services/hmaService/index.ts
```diff
@@ -802,33 +802,40 @@ export class HmaService {
...
-        if (metadata.content_uri)
-          params.append('content_uri', metadata.content_uri);
-        if (metadata.json)
-          params.append('metadata', jsonStringify(metadata.json));
+      // HMA reads `metadata` from the JSON body only; query-param metadata is dropped.
+      const params = new URLSearchParams({ url });
+      let body: string | undefined;
+      if (metadata || note) {
+        body = jsonStringify({
+          ...(metadata ? { metadata } : {}),
```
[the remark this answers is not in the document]
maarkN: This is the shape HMA expects: `_validate_bank_add_metadata` reads `metadata` from the JSON body and builds `BankedContentMetadata(content_id, content_uri, json)` from it ([curation.py#L177-L199](https://github.com/facebook/ThreatExchange/blob/a859350601ba2776cd684d90db75445dad83c6da/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py#L177-L199)). I also checked it against a local HMA 1.1.7, and the entry was stored with `content_id` and `json`.

server/graphql/modules/ncmecOrgSettingsValidation.ts
```diff
@@ -99,3 +100,25 @@ export function parseMediaReviewPolicy(input: NcmecOrgSettingsInputShape): {
...
 }
+
+/** Returns the bank id to store, or null to clear it. Only checks the format;
+ * the resolver checks that the bank belongs to the org. */
+export function parseReportedMediaHashBankId(
+  input: NcmecOrgSettingsInputShape,
+): number | null {
+  const raw = input.reportedMediaHashBankId?.trim() ?? '';
+  if (raw === '') {
+    return null;
```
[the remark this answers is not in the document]
maarkN: True, but it is the same for the other optional NCMEC settings, like `defaultNcmecQueueId`. The client always sends this field, so I prefer to keep it consistent and change all fields together in another PR.

server/services/ncmecService/ncmecReporting.ts
```diff
@@ -1722,6 +1723,69 @@ export default class NcmecReporting {
     await fetchWithRetries();
   }
 
+  async #addReportedMediaToHashBank(input: {
+    orgId: string;
+    reportedMedia: ReadonlyArray<Pick<Media, 'id' | 'typeId' | 'url'>>;
+    reportId: string;
+  }) {
+    const { orgId, reportedMedia, reportId } = input;
+    const reportedMediaHashBankId = await this.reportedMediaHashBankId(orgId);
```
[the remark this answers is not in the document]
maarkN: Good catch, I will use the bank id from `ncmecConfig` instead of reading it again. The preservation request has the same issue, I can fix it in another PR.

Fixed in e5f1225

server/services/ncmecService/ncmecReporting.ts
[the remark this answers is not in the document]
ThisIsMissEm: We should probably use a Queue for this, as we really don't want HMA being down to break NCMEC reporting consistency.

server/workers_jobs/ReportedMediaBankingWorker.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it('lets a failed HMA call through, so the queue retries the job', async () => {
+    const deps = makeDeps({
+      addContentToBank: jest
+        .fn()
+        .mockRejectedValue(new Error('Failed to add content to bank: 500')),
+    });
+
+    await expect(bankReportedMedia(deps, JOB)).rejects.toThrow(
```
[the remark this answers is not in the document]
maarkN: Fixed in 3a241e71.

server/test/integ/ncmec-submission.integ.test.ts
```diff
@@ -152,4 +217,139 @@ describe('NCMEC submitReport (integration)', () => {
...
+      expect(result).toBe('SUCCESS');
+      expect(reportedMediaBankingEnqueue).not.toHaveBeenCalled();
+    },
+    60_000,
+  );
+
+  testWithFixture(
+    'keeps the report successful when the banking jobs cannot be enqueued',
+    async ({ deps, orgId, reportId, userItemTypeId }) => {
+      const { stub, ncmecReporting } = makeReporting(
```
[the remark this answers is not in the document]
maarkN: Fixed in 3a241e71.

server/workers_jobs/ReportedMediaBankingWorker.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export default inject(
+  ['IORedis', 'Tracer', 'HMAHashBankService'],
+  (redis: IORedis.Redis | Cluster, tracer, hmaService) => {
+    let worker: BullWorker<ReportedMediaBankingJobData> | undefined;
+
+    return {
+      type: 'Worker' as const,
+      async run(_signal) {
+        worker = new BullWorker<ReportedMediaBankingJobData>(
```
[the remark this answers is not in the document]
maarkN: Fixed in 3a241e71.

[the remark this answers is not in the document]
maarkN: Fixed in 3a241e71.

server/test/integ/ncmec-submission.integ.test.ts
```diff
@@ -152,4 +217,139 @@ describe('NCMEC submitReport (integration)', () => {
...
+      const job = (itemId: string, url: string) => ({
+        orgId,
+        hashBankId: hashBank.id,
+        ncmecReportId: reportId,
+        itemId,
+        itemTypeId: userItemTypeId,
+        url,
+      });
+      expect(reportedMediaBankingEnqueue).toHaveBeenCalledTimes(1);
+      expect(reportedMediaBankingEnqueue).toHaveBeenCalledWith([
```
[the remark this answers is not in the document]
taobojlen: I think this seems like a sensible thing to fix! I like integration tests :)

docs/integrations/ncmec.md
```diff
@@ -34,20 +34,21 @@ Configure NCMEC reporting under **Settings** → **NCMEC Settings**.
...
+| **More Info URL**                        | URL for additional information about your reporting process (e.g. `https://yourcompany.com/ncmec-info`). Used as the web page URL when "Default internet detail type" is set to "Web page."                                                                                                            |
+| **Default NCMEC Queue**                  | When reviewers click "Enqueue to NCMEC," jobs are sent to this queue. Leave as "Use org default queue" to fall back to the organization's default queue.                                                                                                                                               |
+| **Hash Bank for Reported Media**         | Media from each CyberTip accepted by NCMEC is added to this hash bank so it can be matched if it is uploaded again. Adding runs in the background (`ReportedMediaBankingWorker`), so it never blocks or fails the CyberTip. Only applies to [production submissions](#test-vs-production-submissions). |
```
taobojlen:
```suggestion
| **Hash Bank for Reported Media**         | Media from each CyberTip accepted by NCMEC is added to this hash bank so it can be matched if it is uploaded again. Only applies to [production submissions](#test-vs-production-submissions). |
```

No need to give implementation details in the docs.

server/graphql/modules/ncmec.resolver.test.ts
```diff
@@ -17,15 +17,21 @@ const VALID_INPUT = {
 
 function makeCtx(permissions: readonly UserPermission[]) {
   const updateNcmecOrgSettings = jest.fn(async () => undefined);
+  const getBankById = jest.fn(async (orgId: string, id: number) =>
```
taobojlen, line 20: do we really need to mock this?

server/graphql/modules/ncmec.resolver.test.ts
```diff
@@ -140,3 +146,106 @@ describe('updateNcmecOrgSettings media review policy', () => {
...
+  it('persists a bank that belongs to the org', async () => {
+    const { ctx, updateNcmecOrgSettings, getBankById } = makeCtx([
+      UserPermission.MANAGE_ORG,
+    ]);
+    await Mutation.updateNcmecOrgSettings(
+      {},
+      { input: { ...VALID_INPUT, reportedMediaHashBankId: '42' } },
+      ctx,
+    );
+    expect(getBankById).toHaveBeenCalledWith('org-1', 42);
```
taobojlen, line 160: feels to me like this is testing implementation, a bit of an antipattern. i generally prefer towards integration-y tests, especially with LLMs writing most code now.

server/graphql/modules/ncmec.resolver.test.ts
```diff
@@ -140,3 +146,106 @@ describe('updateNcmecOrgSettings media review policy', () => {
...
+        {},
+        { input: { ...VALID_INPUT, reportedMediaHashBankId: '0' } },
+        ctx,
+      ),
+    ).rejects.toThrow('Selected hash bank was not found.');
+    expect(getBankById).toHaveBeenCalledWith('org-1', 0);
+    expect(updateNcmecOrgSettings).not.toHaveBeenCalled();
+  });
+
+  it('rejects a bank id above the Postgres integer range before looking it up', async () => {
```
taobojlen, line 222: this seems unnecessary.

server/graphql/modules/ncmec.resolver.test.ts
```diff
@@ -140,3 +146,106 @@ describe('updateNcmecOrgSettings media review policy', () => {
...
+  });
+
+  it('rejects a non-numeric bank id', async () => {
+    const { ctx, updateNcmecOrgSettings, getBankById } = makeCtx([
+      UserPermission.MANAGE_ORG,
+    ]);
+    await expect(
+      Mutation.updateNcmecOrgSettings(
+        {},
+        { input: { ...VALID_INPUT, reportedMediaHashBankId: '4x2' } },
```
taobojlen, line 244: is this needed? typescript should prevent this from happening anyway?

server/graphql/modules/ncmecOrgSettingsValidation.ts
```diff
@@ -20,6 +20,7 @@ export type NcmecOrgSettingsInputShape = {
   contactPersonPhone?: string | null;
   mediaReviewRequirement?: string | null;
   minMediaToReview?: number | null;
+  reportedMediaHashBankId?: string | null;
```
taobojlen, line 23: why is this a string and not a number?

server/iocContainer/services/workersAndJobs.ts
```diff
@@ -30,6 +32,11 @@ declare module '../index.js' {
 
 export function registerWorkersAndJobs(bottle: Bottle<Dependencies>) {
   register(bottle, 'ItemProcessingWorker', makeItemProcessingWorker);
+  register(
```
taobojlen: you might need to rebase here after @ThisIsMissEm 's PR btw! https://github.com/roostorg/coop/pull/1197

server/iocContainer/index.ts
```diff
@@ -734,6 +740,11 @@ export default async function getBottle(
   bottle.factory('itemSubmissionRetryQueueBulkWrite', (container) =>
     makeItemSubmissionBulkWrite(container.IORedis, ITEM_SUBMISSION_DLQ_NAME),
   );
+  // Enqueued from the NCMEC submission path, so a Redis outage has to fail
```
taobojlen, line 742: i'm not sure i follow this comment! can you explain what the intent is in case of a redis outage?

IMO, we should *not* add special handling for postgres or redis downtime. those two services are so fundamental to how coop works, we can safely assume they're up or everything breaks anyway.

server/services/hmaService/index.ts
```diff
@@ -950,4 +955,15 @@ export class HmaService {
   }
 }
 
+function toHmaNote(note: string | undefined): string | undefined {
+  if (!note?.trim()) {
+    return undefined;
+  }
+  // HMA counts code points (Python len), not UTF-16 units.
+  if ([...note].length > 255) {
+    throw new Error('note must be 255 characters or less');
```
taobojlen: would it maybe be cleaner here to trim the string to 255 characters in this case? feels like having the hash in HMA is more desirable than ensuring we get the full note in every case

server/services/ncmecService/ncmecReporting.ts
```diff
@@ -2072,6 +2074,41 @@ export default class NcmecReporting {
...
+          const reportedMediaHashBankId =
+            ncmecConfig?.reported_media_hash_bank_id;
+
+          if (
+            reportedMediaHashBankId != null &&
+            reportParams.media.length > 0 &&
+            isTest === false
+          ) {
+            // Banking runs in its own worker. Enqueueing must never fail an
+            // accepted report, so a Redis outage is logged and dropped here.
```
taobojlen, line 2087: IMO we do not need special handling for redis outages. we should assume that our job queue is always available or coop basically falls over entirely.

client/src/webpages/settings/NCMECSettings.tsx
```diff
@@ -83,6 +88,7 @@ type NcmecSettings = {
   contactPersonPhone: string;
   mediaReviewRequirement: GQLNcmecMediaReviewRequirement;
   minMediaToReview: string;
+  reportedMediaHashBankId: string;
```
taobojlen, line 91: as in another comment -- shouldn't this be a number?

client/src/webpages/settings/NCMECSettings.tsx
```diff
@@ -573,6 +583,43 @@ export default function NCMECSettings() {
...
+                </SelectItem>
+                {(data?.hashBanks ?? []).map((bank) => (
+                  <SelectItem key={bank.id} value={bank.id}>
+                    {bank.name}
+                  </SelectItem>
+                ))}
+              </SelectContent>
+            </Select>
+            <Text size="XS" className="text-gray-500">
+              Media from each report accepted by NCMEC is added to this bank so
```
taobojlen, line 617: I think this might be a bit off -- should it be "Media from each report sent to NCMEC"?

server/queues/reportedMediaBankingQueue.ts
```diff
@@ -0,0 +1,87 @@
...
+type RedisConnection = IORedis.Redis | Cluster;
+
+/** Factory for the function that enqueues reported media for banking. */
+function makeReportedMediaBankingEnqueue(redis: RedisConnection) {
+  let queue: Queue<ReportedMediaBankingJobData> | undefined;
+
+  // Built on first use: creating it opens a Redis connection, and most
+  // processes that depend on this service never enqueue anything.
+  function getQueue() {
+    if (queue === undefined) {
```
taobojlen, line 35: this logic differs from `server/queues/itemSubmissionQueue.ts`. do we need the built-on-first-use pattern here? and do we need the `queue.on('error', ...)` and so on?

taobojlen: to be clear everything here *seems* sensible, but since the other queue doesn't do it, i wonder if it's necessary or handled globally.

taobojlen (changes requested):

docs/integrations/ncmec.md
maarkN: Fixed in f4edc7af.

server/services/hmaService/index.ts
maarkN: Fixed in 1e49f86c.

PR #1231, opened 2026-09-16
Consistently use build/ directory for built code

ThisIsMissEm:

Rename the TypeScript output directory from transpiled/ to build/ across server, nodejs-instrumentation and @roostorg/coop-types, so every package emits to the same place.

This is a precursor to being able to use import maps, and normally just how the build output of `tsc` is structured.

Changes:
- server: outDir, both start scripts, Dockerfile, the e2e workflow, both jest configs, both eslint configs, the e2e fixture constant, and the three ignore files. Also adds rootDir so the layout under outDir is explicit rather than inferred from the common ancestor of the input files.
- nodejs-instrumentation: outDir, clean script, Dockerfile, and the start:trace reference from server. Also drops a dead main (nothing resolves this package by name, and the image flattens build/ into /autoinstrumentation, so that path does not exist there) and marks the package private.
- @roostorg/coop-types: outDir plus module, typings, files and exports. Consumer-invisible, because exports declares only ".", so deep imports into transpiled/ were never reachable.

build/ was already ignored by .gitignore and .dockerignore, so this mostly removes the transpiled entries rather than adding new ones.

I also caught a few cases where the package.json `license` was set to ISC despite the entire project being Apache-2.0

There will be a follow up PR that makes the `server/build/` hermetic, such that the `package.json` is inside the `build/` directory in the docker image, which allows import maps to fully work.

## Tests

Full suite run, and passed. Also worked with the container to ensure it started up correctly.

## (Optional) Rollout Plan

n/a

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

## Summary by CodeRabbit

- **Build and Deployment**
  - Standardized compiled application and package output under the `build` directory.
  - Updated container images, server startup commands, and automated end-to-end workflows to use the new build output.
  - Improved packaging so generated artifacts and required package metadata are included where needed.

- **Package Updates**
  - Updated package exports, entry points, and build test paths to reference the new output structure.
  - Updated package licensing to Apache-2.0.
  - Marked the instrumentation package as private.


ThisIsMissEm:
@cubic-dev-ai all the license changes were in accordance with the LICENSE at the root of the repository, so the old `ISC` license was actually incorrect.

types/package.json
```diff
@@ -3,12 +3,12 @@
   "type": "module",
   "version": "2.4.0",
```
juanmrad, line 4: I wonder if we should bump the version given the change.

juanmrad (approved):
LGTM. one minor comment.

types/package.json
ThisIsMissEm: I wasn't sure how version is happening for the package, hence leaving it. Normally I use release tooling that handles versioning automatically. What would we define the version bump as? major, minor or patch?

juanmrad: patch I would guess 🤔 it does change license and exports, but may be ok.

PR #1252, opened 2026-09-17
Remove unnecessary types

ThisIsMissEm:

We had a collection of types declared in `decs.d.ts` which were either unused completely, or unneeded. This gets us down to just two global typescript declarations for `Express.User` and the `String.toLowerCase` / `String.toUpperCase` type overrides.

The `@graphql-tools/schema` type declaration seems to be completely unnecessary, since that duplicated the built-in types exactly.

`@types/restify` was completely unused, so I suspect it was left over from something.

The `typeRoots` in `tsconfig.json` was unneeded as `./shims` doesn't exist anymore and `@types/*` is automatically loaded by typescript.

## Tests

The usual CI checks.

## (Optional) Rollout Plan

n/a

## Checklist

  Did you update related docs?

  Did you update CHANGELOG.md?

---
## Summary by cubic
Removes unused and duplicated type declarations from `server/decs.d.ts` so the `@graphql-tools/schema` shim no longer hides the package's real types. The only behavior change: `makeExecutableSchema` now receives `Context` explicitly, because the shim's broken import was making `TContext` infer as `GraphQLResolveInfo`.

**Details**
- Moves the `Express.User` and `String` case-conversion augmentations into `globals.ts` and adds it to knip's entry list.
- Replaces the hand-written `latlon-geohash` declarations with `@types/latlon-geohash`, and drops the duplicated `@stdlib/stats-binomial-test` declarations.
- Removes the dead `@types/restify` override and the stale `typeRoots` entry pointing at the deleted `./shim`.

<sup>Written for commit 2c5a380c5b8cc37f189ec48e741ea9bc794d05d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/roostorg/coop/pull/1252?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Refactor**
  - Improved type consistency across server-side integrations and request handling.
  - Updated schema context typing for more reliable compile-time checks.
  - Consolidated shared type definitions and improved string case-conversion type support.

- **Chores**
  - Updated project type-checking configuration and validation coverage.
  - Refined server development configuration for more consistent tooling behavior.

- **User Impact**
  - No visible changes to application functionality or user-facing behavior.


server/package.json
```diff
@@ -106,6 +106,7 @@
     "@types/graphql-depth-limit": "^1.1.6",
     "@types/jest": "^29.2.4",
     "@types/js-yaml": "^4.0.5",
+    "@types/latlon-geohash": "^2.0.4",
```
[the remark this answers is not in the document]
ThisIsMissEm: I've manually approved this change.

server/globals.ts
```diff
@@ -0,0 +1,36 @@
+// Global type augmentations for types we don't own.
+//
+// These live in a regular module (rather than an ambient `.d.ts`) so they're
```
taobojlen, line 3: tiny nit: as a reader, i don't think i would wonder "why does this live here and not in a `.d.ts` file?", so IMO this part of the comment might be unnecessarily verbose.

but this may just be my typescript knowledge that's lacking; your call whether to change anything here!

taobojlen (approved):

server/globals.ts
ThisIsMissEm: That's there because I _did_ wonder how this worked because I didn't remember this specific of typescript

PR #1264, opened 2026-09-18
Fix child policy apply-to-sub-policies toggle

### LINKED ISSUE #1255: "Apply to sub-policies" toggle in the policy scores table does nothing
### What happened

In the **User Strikes → Policy Scores** tab, the "Apply to sub-policies" toggle in the
child-policies *table* does nothing. Toggling it does not update
`applyUserStrikeCountConfigToChildren`, so the setting is silently discarded and never
reaches the save mutation.

The same control in the card/tree layout above the table works correctly.

**Cause:** the table's cell renderer passes `onChange` to `<Switch>`, but `Switch` wraps
`@radix-ui/react-switch`, whose `Root` renders a `<button role="switch">`. Buttons do not
fire `change` events, so the handler is never invoked. Radix's API for this is
`onCheckedChange`.

`client/src/webpages/dashboard/userStrikes/PolicyScoresTab.tsx`, in `ChildPoliciesTable`:

```tsx
<Switch
  disabled={editingDisabled}
  onChange={(event) => {
    const { checked } = event.target as HTMLInputElement;
    // …never runs
  }}
/>
```

Compare the working instance in `PolicyScoresTab`'s `renderPolicy`, ~215 lines earlier in
the same file:

```tsx
<Switch
  disabled={!editingPolicies.includes(policy.value.id)}
  defaultChecked={policy.value.applyUserStrikeCountConfigToChildren}
  onCheckedChange={(checked) => { /* … */ }}
/>
```

### Expected

Toggling "Apply to sub-policies" in the table updates the policy's
`applyUserStrikeCountConfigToChildren` value and persists on save, matching the behaviour
of the equivalent control in the card layout.

### Fix

Two changes in `ChildPoliciesTable`'s cell renderer:

1. `onChange` → `onCheckedChange={(checked) => …}`, dropping the `event.target` cast.
2. Add the missing `checked` / `defaultChecked` prop — the table's switch is currently
   uncontrolled and does not reflect the stored value either.

`StrikeEnabledActionsTab.tsx` is a good reference for the intended shape; every other
`Switch` in `client/` already uses `onCheckedChange`.

### Version/commit

`6037e3f0`

### Affected browser(s)

All — not browser-specific.

### Relevant logs/errors

No runtime error; the handler simply never fires. It surfaces as a type error under
TypeScript 6, which rejects the `EventTarget & HTMLButtonElement` → `HTMLInputElement`
cast that was concealing the mismatch:

```shell
src/webpages/dashboard/userStrikes/PolicyScoresTab.tsx(514,37): error TS2352:
Conversion of type 'EventTarget & HTMLButtonElement' to type 'HTMLInputElement' may be a
mistake because neither type sufficiently overlaps with the other. If this was
intentional, convert the expression to 'unknown' first.
```

### Anything else

Found while bumping `client/` to TypeScript 6.0.3. That PR keeps the diff mechanical and
silences the cast via `unknown` with a `TODO` pointing here, rather than changing
behaviour — hence this separate issue.

A regression test belongs with the fix; `client/src/coop-ui/Switch.test.tsx` already
covers the primitive, so the gap is at the `ChildPoliciesTable` level.

TayfurYldz:

Fixes #1255. The “Apply to sub-policies” switch in the child policies table now uses the current UI state and saves the selected value for the correct child policy. Draft score state is also cleared after a successful save so stale values are not reused.

## Tests

- `npx vitest run src/webpages/dashboard/userStrikes/PolicyScoresTab.test.tsx` — 2 passed
- ESLint on both changed files — passed
- `npx tsc --noEmit` — passed
- Prettier — passed
- `git diff --check` — passed

## (Optional) Rollout Plan

None.

## Checklist

- [ ] **If you changed anything user-facing:** Did you update related docs?
- [ ] **If the change is notable:** Did you update CHANGELOG.md?
- [ ] Database migration check
- [ ] Signal classification check

## Summary by CodeRabbit

- **Bug Fixes**
  - Policy score changes now correctly reflect the latest saved or pending values.
  - Discarding edits reliably removes policy changes.
  - Child-policy toggle updates now save the correct setting and policy.
  - Saved policy entries are cleared after successful updates.


ThisIsMissEm:
@TayfurYldz please make sure you fill out the pull request template, looks like you've one outstanding issue from cubic, and then I can approve the workflows to run.

ThisIsMissEm:
Can you please make sure you're following our [submitting code changes](https://community.roost.tools/software-development-practices/submitting-code-changes.html) policy?

juanmrad:
👋🏻 Hi @TayfurYldz, thanks for working on this. Before we do another review, could you please:

- Describe in your own words why the save/refetch lifecycle changes are needed in addition to the switch fix. 
- Did you manually verify both unchecked -> checked and checked -> unchecked persist after save/reload, ideally with a screenshot or recording.

We ask contributors to document their manual verification before requesting review so reviewers can focus on design and correctness rather than reproducing baseline validation. Thank you!

PR #1268, opened 2026-09-19
Rename environment variable ITEM_INVESTIGATION_AND_STRIKES_ENABLED to SCYLLA_ENABLED

ThisIsMissEm:

This prevents an unreleased feature (#889 / #918) from being released with an environment variable that we'd be renaming in #1235 — the only "breaking" change there for this feature would then be dropping `yes` and `no` as supported boolean values, but that's happening across all the boolean flags.

## Tests

standard CI/CD

## (Optional) Rollout Plan

n/a

## Checklist

  **Did you update related docs?** Original PR didn't document new environment variable, so leaving as undocumented for now.

  Did you update CHANGELOG.md?

---
## Summary by cubic
Renames the `ITEM_INVESTIGATION_AND_STRIKES_ENABLED` environment variable to `SCYLLA_ENABLED` to match the `SCYLLA_*` connection settings it controls. The old variable name is no longer read, so anyone who set it must update their environment; the default remains enabled when unset, so an unupdated config silently falls back to Scylla enabled. No other behavior changes.

<sup>Written for commit 8733d05c03c8072307f3b3933a893ff3511d5d88. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/roostorg/coop/pull/1268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Changed**
  - Renamed the Scylla feature toggle environment variable to `SCYLLA_ENABLED`.
  - Scylla remains enabled by default; setting `SCYLLA_ENABLED=false` disables the feature.
  - Updated configuration guidance and related documentation to reflect the new variable name.


server/iocContainer/index.ts
```diff
@@ -783,9 +783,7 @@ export default async function getBottle(
             executionContext,
           );
         },
-        itemInvestigationAndStrikesEnabled(
-          process.env.ITEM_INVESTIGATION_AND_STRIKES_ENABLED,
-        ),
+        itemInvestigationAndStrikesEnabled(process.env.SCYLLA_ENABLED),
```
ThisIsMissEm, line 786: I've deliberately kept the existing name as this method disappears entirely in #1235, and as it is, it's not using `isEnvTrue` from `server/iocContainer/utils.ts` and negating it, but instead implementing the inverse — this is why I flattened all env vars to `Env.schema.boolean()` in #1235, to prevent issues like this.

cassidyjames (approved):

PR #1272, opened 2026-09-21
Enforce safe item type configuration updates

taobojlen:
## Summary by cubic
Enforces backward-compatible and atomic item type updates so unsafe configuration changes are rejected before any write happens.

- Rejects schema changes that rename or remove fields, alter field types or container shapes, add required fields, or make optional fields required.
- Validates that hidden fields exist in the schema; `hiddenFields` is now settable on create/update, with omitted updates preserving existing values and an empty list clearing them.
- Validates field role references against the schema, including expected scalar types and kind-specific dependencies (e.g., `parentId` requires `threadId` and `createdAt`).
- Item type name conflicts now surface as a dedicated `ItemTypeNameAlreadyExistsError` instead of a generic database error, and failed updates leave the original row and hidden fields untouched.
- Cross-organization updates and hidden-field writes now return a not-found error instead of silently no-oping.

<sup>Written for commit 1f907935195d2ebfd43713a1748207154fa0a78e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/roostorg/coop/pull/1272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


taobojlen:
@cubic-dev-ai review this PR

server/services/manualReviewToolService/modules/JobRendering.ts
```diff
@@ -24,18 +30,47 @@ export default class JobRendering {
...
-      .insertInto('manual_review_tool.manual_review_hidden_item_fields')
-      .values({
-        org_id: opts.orgId,
-        item_type_id: opts.itemTypeId,
-        hidden_fields: [...opts.hiddenFields],
-      })
-      .onConflict((oc) =>
-        oc.columns(['org_id', 'item_type_id']).doUpdateSet({
+    const setHiddenFields = async (query: Kysely<JobRenderingPg>) => {
+      if (opts.hiddenFields.length === 0) {
```
taobojlen: Before, if a caller passed in `hiddenFields: []`, then we'd keep an orphan row in the database with an empty array. Now, we delete the existing rows for hidden fields instead. (This is not strictly necessary but a small cleanup).

server/services/manualReviewToolService/modules/JobRendering.ts
```diff
@@ -24,18 +30,47 @@ export default class JobRendering {
...
+        .where('id', '=', opts.itemTypeId)
+        .where('org_id', '=', opts.orgId)
+        .forUpdate()
+        .executeTakeFirst();
+      if (itemType === undefined) {
+        throw makeNotFoundError('Item type not found', {
+          shouldErrorSpan: false,
+        });
+      }
+      assertHiddenFieldsExist(itemType.fields, opts.hiddenFields);
```
taobojlen: This validates that the hidden fields being inserted are actually real fields.

server/services/manualReviewToolService/manualReviewToolService.ts
```diff
@@ -392,6 +398,28 @@ export class ManualReviewToolService {
     );
   }
 
+  forTransaction(trx: Transaction<CombinedPg>) {
```
taobojlen: This function gives us a short-lived `ManualReviewToolService` whose database queries use the supplied transaction instead of the normal connection pool.

server/services/moderationConfigService/modules/ItemTypeOperations.ts
```diff
@@ -128,7 +150,7 @@ export default class ItemTypeOperations {
   }
 
   /** Drop cached latest item types for this org after a DB write (MV + in-memory cache can disagree). */
-  private async invalidateLatestItemTypesCache(orgId: string): Promise<void> {
+  async invalidateLatestItemTypesCache(orgId: string): Promise<void> {
```
taobojlen: Callers can now manage transactions that span multiple services (in this case, both the item type service and the manual review service). So they need a way to invalidate this instance's cache after the transaction commits.

